PR #5, opened 2016-07-11
Add licenses

Ishiihara:


Ishiihara:
@ewencp Updated the PR to include `notices` and `license.html`. I also add in maven to clean up the contents in `notices` and `licences`. Not sure this is the best way to handle this, but `mvn -P licenses-source clean package` can be used to update the licenses and notices.

gwenshap:
there are no "notice" files for the Apache components, is this intentional?

ewencp:
@Ishiihara LGTM

ewencp:
@gwenshap They may not be including the NOTICE in their binary distro.

PR #7, opened 2016-07-22
Bump to 3.1.0-SNAPSHOT

Ishiihara:


ewencp:
@Ishiihara I think the licenses version should be 3.1.0-SNAPSHOT too? Otherwise LGTM

Ishiihara:
@ewencp Good catch. Will update that as well.

PR #8, opened 2016-07-26
Avoid null pointer exception

Ishiihara:


Ishiihara:
@ewencp Should be an easy review.

ewencp:
@Ishiihara LGTM, though it looks like we've got some test failures that only happen on Jenkins. Although I'm also a bit confused as I see a different number of tests locally than on Jenkins.

PR #11, opened 2016-07-29
Fix data format

Ishiihara:


Ishiihara:
@ewencp Ping for review. Thanks!

ewencp:
@Ishiihara Change looks straightforward enough. Seems like the failing test may be only happening on Jenkins, but `mapper_parse_exception` seems like an odd thing to get in a transient failure (i.e. rather than sometime timeout related or just the failed assertion)?

ewencp:
retest this please

PR #16, opened 2016-08-06
Add support for optional schema for non primitive types.

kyal:
Two issues addressed here in this commit:
1) NPE thrown in preProcessValue when a field is of type optional Struct and its value is null.
The input value is perfectly valid but preProcessValue is called on the null struct which then attempts to dereference its fields according to the schema. Added null check to preProcessValue to return the default value or null if the field is optional.

2) Concurrent collection modification exception throw during batch request processing in HTTPClient#constructBulk line 58.

task.put() -> writer.write() can attempt to append a SinkRecord to the first batch in the 'requests' list at the same time that a BulkTask attempts to send it to the HTTP client. This arises from the peekFirst() result being enqueued and after that happens then the records.pollFirst() removes the batch. This code locks requests and dequeues the batch we intend to use with BulkTask first so a batch cannot have records added to it once it is associated with a BulkTask.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -175,6 +175,11 @@ public class DataConverter {
       switch (schemaName) {
         case Decimal.LOGICAL_NAME:
           builder = SchemaBuilder.float64();
+          if (schema.isOptional())
+            builder.optional();
+
+          if (schema.defaultValue() != null)
+            builder.defaultValue(preProcessValue(schema.defaultValue(), schema, schema));
           return builder.build();
         case Date.LOGICAL_NAME:
         case Time.LOGICAL_NAME:
@@ -190,6 +195,8 @@ public class DataConverter {
       case ARRAY:
         valueSchema = schema.valueSchema();
         builder = SchemaBuilder.array(preProcessSchema(valueSchema));
+        if (schema.isOptional())
+          builder.optional();
         return builder.build();
       case MAP:
         keySchema = schema.keySchema();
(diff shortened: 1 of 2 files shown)


Ishiihara:
ok to test

kyal:
I don't understand the above automated build test failures. Those tests succeed with the proper values on those exact lines when I run JUnit with the committed code.

shikhar:
ok to retest

shikhar:
retest this please

src/main/java/io/confluent/connect/elasticsearch/internals/BulkProcessor.java
```diff
@@ -93,11 +93,11 @@ public void run() {
           if (guaranteeOrdering) {
             muted = true;
           }
-          executorService.submit(new BulkTask(batch));
```
shikhar: @kyal this extraneous change in `BulkProcessor` is causing the test fails

shikhar: Ah I see you were trying to fix 

```
2) Concurrent collection modification exception throw during batch request processing in HTTPClient#constructBulk line 58.
```

as per your PR description.

shikhar:
@kyal the `DataConverter` change LGTM, I'm looking into the `BulkProcessor` issue and transient test failures.

Meanwhile, can you please sign the CLA: http://clabot.confluent.io/cla

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -200,12 +207,16 @@ static Schema preProcessSchema(Schema schema) {
             .field(MAP_KEY, preProcessSchema(keySchema))
             .field(MAP_VALUE, preProcessSchema(valueSchema))
             .build());
+        if (schema.isOptional())
```
mwong38: Is it better to `break` and have one single `if (schema.isOptional()) ...` block at the bottom rather than have duplicate code?

mwong38:
Here's a unit test for testing the optional schema function if anyone is interested:

https://gist.github.com/mwong38/01f02de1e819f689618bb09845c47fab.js

shikhar:
Thanks @kyal and @mwong38, I created an updated PR here #36

PR #17, opened 2016-08-09
Fix mixup in order of log arguments that causes a stacktrace to be omitted.

ewencp:


ewencp:
@Ishiihara Trivial fix, should be an easy review.

ewencp:
retest this please

PR #18, opened 2016-08-10
Converting source object to string when building Bulk object

mmolimar:
The argument passed to Index.Builder is an ObjectNode (from jackson lib) in HttpClient class. When JestClient executes the action to ES, it serializes that object by means of Gson lib (to string) and creates a JSON object with another format that it actually is because of Gson is using ReflectiveTypeAdapterFactory -ObjectNode is not annotated with JsonAdapter-. Something like this happens:
- Original (source field):
{"field1": "value1", "field2" : "value2"}
- Converted (source field):
{"_children": {"field1" : {"_value" : "value1"}, "field2": {"_value" : "value2"}}}

### DIFF
--- a/docs/elasticsearch_connector.rst
+++ b/docs/elasticsearch_connector.rst
@@ -82,14 +82,7 @@ is available in Elasticsearch::
        "_id" : "test-elasticsearch-sink+0+0",
        "_score" : 1.0,
        "_source" : {
-         "_children" : {
-           "f1" : {
-             "_value" : "value1"
-           }
-         },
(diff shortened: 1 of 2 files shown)


ewencp:
@mmolimar We actually had a basically identical patch that just hadn't quite gotten merged yet. I just merged and updated the `0.10.0.0` branch. I'm going to close this as it is now redundant, but thanks for the contribution!

mmolimar:
Great Ewen!
Many thanks ;-)

Issue #14, opened 2016-08-03
Exception when commiting offsets, Bulk Insert


srijiths:
When ES Sink bulk insertion and commit offsets happens , there are exceptions. Created a gist . Pleas see the exceptions 
```
org.apache.kafka.connect.errors.ConnectException: Flush failed with non retriable exception.
	at io.confluent.connect.elasticsearch.ElasticsearchWriter.flush(ElasticsearchWriter.java:319)
	at io.confluent.connect.elasticsearch.ElasticsearchSinkTask.flush(ElasticsearchSinkTask.java:137)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.commitOffsets(WorkerSinkTask.java:275)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:155)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:142)
	at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:140)
	at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:175)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: org.apache.kafka.connect.errors.ConnectException: Cannot finish flush messages within 10000
	at io.confluent.connect.elasticsearch.ElasticsearchWriter.flush(ElasticsearchWriter.java:314)
	... 11 more
[2016-08-03 03:05:21,942] ERROR Rewinding offsets to last committed offsets (org.apache.kafka.connect.runtime.WorkerSinkTask:278)
[2016-08-03 03:05:21,942] ERROR Commit of WorkerSinkTask{id=elasticsearch-sink-0} offsets threw an unexpected exception:  (org.apache.kafka.connect.runtime.WorkerSinkTask:180)
org.apache.kafka.connect.errors.ConnectException: Flush failed with non retriable exception.
	at io.confluent.connect.elasticsearch.ElasticsearchWriter.flush(ElasticsearchWriter.java:319)
[...]

Issue #13, opened 2016-08-03
Unexpected error when registering ES connect


srijiths:
I have an MQTT connector configured in a connect-distributed environment. Then i tried to setup ES Sink connector. I am getting below exception.
![screenshot from 2016-08-03 08 23 19](https://cloud.githubusercontent.com/assets/1046319/17352358/aaba8274-5953-11e6-8962-d5a6240101d8.png)

My ES Sink configuration is like this
![screenshot from 2016-08-03 08 25 22](https://cloud.githubusercontent.com/assets/1046319/17352440/30bb8cd8-5954-11e6-87e9-ea9ddc17e014.png)
Due to this the connect interface itself is failing. Any pointers on this  ? Thanks

ewencp:
This doesn't really look specific to the ES connector. It looks from the stack trace like we're somehow getting a `null` config map somehow. Is the MQTT connector required to reproduce this? Is it reproducible just by running the single command you provided on a new Connect cluster?

ewencp:
@srijiths That seems fine, but I can't reproduce the issue with the sequence of commands you provided. What version of Kafka/Connect are you running? I doubt the combination of MQTT and ES connectors is the issue since the stacktrace doesn't seem like it has anything to do with a conflict.

srijiths:
@ewencp i am using confluvent 3.0.0 version. I tried to register above ES connector again. 

When i observed it has WARN on Concurrent Modification Exception from BulkProcessor class followed by JAVA HEAP SIZE exception. I have 4G alloted to JAVA. Then it follows a series of exceptions and kafka connect , rest evrything stopped. Since its a lengthy stack trace , created a gist. Please see here

https://gist.github.com/srijiths/459725df0137c881090f7a03565f784f

shikhar:
@srijiths this appears to be the root of the problem: `java.lang.OutOfMemoryError: Java heap space`

1. It's possible that the ES connector is buffering so much data that you are running out of heap because of that. `max.buffered.records` defaults to 100000 which is rather high, you could try lowering that.

2. Are you running with customized heap settings? Using the standard scripts, you'd end up with `KAFKA_HEAP_OPTS=-Xmx256M` .You may want to consider growing that to say `512M` if lowering the buffered records config does not help.

shikhar:
Please reopen if there are any unexpected issues here. The defaults for `max.buffered.records` and `batch.size` have been significantly lowered.

PR #21, opened 2016-09-01
CC-301: Lower default values of `max.buffered.records` and `batch.size`

shikhar:
Also some slight cleanup around throwing `RetriableException` when `max.buffered.records` is hit

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -67,12 +67,12 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   private static final String MAX_BUFFERED_RECORDS_DOC =
       "Approximately the max number of records each task will buffer. This config controls the memory usage for each task. When the number of "
       + "buffered records is larger than this value, the partitions assigned to this task will be paused.";
-  private static final long MAX_BUFFERED_RECORDS_DEFAULT = 100000;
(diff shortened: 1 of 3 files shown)


shikhar:
cc @ewencp for review

shikhar:
Forgot to note motivation: probably helps with #13 and #14

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -67,12 +67,12 @@
...
       "Approximately the max number of records each task will buffer. This config controls the memory usage for each task. When the number of "
       + "buffered records is larger than this value, the partitions assigned to this task will be paused.";
-  private static final long MAX_BUFFERED_RECORDS_DEFAULT = 100000;
+  private static final long MAX_BUFFERED_RECORDS_DEFAULT = 20000;
   private static final String MAX_BUFFERED_RECORDS_DISPLAY = "Max Number of Records to Buffer";
 
   public static final String BATCH_SIZE_CONFIG = "batch.size";
   private static final String BATCH_SIZE_DOC = "The number of requests to process as a batch when writing to Elasticsearch.";
-  private static final int BATCH_SIZE_DEFAULT = 10000;
+  private static final int BATCH_SIZE_DEFAULT = 2000;
```
ewencp: It'd be worth looking at how large the request payload gets to be with an average workload and this batch size. It's easy to batch *too* much data too, although this is at least only a maximum. We can revisit this when we have some perf data to look at.

ewencp:
LGTM

PR #22, opened 2016-09-02
CC-305: Log error from BulkProcessor and prevent the writer from adding furter records by setting the exception field

shikhar:


shikhar:
addressing this in another PR

PR #26, opened 2016-09-09
BulkProcessor refactor

shikhar:
The first commit is unrelated but was convenient to fold into this PR.

The main aim is to make the concurrency story in `BulkProcessor` clearer. 

Note that `io.confluent.connect.elasticsearch.bulk` is now independent of any ES connector-specific bits so we can pull it out if we need similar batching elsewhere. The previous impl tried to do this but there were a few gaps.


shikhar:
@ewencp need to add tests around retries but otherwise reviewable

src/main/java/io/confluent/connect/elasticsearch/internals/HttpClient.java
```diff
@@ -1,95 +0,0 @@
...
-    // We shutdown the JEST client when sink tasks are stopped.
-  }
-
-  private Bulk constructBulk(RecordBatch batch, Callback<Response> callback) {
-    Bulk.Builder builder = new Bulk.Builder();
-    List<ESRequest> requests = batch.requests();
-    for (ESRequest request: requests) {
-      JsonNode data = null;
-      try {
-        data = objectMapper.readTree(request.getPayload());
```
shikhar: This was a useless JSON reparse that is no longer happening, so should shave some cycles.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -131,188 +93,102 @@
...
+          client, type, ignoreKey, ignoreSchema, topicConfigs, flushTimeoutMs, maxBufferedRecords, maxInFlightRequests, batchSize, lingerMs, maxRetry, retryBackoffMs
+      );
     }
   }
 
   public void write(Collection<SinkRecord> records) {
-    if (bulkProcessor.getException() != null) {
-      throw new ConnectException("BulkProcessor failed with non-retriable exception", bulkProcessor.getException());
-    }
-    if (bulkProcessor.getTotalBufferedRecords() + records.size() > maxBufferedRecords) {
```
shikhar: This represents a behavior change, now it's going to block in `bulkProcessor.add()` till buffer space becomes available. Maybe it's better to throw a `RetriableException` if it helps with consumer liveness? If so I can restore that.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -0,0 +1,429 @@
...
+  }
+
+  /**
+   * Add a record, may block upto {@code timeoutMs} if at capacity with respect to {@code maxBufferedRecords}.
+   *
+   * If any task has failed prior to or while blocked in the add, {@link ConnectException} will be thrown with that error.
+   */
+  public synchronized void add(R record, long timeoutMs) {
+    final long addStartTimeMs = time.milliseconds();
+    for (long elapsedMs = time.milliseconds() - addStartTimeMs;
```
ewencp: This code is called in a pretty tight loop since it is invoked for every record. Might be worth trying to avoid at least 1 of the 2 mandatory calls to `time.milliseconds()` as those calls aren't always cheap (we've seen performance problems w/ it in `KafkaProducer` before). You could probably even get rid of both if you check the conditions for adding a record before even recording the start time.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -0,0 +1,429 @@
...
+      @Override
+      public void uncaughtException(Thread t, Throwable e) {
+        log.error("Uncaught exception in BulkProcessor thread {}", t, e);
+        failAndStop(e);
+      }
+    };
+    return new ThreadFactory() {
+      @Override
+      public Thread newThread(Runnable r) {
+        final Thread t = new Thread(r, "BulkProcessor-t" + threadCounter.getAndIncrement());
```
ewencp: Do you want to attach more info to the thread name since this won't identify which BulkProcessor they belong to. For example, if you have multiple tasks running in the same process.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -0,0 +1,429 @@
...
+          final BulkResponse bulkRsp = bulkClient.execute(bulkReq);
+          if (bulkRsp.isSucceeded()) {
+            return bulkRsp;
+          }
+          retriable = bulkRsp.isRetriable();
+          throw new ConnectException("Bulk request failed: " + bulkRsp.getErrorInfo());
+        } catch (Exception e) {
+          if (retriable && remainingRetries > 0) {
+            log.warn("Failed to execute batch {} of {} records, retrying after {} ms", batchId, batch.size(), retryBackoffMs, e);
+            time.sleep(retryBackoffMs);
```
ewencp: Should we be checking for whether we need to shut down, and/or using `wait` instead of `time.sleep`? I am wondering if the # of retries can basically exceed the amount of time we want to wait for, e.g., shutdown, in which case we'd still be trying to send data even though we really wanted to have stopped already. This obviously gets a bit tricky since you want to flush during shutdown, but maybe there's some final flag we should be setting that will force these tasks to bail out even before all retries are exhausted?

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -0,0 +1,429 @@
...
+
+  private void failAndStop(Throwable t) {
+    error.compareAndSet(null, toConnectException(t));
+    stop();
+  }
+
+  /**
+   * @return count of currently buffered records
+   */
+  public synchronized int unsentRecords() {
```
ewencp: A bunch of these accessors seem to be unused. Remove them until they are actually needed?

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -0,0 +1,429 @@
...
+   * @return count of batches that have been created
+   */
+  public long createdBatches() {
+    return createdBatches.get();
+  }
+
+  /**
+   * @return count of batches successfully executed
+   */
+  public long successfulBatches() {
```
ewencp: Without these unsynchronized accessors, it doesn't look like we need `successfulBatches` and `successfulRecords` to be `AtomicLong`s since they are only touched in the (`synchronized`) `onBatchCompletion`.

src/test/java/io/confluent/connect/elasticsearch/bulk/BulkProcessorTest.java
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  @Test
+  public void addBlocksWhenBufferFull() {
+    final Client client = new Client();
+
+    final int maxBufferedRecords = 1;
+    final int maxInFlightBatches = 1;
+    final int batchSize = 1;
+    final int lingerMs = 10;
```
ewencp: I'd probably make this very conservatively high any time you want lingering to have basically no effect. Sometimes jenkins gets surprisingly slow, especially when running multiple jobs at once.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
shikhar: Good idea, I'll add object ID or a static counter field that goes into this.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
shikhar: Cool, will add an initial check for the add precondition to avoid these calls altogether in the happy path

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
shikhar: Ack

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
shikhar: Ack, don't need them at all actually, was just for exposing some stats (but should probably remove them until they are actually needed as you pointed out)

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
shikhar: I think it's fine as is... if the shutdown timeout is exceeded while some tasks are still trying to send data, the `executor.shutdownNow()` from `awaitStop()` would cause this thread to wakeup with an `InterruptedException` I believe.

shikhar: I do wonder though whether we should even call `bulkProcessor.flush()` from `ElasticsearchWriter.stop()`. I can see how it's useful for HDFS where offsets are being managed by the connector, but here I'd expect the framework to re-put() those records upon anyway, since the framework is tracking the last time the _framework_ called flush.

src/test/java/io/confluent/connect/elasticsearch/bulk/BulkProcessorTest.java
shikhar: ah yeah, I'm doing that in the `flushing()` test but in this one it doesn't matter since the BP worker thread is not started

shikhar:
@ewencp addressed your comments

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
ewencp: Well, the framework tries to go through a flush + offset commit anyway. In practice, it probably doesn't matter since a) ideally there shouldn't be any data left to flush (if that last commit was successful) and b) even if there is, it's data where we expect the possibility of duplicates anyway. But for (b) maybe it is best to try to minimize that.

ewencp:
@shikhar Followed up on the one item about whether flush() should be called in stop(), but otherwise LGTM

PR #28, opened 2016-09-21
bump kafka to 0.10.2.0-SNAPSHOT

graphaelli:
confluentinc/kafka@6acbf27aebbe5b6b4f230ce44ea5151c4c53ffbd makes using 0.10.1.0-SNAPSHOT quite challenging

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
 
     <properties>
         <confluent.version>3.1.0-SNAPSHOT</confluent.version>
-        <kafka.version>0.10.1.0-SNAPSHOT</kafka.version>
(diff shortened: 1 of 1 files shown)


shikhar:
ok to test

ewencp:
@graphaelli There is an `0.10.1` branch in apache/kafka that tracks any upcoming 0.10.1.X releases. We don't want to move beyond that for the moment, so this will probably continue tracking 0.10.1.0-SNAPSHOT for the time being. Using it should be as easy as building that branch of Kafka.

graphaelli:
Ah that makes sense, I had assumed this project was tracking trunk and just missed the version bump on the dependency.

PR #30, opened 2016-09-26
CC-312: fence writes with offset as ES document version

shikhar:
The first commit is unrelated but was helpful for test changes.


shikhar:
@ewencp for review

shikhar:
TODO: documentation update, as it discusses reordering from batching/pipelining/retrying

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
```diff
@@ -134,8 +136,33 @@ public void testCompatible() throws Exception {
...
+
+    final ElasticsearchWriter writer = initWriter(client, ignoreKey, false, Collections.<String, TopicConfig>emptyMap());
+    writer.write(Arrays.asList(sinkRecord0, sinkRecord1));
+    writer.flush();
+
+    // write the record with earlier offset again
+    writeDataAndRefresh(writer, Collections.singleton(sinkRecord0));
+
+    // last write should have been ignored due to version conflict
+    verifySearchResults(Collections.singleton(sinkRecord1), ignoreKey);
```
ewencp: Is this actually a sufficient test given how `verifySearchResults` works? It only validates that the IDs match, but the IDs for these documents will be the same. This test passes even if you pass `sinkRecord0` here instead of `sinkRecord1`.

The version is probably returned, so to properly validate I think you'd really want a variant of `verifySearchResults` where you can pass expected versions in. (And arguably might want to validate both before and after the second write.)

shikhar: Good catch!

shikhar:
@ewencp fixed the test issue in 7fdccb20836cbf707247da2e7086021015b17b4d, now verifying the source for each document

PR #36, opened 2016-10-13
Support optional & default values when pre-processing schemas & values	

shikhar:
Replaces #16

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -53,11 +53,13 @@ public class DataConverter {
     if (key == null) {
       throw new ConnectException("Key is used as document id and can not be null.");
     }
-    Schema.Type schemaType;
+
+    final Schema.Type schemaType;
     if (keySchema == null) {
       schemaType = ConnectSchema.schemaType(key.getClass());
-      if (schemaType == null)
+      if (schemaType == null) {
         throw new DataException("Java class " + key.getClass() + " does not have corresponding schema type.");
+      }
     } else {
       schemaType = keySchema.type();
     }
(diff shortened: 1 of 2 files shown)


shikhar:
Although this PR refactors the test, I verified the existing version of `DataConverterTest` passes with the changes being made here to `DataConverter`.

shikhar:

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -120,39 +120,59 @@ static Schema preProcessSchema(Schema schema) {
...
         return schema;
+      }
     }
   }
 
+  private static SchemaBuilder copySchemaBasics(Schema source, SchemaBuilder target) {
+    if (source.isOptional()) {
+      target.optional();
+    }
+    if (source.defaultValue() != null && source.type() != Schema.Type.STRUCT) {
```
ewencp: I'm guessing default values on structs will be unusual, but not strictly speaking disallowed afaik. Any reason we're skipping this step for structs?

shikhar: There's a bit of weird circularity there which I don't think can be resolved. To create the default value you need to supply the struct `Schema` to the `Struct` constructor, and to get the struct `Schema` from the struct `SchemaBuilder` you need to specify the `defaultValue`...

ewencp: You actually can make this work because `SchemaBuilder` implements the `Schema` interface. So to construct the default value, you create it using the (incomplete) `SchemaBuilder`.

shikhar: Hmm, this is actually ok, no exception because of instance equality on the `SchemaBuilder`:

```java
    @Test
    public void structDefaultValue() {
        SchemaBuilder builder = SchemaBuilder.struct();
        builder.defaultValue(new Struct(builder));
        builder.build();
    }
```

so I must have been doing something off

ewencp: @shikhar To be fair, I'm actually not entirely sure how well this works in practice since you get an "incomplete" schema in the default value. It's definitely not an ideal solution since you can do weird things (e.g. create partial schema, create default value, then add more fields to the schema such that the default value is invalid).

PR #41, opened 2016-10-22
Fix topic index override

### LINKED ISSUE #40: topic.index.map is ignored on record put
According to documentation, `topic.index.map` config defines mapping between Kafka topics and Elasticsearch indices. I have the following map configured: `topic.index.map=mytopic:myindex`, so I expect records from `mytopic` to be added to `myindex`. But in `io.confluent.connect.elasticsearch.ElasticsearchWriter#write` method I see another behavior - topic name is used for schema mapping, despite the fact that mapped index name was used during initialization and index creation. So I always receive __index_not_found_exception__ in `io.confluent.connect.elasticsearch.Mapping#createMapping` as long as I have index `myindex`, but connector is trying to find `mytopic` one. Quick fix solved this issue for me:

    String mappedIndex = topicToIndexMap.get(sinkRecord.topic());
    final String index = mappedIndex != null ? mappedIndex : sinkRecord.topic();

...but I'm not sure if I've got the idea of the given config correctly and using it in a proper way. Could you kindly check if it's bug or not?

Thanks

shikhar:
@ewencp for review

fixes #40

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
@@ -190,7 +190,8 @@ public class ElasticsearchWriter {
 
   public void write(Collection<SinkRecord> records) {
     for (SinkRecord sinkRecord : records) {
-      final String index = sinkRecord.topic();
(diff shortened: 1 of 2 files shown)


ewencp:
@shikhar LGTM. Not sure if it's a good or bad thing that we had the test in place and it was just buggy...

PR #42, opened 2016-10-22
3.1.x to master

shikhar:


ewencp:
Seems fine. @shikhar when these are trivial merges, feel free to skip over the PR overhead for the time being. It's fine if you want the additional Jenkins test, but especially where we are now where master and 3.1.x are basically the same anyway, its fine to optimize a bit :)

shikhar:
@ewencp easier to PR from my phone ;)

shikhar:
PR'ing definitely carries the danger of hitting `Delete branch`. Now restored.

PR #49, opened 2016-11-10
Schema.Type.BYTES should map to 'binary' ES datatype

shikhar:


src/main/java/io/confluent/connect/elasticsearch/Mapping.java
```diff
@@ -166,6 +167,9 @@ private static JsonNode inferPrimitive(String type, Object defaultValue) {
         case ElasticsearchSinkConnectorConstants.STRING_TYPE:
           defaultValueNode = JsonNodeFactory.instance.textNode((String) defaultValue);
           break;
+        case ElasticsearchSinkConnectorConstants.BINARY_TYPE:
+          defaultValueNode = JsonNodeFactory.instance.binaryNode(bytes(defaultValue));
```
shikhar: On the `DataConverter` side of things, `JsonConverter` already takes care of `BYTES` schema to be treated as Jackson `binaryNode(..)`(translated by Jackson to JSON as base64, which is what ES expects for `binary` fields)

shikhar:

ewencp:
LGTM

PR #50, opened 2016-11-11
fix for mapping exists check

mrauter:
doesMappingExist checks if the "type" root key exists but json result is:
```
{
  "mappings" : {
    "type" : {
    ...
    }
}
```
and if no mapping exists: {}

so just check if the result is not empty... or as alternative:
`JsonObject typeJson = resultJson.getAsJsonObject("mappings").getAsJsonObject(type);`

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/Mapping.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/Mapping.java
@@ -79,8 +79,7 @@ public class Mapping {
     if (resultJson == null) {
       return false;
     }
-    JsonObject typeJson = resultJson.getAsJsonObject(type);
-    return typeJson != null;
+    return (resultJson.entrySet().size() != 0);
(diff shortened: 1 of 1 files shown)


shikhar:
ok to test

shikhar:
@mrauter can you please sign the CLA here, the CLABot goofed up: clabot.confluent.io/cla

mrauter:
@shikhar done

shikhar:
This should have a new test in `MappingTest`.

I think `resultJson.getAsJsonObject("mappings").getAsJsonObject(type)` as you proposed with an additional null-check seems safest:
```
JsonObject mappingsJson = resultJson.getAsJsonObject("mappings");
return mappingsJson != null && mappingsJson.getAsJsonObject(type) != null;
```

mrauter:
yes i agree

PR #53, opened 2016-11-16
CC-392: Fix mapping existence check

shikhar:
Tweaked version of #50 with test

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/Mapping.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/Mapping.java
@@ -66,21 +66,19 @@ public class Mapping {
   }
 
   /**
-   * Check the whether a mapping exists or not for a type.
-   * @param client The client to connect to Elasticsearch.
-   * @param index The index to write to Elasticsearch.
-   * @param type The type to check.
-   * @return Whether the type exists or not.
+   * Get the JSON mapping for given index and type. Returns {@code null} if it does not exist.
    */
-  public static boolean doesMappingExist(JestClient client, String index, String type) throws IOException {
-    GetMapping getMapping = new GetMapping.Builder().addIndex(index).addType(type).build();
-    JestResult result = client.execute(getMapping);
-    JsonObject resultJson = result.getJsonObject().getAsJsonObject(index);
-    if (resultJson == null) {
-      return false;
+  public static JsonObject getMapping(JestClient client, String index, String type) throws IOException {
+    final JestResult result = client.execute(new GetMapping.Builder().addIndex(index).addType(type).build());
+    final JsonObject indexRoot = result.getJsonObject().getAsJsonObject(index);
+    if (indexRoot == null) {
+      return null;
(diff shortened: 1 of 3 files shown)


shikhar:

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
```diff
@@ -66,27 +58,21 @@ public static void createMapping(JestClient client, String index, String type, S
   }
 
   /**
-   * Check the whether a mapping exists or not for a type.
-   * @param client The client to connect to Elasticsearch.
-   * @param index The index to write to Elasticsearch.
-   * @param type The type to check.
-   * @return Whether the type exists or not.
+   * Get the JSON mapping for given index and type. Returns {@code null} if it does not exist.
```
shikhar: The test code was retrieving the mapping correctly, while the existence check here was wrong. So I thought it best to just have a retrieval method here such that `null` return indicates missing.

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
```diff
@@ -47,14 +47,6 @@
...
   private static final Logger log = LoggerFactory.getLogger(Mapping.class);
 
-  /**
-   * Create an explicit mapping.
-   * @param client The client to connect to Elasticsearch.
-   * @param index The index to write to Elasticsearch.
-   * @param type The type to create mapping for.
-   * @param schema The schema used to infer mapping.
-   * @throws IOException
-   */
```
kkonstantine: Why is the javadoc removed? I would expect it to expand to cover all public methods.

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
```diff
@@ -66,27 +58,21 @@ public static void createMapping(JestClient client, String index, String type, S
   }
 
   /**
-   * Check the whether a mapping exists or not for a type.
-   * @param client The client to connect to Elasticsearch.
-   * @param index The index to write to Elasticsearch.
-   * @param type The type to check.
-   * @return Whether the type exists or not.
```
kkonstantine: Same javadoc question here.

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
shikhar: I removed the Javadoc because they did not add any information beyond what the method and parameter names already conveyed. Since this is implementation rather than library code, I think it's reasonable to omit docs in such cases.

kkonstantine: I tend to prefer to add javadocs to every public method (and class) even when the argument names are self-explanatory, unless of course they override methods defined in interfaces or other classes. 

But I'm definitely not willing to force this opinion to others.

shikhar:
Reverted the unnecessary Javadoc removal.

PR #56, opened 2016-11-25
add jest node discovery feature

mrauter:
enables periodic node discovery of jest for high availability

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -40,6 +40,7 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   public static final String TOPIC_KEY_IGNORE_CONFIG = "topic.key.ignore";
   public static final String SCHEMA_IGNORE_CONFIG = "schema.ignore";
   public static final String TOPIC_SCHEMA_IGNORE_CONFIG = "topic.schema.ignore";
+  public static final String NODE_DISCOVERY_FREQUENCY = "discovery.frequency";
 
   protected static ConfigDef baseConfigDef() {
     final ConfigDef configDef = new ConfigDef();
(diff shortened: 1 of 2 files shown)


ewencp:
ok to test

@mrauter Sorry for the confusion with the bot, you can sign [here](http://clabot.confluent.io/cla)

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -57,6 +58,10 @@ protected static ConfigDef baseConfigDef() {
           .define(MAX_IN_FLIGHT_REQUESTS_CONFIG, Type.INT, 5, Importance.MEDIUM,
                   "The maximum number of indexing requests that can be in-flight to Elasticsearch before blocking further requests.",
                   group, 5, Width.SHORT, "Max In-flight Requests")
+          .define(NODE_DISCOVERY_FREQUENCY, Type.LONG, 60, Importance.MEDIUM,
```
ewencp: Hmm, is there going to be an equivalent in every library? This seems like it exposes implementation details. For example, if we decided to switch to the [Java REST client maintained alongside ES](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/index.html), which didn't exist when we initially implemented this, would there be something equivalent we could use?

More generally, I'm a bit curious about the use of such a config. If you need this to handle your cluster's nodes moving to different hosts, doesn't this mean your application would break if you, e.g., had a process failure and had to restart from the original config? If not, doesn't that imply you already have some sort of discovery mechanism (DNS, load balancer, etc) that handles this for you? Or am I misunderstanding the use case for node discovery?

mrauter: I use this feature to detect node failures/outages because otherwise JEST doesn't recognize if a node is not available anymore. The node discovery feature updates the node state periodically.

The Java REST client of Elastic has build in a feature like this.

nirajpatel:
I agree that this feature is pretty important especially when using Elasticsearch in the cloud.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
nirajpatel: I agree that this feature is pretty important especially when using Elasticsearch in the cloud.

levzem:
we have switched over to the Elasticsearch java client in #468 , hence closing

PR #69, opened 2017-02-28
Fix map output for String keyed maps

cotedm:
@kkonstantine this is a first run at fixing map output for String keyed maps. Things I know are problematic currently:
1. `MappingTest` ends up with a StackOverflowException because we are recursing on a map of strings in that test. Still looking into that.
2. We should decide if only String keyed maps should behave this way. I'm not sure what the ES standard is.
3. One weird thing I found was that if in `DataConverterTest.stringKeyedMap` I use a BigDecimal for validation instead of an INT32, I have a hard time validating the map values because the `DataConverter.preProcessValues` returns a `Double` object instead of a `BigDecimal` as I would expect it to. This seems to be consistent with and without my changes, so it's good to verify if that's expected behavior.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -131,14 +131,13 @@ public class DataConverter {
         Schema valueSchema = schema.valueSchema();
         String keyName = keySchema.name() == null ? keySchema.type().name() : keySchema.name();
         String valueName = valueSchema.name() == null ? valueSchema.type().name() : valueSchema.name();
-        Schema elementSchema;
         if (keySchema.type() == Schema.Type.STRING) {
           return SchemaBuilder.map(preProcessSchema(keySchema), preProcessSchema(valueSchema)).build();
         }
-        elementSchema = SchemaBuilder.struct().name(keyName + "-" + valueName)
-                  .field(MAP_KEY, preProcessSchema(keySchema))
-                  .field(MAP_VALUE, preProcessSchema(valueSchema))
-                  .build();
+        Schema elementSchema = SchemaBuilder.struct().name(keyName + "-" + valueName)
+                .field(MAP_KEY, preProcessSchema(keySchema))
+                .field(MAP_VALUE, preProcessSchema(valueSchema))
+                .build();
         return copySchemaBasics(schema, SchemaBuilder.array(elementSchema)).build();
       }
       case STRUCT: {
@@ -194,8 +193,6 @@ public class DataConverter {
     }
 
     Schema.Type schemaType = schema.type();
-    Schema keySchema;
-    Schema valueSchema;
     switch (schemaType) {
       case ARRAY:
         Collection collection = (Collection) value;
@@ -205,19 +202,19 @@ public class DataConverter {
         }
         return result;
       case MAP:
-        keySchema = schema.keySchema();
-        valueSchema = schema.valueSchema();
-        ArrayList<Struct> mapStructs = new ArrayList<>();
-        Map<?, ?> map = (Map<?, ?>) value;
+        Schema keySchema = schema.keySchema();
         Schema newValueSchema = newSchema.valueSchema();
+        Schema valueSchema = schema.valueSchema();
+        Map<?, ?> map = (Map<?, ?>) value;
         if (keySchema.type() == Schema.Type.STRING) {
-          Map<Object, Object> toReturn = new HashMap<>();
+          Map<Object, Object> processedMap = new HashMap<>();
           for (Map.Entry<?, ?> entry: map.entrySet()) {
-            toReturn.put(preProcessValue(entry.getKey(), keySchema, newSchema.keySchema()),
-                    preProcessValue(entry.getValue(), valueSchema, newSchema.valueSchema()));
+            processedMap.put(preProcessValue(entry.getKey(), keySchema, newSchema.keySchema()),
+                    preProcessValue(entry.getValue(), valueSchema, newValueSchema));
           }
-          return toReturn;
+          return processedMap;
         }
+        ArrayList<Struct> mapStructs = new ArrayList<>();
         for (Map.Entry<?, ?> entry: map.entrySet()) {
           Struct mapStruct = new Struct(newValueSchema);
           mapStruct.put(MAP_KEY, preProcessValue(entry.getKey(), keySchema, newValueSchema.field(MAP_KEY).schema()));
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -37,6 +37,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConstants.MAP_KEY;
@@ -135,9 +136,9 @@ public class DataConverter {
           return SchemaBuilder.map(preProcessSchema(keySchema), preProcessSchema(valueSchema)).build();
         }
         Schema elementSchema = SchemaBuilder.struct().name(keyName + "-" + valueName)
-                .field(MAP_KEY, preProcessSchema(keySchema))
-                .field(MAP_VALUE, preProcessSchema(valueSchema))
-                .build();
+             .field(MAP_KEY, preProcessSchema(keySchema))
+             .field(MAP_VALUE, preProcessSchema(valueSchema))
+             .build();
         return copySchemaBasics(schema, SchemaBuilder.array(elementSchema)).build();
       }
       case STRUCT: {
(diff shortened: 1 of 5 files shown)


ewencp:
@cotedm re: 3, I believe the conversion you're seeing happens because we need to convert to types supported in JSON and numeric types just end up as doubles in JS. Though I guess it might not be necessary -- I'm not sure if Jackson will handle BigDecimal or not.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -130,10 +131,14 @@ static Schema preProcessSchema(Schema schema) {
         Schema valueSchema = schema.valueSchema();
         String keyName = keySchema.name() == null ? keySchema.type().name() : keySchema.name();
         String valueName = valueSchema.name() == null ? valueSchema.type().name() : valueSchema.name();
-        Schema elementSchema = SchemaBuilder.struct().name(keyName + "-" + valueName)
-            .field(MAP_KEY, preProcessSchema(keySchema))
-            .field(MAP_VALUE, preProcessSchema(valueSchema))
-            .build();
+        Schema elementSchema;
```
kkonstantine: No need for early declaration. `elementSchema` is not used in the block for String types.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -130,10 +131,14 @@ static Schema preProcessSchema(Schema schema) {
...
+        if (keySchema.type() == Schema.Type.STRING) {
+          return SchemaBuilder.map(preProcessSchema(keySchema), preProcessSchema(valueSchema)).build();
+        }
+        elementSchema = SchemaBuilder.struct().name(keyName + "-" + valueName)
```
kkonstantine: Also, with the above comment, the lines that appear as changed here don't need to change, unless you want to apply the new indentation.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -205,6 +210,14 @@ static Object preProcessValue(Object value, Schema schema, Schema newSchema) {
         ArrayList<Struct> mapStructs = new ArrayList<>();
         Map<?, ?> map = (Map<?, ?>) value;
         Schema newValueSchema = newSchema.valueSchema();
+        if (keySchema.type() == Schema.Type.STRING) {
+          Map<Object, Object> toReturn = new HashMap<>();
```
kkonstantine: I'd use a slightly better name. Since the returned object below mentions what it is (`mapStructs`) you can do the same here. Suggestions: `rawMap`, `plainMap`, `processedMap` ... etc

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -205,6 +210,14 @@ static Object preProcessValue(Object value, Schema schema, Schema newSchema) {
         ArrayList<Struct> mapStructs = new ArrayList<>();
```
kkonstantine: Declaration needs to move below, closer to where it's used.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -205,6 +210,14 @@ static Object preProcessValue(Object value, Schema schema, Schema newSchema) {
...
         Map<?, ?> map = (Map<?, ?>) value;
         Schema newValueSchema = newSchema.valueSchema();
```
kkonstantine: Move below, or better reuse instead of `newSchema.valueSchema()` in the "string" case.

kkonstantine:
The changes look good. 

It's interesting how String keys were used in the `DataConverterTest` and non-String keys in `ElasticsearchWriterTest`. I wonder if we had done the opposite whether the issue would have been surfaced earlier. 

The issue with `MappingTest` is that `preProcessSchema` on a map now returns a map instead of a struct when keys are strings. Therefore, we need to break the infinite recursion in the `Type.MAP` when verifying. 

Having noticed that, I'd be happy to see a test case slightly more high level as well. I'll paste below a test I wrote before you submitted the PR. Feel free to re-use it.

kkonstantine:
The test should go to `ElasticsearchWriterTest`

```java
  @Test
  public void testStringKeydMap() throws Exception {
    boolean ignoreKey = false;
    boolean ignoreSchema = false;

    Schema mapSchema = SchemaBuilder.map(Schema.STRING_SCHEMA, Schema.INT32_SCHEMA).build();

    Map<String, Integer> map = new HashMap<>();
    map.put("One", 1);
    map.put("Two", 2);

    SinkRecord sinkRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, key, mapSchema, map, 0);

    ElasticsearchWriter writer = initWriter(client, ignoreKey, ignoreSchema);
    writeDataAndRefresh(writer, Collections.singletonList(sinkRecord));

    Collection<?> expectedRecords = Collections.singletonList(new ObjectMapper().writeValueAsString(map));
    verifySearchResults(expectedRecords, TOPIC, ignoreKey, ignoreSchema);
  }
```
You'll need a generalization of `verifySearchResults` in `ElasticsearchSinkTestBase` as follows: 

```java

  protected void verifySearchResults(Collection<?> records, String index, boolean ignoreKey, boolean ignoreSchema) throws IOException {
```
and then: 
```java
    for (Object record : records) {
      if (record instanceof SinkRecord) {
        IndexableRecord indexableRecord = DataConverter.convertRecord((SinkRecord) record, index, TYPE, ignoreKey, ignoreSchema);
        assertEquals(indexableRecord.payload, hits.get(indexableRecord.key.id));
      } else {
        assertEquals(record, hits.get("key"));
      }
    }
  }
```

kkonstantine:
Thanks @cotedm. Will re-review once the mapping test is fixed.

cotedm:
@kkonstantine I think I hit everything you asked for and added your suggested test. Tests are all passing locally now. Please have another look if you would :)

cotedm:
Tests failed due to a timeout in flushing, so appear unrelated. Retest this please.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -130,10 +131,13 @@ static Schema preProcessSchema(Schema schema) {
...
         String keyName = keySchema.name() == null ? keySchema.type().name() : keySchema.name();
         String valueName = valueSchema.name() == null ? valueSchema.type().name() : valueSchema.name();
+        if (keySchema.type() == Schema.Type.STRING) {
+          return SchemaBuilder.map(preProcessSchema(keySchema), preProcessSchema(valueSchema)).build();
+        }
         Schema elementSchema = SchemaBuilder.struct().name(keyName + "-" + valueName)
-            .field(MAP_KEY, preProcessSchema(keySchema))
-            .field(MAP_VALUE, preProcessSchema(valueSchema))
-            .build();
+                .field(MAP_KEY, preProcessSchema(keySchema))
```
kkonstantine: nit: Sorry to return to this, but seems that the previous setting was the correct one re: indentation in this project. Tab needs to correspond to 2 spaces, thus this should be 4 spaces instead of 8 after the change. I wouldn't mind fixing bad indentation but the former was the correct one.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -200,11 +202,19 @@ static Object preProcessValue(Object value, Schema schema, Schema newSchema) {
         }
         return result;
       case MAP:
-        keySchema = schema.keySchema();
-        valueSchema = schema.valueSchema();
-        ArrayList<Struct> mapStructs = new ArrayList<>();
-        Map<?, ?> map = (Map<?, ?>) value;
+        Schema keySchema = schema.keySchema();
         Schema newValueSchema = newSchema.valueSchema();
+        Schema valueSchema = schema.valueSchema();
```
kkonstantine: nit: Since we declare a variable that is used only once below (that is: `Schema valueSchema = schema.valueSchema();`) let's add `Schema newKeySchema = newSchema.keySchema()` as well. 
Also I suggest that they'd be added in an order that feels more consistent (old first then new ones, or key schemas first then value schemas next, again old then new).

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -200,11 +202,19 @@ static Object preProcessValue(Object value, Schema schema, Schema newSchema) {
...
+        Map<?, ?> map = (Map<?, ?>) value;
+        if (keySchema.type() == Schema.Type.STRING) {
+          Map<Object, Object> processedMap = new HashMap<>();
+          for (Map.Entry<?, ?> entry: map.entrySet()) {
+            processedMap.put(preProcessValue(entry.getKey(), keySchema, newSchema.keySchema()),
+                    preProcessValue(entry.getValue(), valueSchema, newValueSchema));
```
kkonstantine: nit: incorrect indentation 8 -> 4

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -200,11 +202,19 @@ static Object preProcessValue(Object value, Schema schema, Schema newSchema) {
...
+          }
+          return processedMap;
+        }
+        ArrayList<Struct> mapStructs = new ArrayList<>();
```
kkonstantine: nit: Ideally it'd be `List<Struct>` (for instance in the test we cast to List anyway). If you choose to fix this, pls fix the declaration of `ArrayList<Object>` in the ARRAY case as well.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
cotedm: so this one we can't actually do because of the fallthrough with the case statement. When I tried adding a `newKeySchema` outside the scope of the string keyed map, I get test failures because of problems looking up a key schema for non-map objects. It's a bit ugly, but I think it's the best we can do. `valueSchema` does end up getting used in the `STRUCT` case as well, so it has multiple uses further down. I flipped the order though because that was nonsense :)

cotedm:
@kkonstantine fixed up the nits minus one because it breaks some tests to do it. If you can take one more pass through (and assuming jenkins tests pass) we can get it merged after that. Thanks!

kkonstantine:
Pretty implicit side-effect from the call on `newSchema.keySchema()`. That's ok for now. Thanks

woodlee:
@cotedm @kkonstantine Is there an open issue or other documentation somewhere of the symptoms of the problem this is solving? I've encountered some behavior (on CP 3.2.2) that I _think_ this will address but it's hard for me to be sure just from looking at the changes.

ewencp:
@woodlee The issue was that we weren't properly special-casing when a Map had string keys. Connect supports flexible maps where keys can be any type because there are some serialization formats that support this. JSON does not, so to handle the general Map type we need to convert to a list of key/value pairs:
```
[ { "key": <complex key type>, "value": <complex value type> }, { "key": <complex key type>, "value": <complex value type> }, ... ]
```

 But in the case we know the keys are all strings, you'd expect to just get a normal JSON object output:
```
{ "string key1": <complex value type>, "string key2": <complex value type>, ... }
```

This PR fixed that so we get the expected output for Maps that have string keys.

woodlee:
@ewencp Thanks for the explanation!

My issue was definitely related... I was seeing behavior on 3.2.2  where a connector would sometimes index string-to-string maps in the first style, and sometimes in the second, despite no difference in the Avro schema between records. I never pinned down the causative conditions precisely but it seemed related to whether I had pre-created the index and associated type mapping in ES before creating the connector. I could try to isolate the conditions and report back if it would be of any value (maybe not, given this PR).

In this particular instance the maps are of HTTP header key:value pairs, so even though the keys are always strings I actually consider the first way of mapping to be preferable so as to avoid a type explosion from dynamic mapping in ES. I'm sure I can tweak the mapping to prevent that once events start coming through in the new way, but just something to consider (e.g., might be worth making this configurable or something).

PR #77, opened 2017-04-07
Add functionality for basic authentication.

combatdud3:
Added config variables for CONNECTION_USERNAME and CONNECTION_PASSWORD
Add check to for CONNECTION_USERNAME being defined, if it is initialise
the JestClient with defaultCredentials, otherwise initialise JestClient
as usual.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -27,6 +27,8 @@ import java.util.Map;
 public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
 
   public static final String CONNECTION_URL_CONFIG = "connection.url";
+  public static final String CONNECTION_USERNAME = "connection.username";
+  public static final String CONNECTION_PASSWORD = "connection.password";
   public static final String BATCH_SIZE_CONFIG = "batch.size";
   public static final String MAX_IN_FLIGHT_REQUESTS_CONFIG = "max.in.flight.requests";
   public static final String MAX_BUFFERED_RECORDS_CONFIG = "max.buffered.records";
@@ -51,6 +53,12 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
           .define(CONNECTION_URL_CONFIG, Type.LIST, Importance.HIGH,
                   "List of Elasticsearch HTTP connection URLs e.g. ``http://eshost1:9200,http://eshost2:9200``.",
                   group, ++order, Width.LONG, "Connection URLs")
+          .define(CONNECTION_USERNAME, Type.STRING, Importance.LOW,
+                  "The username to use for HTTP Basic Authentication.",
+                  group, ++order, Width.LONG, "Connection Username")
+          .define(CONNECTION_PASSWORD, Type.STRING, Importance.LOW,
+                  "The password to use for HTTP Basic Authentication.",
+                  group, ++order, Width.LONG, "Connection Password")
           .define(BATCH_SIZE_CONFIG, Type.INT, 2000, Importance.MEDIUM,
                   "The number of records to process as a batch when writing to Elasticsearch.",
                   group, ++order, Width.SHORT, "Batch Size")
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -76,6 +76,13 @@ public class ElasticsearchSinkTask extends SinkTask {
 
       if (client != null) {
         this.client = client;
+      } else if (config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_USERNAME)) {
+        List<String> address = config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG);
+        String username = config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_USERNAME);
+        String password = config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_PASSWORD);
+        JestClientFactory factory = new JestClientFactory();
+        factory.setHttpClientConfig(new HttpClientConfig.Builder(address).defaultCredentials(username, password).multiThreaded(true).build());
+        this.client = factory.getObject();
       } else {
         List<String> address = config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG);
         JestClientFactory factory = new JestClientFactory();


combatdud3:
Forgive me for any lack of etiquette. I am not used to collaborating on GitHub and I am definitely not a decent Java programmer. But hopefully this small change is up to snuff.

combatdud3:
[clabot:check]

ewencp:
ok to test

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -51,6 +53,12 @@ protected static ConfigDef baseConfigDef() {
           .define(CONNECTION_URL_CONFIG, Type.LIST, Importance.HIGH,
                   "List of Elasticsearch HTTP connection URLs e.g. ``http://eshost1:9200,http://eshost2:9200``.",
                   group, ++order, Width.LONG, "Connection URLs")
+          .define(CONNECTION_USERNAME, Type.STRING, Importance.LOW,
+                  "The username to use for HTTP Basic Authentication.",
+                  group, ++order, Width.LONG, "Connection Username")
+          .define(CONNECTION_PASSWORD, Type.STRING, Importance.LOW,
```
ewencp: Could we change the Type here to `Type.PASSWORD`? This will mask the output when the value is logged so passwords aren't accidentally leaked via logs.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -76,6 +76,13 @@ public void start(Map<String, String> props, JestClient client) {
 
       if (client != null) {
         this.client = client;
+      } else if (config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_USERNAME)) {
```
ewencp: Username is listed as `Type.STRING` in the config, so this should be `config.getString(...)`.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -76,6 +76,13 @@ public void start(Map<String, String> props, JestClient client) {
...
+        List<String> address = config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG);
+        String username = config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_USERNAME);
+        String password = config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_PASSWORD);
+        JestClientFactory factory = new JestClientFactory();
+        factory.setHttpClientConfig(new HttpClientConfig.Builder(address).defaultCredentials(username, password).multiThreaded(true).build());
```
ewencp: This code is basically the same as the final `else` branch except you add in the `defaultCredentials(...)` call. I'd probably just combine the two branches and split the construction of the `HttpClientConfig` across multiple lines, e.g. something like:

```
JestClientFactory factory = new JestClientFactory();
List<String> addresses = config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG);
HttpClientConfig.Builder clientBuilder = new HttpClientConfig.Builder(addresses).multiThreaded(true);
String username = config.getString(ElasticsearchSinkConnectorConfig.CONNECTION_USERNAME);
if (username != null) {
  String password = config.getString(ElasticsearchSinkConnectorConfig.CONNECTION_PASSWORD);
  clientBuilder.defaultCredentials(username, password);
}
factory.setHttpClientConfig(clientBuilder.build());
this.client = factory.getObject();
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -51,6 +53,12 @@ protected static ConfigDef baseConfigDef() {
           .define(CONNECTION_URL_CONFIG, Type.LIST, Importance.HIGH,
                   "List of Elasticsearch HTTP connection URLs e.g. ``http://eshost1:9200,http://eshost2:9200``.",
                   group, ++order, Width.LONG, "Connection URLs")
+          .define(CONNECTION_USERNAME, Type.STRING, Importance.LOW,
+                  "The username to use for HTTP Basic Authentication.",
+                  group, ++order, Width.LONG, "Connection Username")
```
ewencp: Both of these fields should probably be `Width.MEDIUM` -- usernames and passwords aren't all that long compared to, e.g., the list of connection URLs.

combatdud3:
@ewencp Thanks for the feedback! I've made the changes you suggested, but I'm not sure what caused the test to fail. Is it something I can fix?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -51,6 +53,12 @@ protected static ConfigDef baseConfigDef() {
           .define(CONNECTION_URL_CONFIG, Type.LIST, Importance.HIGH,
                   "List of Elasticsearch HTTP connection URLs e.g. ``http://eshost1:9200,http://eshost2:9200``.",
                   group, ++order, Width.LONG, "Connection URLs")
+          .define(CONNECTION_USERNAME, Type.STRING, Importance.LOW,
```
ewencp: Not sure why the `cause` from the exception is getting lost in the test output, but the reason for the error is that both username and password are still required parameters based on this code. This line should be:

```
.define(CONNECTION_USERNAME, Type.STRING, null, Importance.LOW,
```

to define an `null` default -- even if the default is `null`, it still needs to be explicitly specified to indicate there is a default value (and therefore the setting is not required to be specified by the user). Same with the password, which needs a default value to avoid requiring the user to enter a value even if they don't want to use this auth method.

combatdud3: Thanks! I'll make sure they're optional and recommit.

ewencp:
retest this please

combatdud3:
@ewencp I apologise for having to lean on you so much for assistance. I'm not sure what I've done to make the build unstable, do you have any pointers?

Also is there any documentation around the build process so I can try and run the tests without spamming your Jenkins instance, specifically around the connect-api and connect-json SNAPSHOT jars. I've managed to build them from Kafka source, but I can't get the kafka-connect-elasticsearch project to find them during the build process.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -77,9 +77,15 @@ public void start(Map<String, String> props, JestClient client) {
...
         this.client = client;
       } else {
-        List<String> address = config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG);
         JestClientFactory factory = new JestClientFactory();
-        factory.setHttpClientConfig(new HttpClientConfig.Builder(address).multiThreaded(true).build());
+        List<String> addresses = config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG);
+        HttpClientConfig.Builder clientBuilder = new HttpClientConfig.Builder(addresses).multiThreaded(true);
+        String username = config.getString(ElasticsearchSinkConnectorConfig.CONNECTION_USERNAME);
+        if (username != null) {
+          String password = config.getString(ElasticsearchSinkConnectorConfig.CONNECTION_PASSWORD);
```
timd112: FWIW, this patch works for us in v 3.2.1, however we had to change this line to:
`String password = config.getPassword(ElasticsearchSinkConnectorConfig.CONNECTION_PASSWORD).value()`

I'm not sure if anything has changed up stream in master that makes this update unnecessary, but thought I'd mention it.

timd112:
FWIW, this patch works for us in v 3.2.1, however we had to change this line to:
https://github.com/confluentinc/kafka-connect-elasticsearch/pull/77/files#diff-262b1544646777f47990dbf5b68911ddR85

`String password = config.getPassword(ElasticsearchSinkConnectorConfig.CONNECTION_PASSWORD).value()`

I'm not sure if anything has changed up stream in master that makes this update unnecessary, but thought I'd mention it.

cricket007:
Close this?  Auth added in https://github.com/confluentinc/kafka-connect-elasticsearch/commit/dcb664cb78a5a1264c26511b76b02678cc4514e1

rhauch:
@circuitousNerd thanks again for contributing. @rayokota ended up implementing this in a different PR (#215), so I'm going to close this without merging. Feel free to reopen if that doesn't address your needs.

PR #83, opened 2017-05-02
ST-336: Move slf4j-simple into test scope so it is not pulled into the packaged version.

ewencp:


ewencp:
@confluentinc/connect This extra slf4j backend was leaking in and causing unexpected logging behavior (i.e. the Kafka log4j settings were ignored and the simple logger used instead)

ewencp:
Yeah, it wouldn't hurt to do so; in fact, I think this is the only connector using it. For just getting test output, slf4j-simple is kinda handy shorthand if you're ok with the verbosity and no control.

PR #88, opened 2017-05-14
Inherit from common pom

ewencp:


src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -119,120 +134,170 @@ static Schema preProcessSchema(Schema schema) {
...
+
+  private static Object preProcessMapValue(Object value, Schema schema, Schema newSchema) {
+    Schema keySchema = schema.keySchema();
+    Schema valueSchema = schema.valueSchema();
+    Schema newValueSchema = newSchema.valueSchema();
+    Map<?, ?> map = (Map<?, ?>) value;
+    if (keySchema.type() == Schema.Type.STRING) {
+      Map<Object, Object> processedMap = new HashMap<>();
+      for (Map.Entry<?, ?> entry: map.entrySet()) {
+        processedMap.put(preProcessValue(entry.getKey(), keySchema, newSchema.keySchema()),
```
kkonstantine: I thought when splitting long argument lines, it was one argument per line and the closing parenthesis in its own line. Correct me if I'm wrong. (Checkstyle doesn't seem to care).

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -119,120 +134,170 @@ static Schema preProcessSchema(Schema schema) {
...
+    Schema valSchema = preProcessSchema(schema.valueSchema());
+    return copySchemaBasics(schema, SchemaBuilder.array(valSchema)).build();
+  }
+
+  private static Schema preProcessMapSchema(Schema schema) {
+    Schema keySchema = schema.keySchema();
+    Schema valueSchema = schema.valueSchema();
+    String keyName = keySchema.name() == null ? keySchema.type().name() : keySchema.name();
+    String valueName = valueSchema.name() == null
+                       ? valueSchema.type().name()
```
kkonstantine: nit: My impression was that these split on literals would be like indenting consecutive method calls. Aligned at the very start of the line, 2 tabs in (== 4 spaces in this repo's settings). Never mind if that's not the case.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -78,10 +83,18 @@ private static String convertKey(Schema keySchema, Object key) {
...
+      String index,
+      String type,
+      boolean ignoreKey,
+      boolean ignoreSchema
+  ) {
     final String id;
     if (ignoreKey) {
-      id = record.topic() + "+" + String.valueOf((int) record.kafkaPartition()) + "+" + String.valueOf(record.kafkaOffset());
+      id = record.topic()
+           + "+" + String.valueOf((int) record.kafkaPartition())
```
kkonstantine: Same question as below with the alignment of literals in multiple lines (see comment on ternary operator alignment)

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -88,27 +131,22 @@ protected static ConfigDef baseConfigDef() {
       int order = 0;
       configDef
           .define(TYPE_NAME_CONFIG, Type.STRING, Importance.HIGH,
-                  "The Elasticsearch type name to use when indexing.",
+                  TYPE_NAME_DOC,
                   group, ++order, Width.SHORT, "Type Name")
```
kkonstantine: Also a bit surprised that checkstyle doesn't require one argument per line. Maybe we turned this off.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
ewencp: This is just how intellij decided to indent based on the google style I have installed. I'm not picky, but do want to get people using the common IDEs to end up with similar settings. @norwood how have you been handling these cases elsewhere?

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
ewencp: Yeah, this just carried over from existing code that was styled differently, I've reformatted here. Style stuff like this that isn't enforced by checkstyle is just going to take time & a watchful eye to get it all converted. I'm absolutely certain there is other code that wasn't touched by this diff that also isn't formatted with this new style.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
ewencp: Not sure it's even an option with checkstyle. tbh while I generally don't mind 1-per-line when you have to overflow to multiple lines, it can get annoying in these cases where we have a lot of calls with a lot of arguments (though this is also why people want a builder interface for these...)

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
kkonstantine: Yeah, I noticed it's doing the same for me too. I'm fine with it. Not picky either.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
kkonstantine: This is specific to configs and the calls to `define`. I have to admit, for methods that take 4+ arguments it's much easier for me to figure out whether it's the version that accepts a default or not for instance, when it's one arg per line. I wasn't the first to use 1-arg per line but I think is good to catch mismatches/missing arguments. Finally it's easier to reformat with Alt-Cmd-L when intellij is used. In any case, don't feel too strong about it, as with most stylistic comments.

dguy: I think checkstyle just cares about line length. Must admit if i'm going to split args over multiple lines then i'd prefer 1 arg per line as it makes it much easier for me to grok. Subjective, though

ewencp:
retest this please

PR #91, opened 2017-06-02
Drop invalid messages if needed

patsak:


ewencp:
@patsak Are there cases where we expect to fail to convert the data? I'm not sure whether a setting for this makes sense -- if there's a bug in the conversion code, we should just fix that.

patsak:
@ewencp  Producer has written message without key for example - https://github.com/confluentinc/kafka-connect-elasticsearch/blob/master/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java#L56-L58
Or user make a mistake and write messages for other topic.  
In this cases connector stuck on this messages forever.

patsak:
[clabot:check]

patsak:
@ewencp I've retargeted branch with merge conflicts resolving and style fixes.

ewencp:
ok to test

ewencp:
Great, looks good. Thanks for the contribution @patsak!

PR #105, opened 2017-07-27
Update quickstart to use Confluent CLI.

kkonstantine:


docs/elasticsearch_connector.rst
```diff
@@ -27,42 +27,96 @@ Quickstart
...
+    Starting zookeeper
+    zookeeper is [UP]
+    Starting kafka
+    kafka is [UP]
+    Starting schema-registry
+    schema-registry is [UP]
+    Starting kafka-rest
+    kafka-rest is [UP]
+    Starting connect
+    connect is [UP]
```
ewencp: We can get rid of the paragraph right after this since we told them how to start the services.

docs/elasticsearch_connector.rst
```diff
@@ -27,42 +27,96 @@ Quickstart
...
+       "name": "elasticsearch-sink"
+     },
+     "tasks": []
+   }
+
+.. tip::
+
+   This is equivalent to running::
+
+   $ confluent load elasticsearch-sink -d etc/kafka-connect-elasticsearch/quickstart-elasticsearch.properties
```
ewencp: This seems like it's a distracting tip about how to use the CLI that doesn't really fit in with the flow of the quickstart.

docs/elasticsearch_connector.rst
kkonstantine: done

docs/elasticsearch_connector.rst
kkonstantine: I'm looking for a way to show how one can pass a configuration that is not included in the pre-defined ones. It could be useful to people looking for a way to pass a file without editing the one under `etc/`. Anyway, removing because the json input is not demonstrated.

kkonstantine:
Thanks!

PR #111, opened 2017-08-19
Preemptive HTTP Basic auth 

sscarduzio:
The issue with non preemptive authentication is that it has to send first a non-authenticated request (resulting in a 401) and **only then** it sends the credentials across.

Since HTTP is stateless, this makes every connection very slow. I had to update to the latest Jest in order to achieve this.

Feel free to suggest any changes!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,7 +43,7 @@
         <lucene.version>5.5.2</lucene.version>
         <jna.version>4.2.1</jna.version>
         <hamcrest.version>2.0.0.0</hamcrest.version>
-        <jest.version>2.0.0</jest.version>
+        <jest.version>5.3.2</jest.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
     </properties>
 
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchAuth.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchAuth.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright 2016 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ **/
+
+package io.confluent.connect.elasticsearch;
+
+import com.google.common.collect.Sets;
+import io.searchbox.client.config.HttpClientConfig;
+import org.apache.http.HttpHost;
+import org.apache.http.auth.AuthScope;
+import org.apache.http.auth.UsernamePasswordCredentials;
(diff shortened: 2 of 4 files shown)


sscarduzio:
[clabot:check]

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchAuth.java
```diff
@@ -0,0 +1,60 @@
...
+
+    BasicCredentialsProvider credentialsProvider = new BasicCredentialsProvider();
+    for (HttpHost h : targetHosts) {
+      AuthScope as =  new AuthScope(h.getHostName(), h.getPort());
+      UsernamePasswordCredentials cr = new UsernamePasswordCredentials(username, password);
+      credentialsProvider.setCredentials(as,cr);
+    }
+    clientConfig.credentialsProvider(credentialsProvider);
+
+    clientConfig.preemptiveAuthTargetHosts(targetHosts);
```
rhauch: Can I ask why you created a separate class to do this? Couldn't it have been done as a simple static method that adds the authentication details to the `HttpClientConfig.Builder`? So, rather than doing this:

     new ElasticsearchAuth(address, clientConfig, username, password).invoke();

you might instead do the following:

     ElasticsearchAuth.addAuthorizationToConfig(address, clientConfig, username, password);

What would you think about having this static method in `ElasticsearchSinkTask`?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -27,6 +27,14 @@
 public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
 
   public static final String CONNECTION_URL_CONFIG = "connection.url";
+
+  public static final String CONNECTION_USERNAME_CONFIG = "connection.username";
+  public static final String CONNECTION_USERNAME_DOC =
+      "HTTP Basic Auth username (towards Elasticsearch)";
```
rhauch: Maybe "Username used to authenticate with Elasticsearch. The default is null, and authentication will only be performed if a non-null username and password are specified."?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -27,6 +27,14 @@
...
+  public static final String CONNECTION_PASSWORD_CONFIG = "connection.password";
+  public static final String CONNECTION_PASSWORD_DOC =
+      "HTTP Basic Auth password (towards Elasticsearch)";
```
rhauch: Maybe "Password used to authenticate with Elasticsearch. The default is null, and authentication will only be performed if a non-null username and password are specified."?

pom.xml
```diff
@@ -43,7 +43,7 @@
         <lucene.version>5.5.2</lucene.version>
         <jna.version>4.2.1</jna.version>
         <hamcrest.version>2.0.0.0</hamcrest.version>
-        <jest.version>2.0.0</jest.version>
+        <jest.version>5.3.2</jest.version>
```
rhauch: Is this jump necessary to get the authorization functionality in Jest?

cricket007:
Close this?  Auth added in https://github.com/confluentinc/kafka-connect-elasticsearch/commit/dcb664cb78a5a1264c26511b76b02678cc4514e1

rhauch:
@sscarduzio, thanks again for contributing. @rayokota ended up implementing this in a different PR (#215), so I'm going to close this without merging. Feel free to reopen if that doesn't address your needs.

PR #114, opened 2017-09-05

rhauch:

We need at least two configuration parameters to control this approach. Since `retry.backoff.ms` already exists, we can co-opt it to define the minimum time to wait during retries, but we need another configuration property to define the maximum time to wait. (We could expose parameters to control the exponential part of the equation, but that’s unnecessarily complicated.) This new algorithm computes the normal exponential backoff based upon the `retry.backoff.ms` (the minimum value) and the `max.retry.backoff.ms` value, bounding the result to be within these two values, and then choosing a random value within that range.

Note that to maintain backward compatibility, we always wait for exactly the minimum time if it is equal to or exceeds the maximum time to wait. This might happen if an existing connector configuration defines the `retry.backoff.ms` value but doesn’t set the newer `max.retry.backoff.ms`. Note the default value of `max.retry.backoff.ms` is 10 seconds and hopefully larger than most values of `retry.backoff.ms` that might be used.

**This PR should only be merged onto the `3.3.x` branch; see PR #115 for the `3.4.x` and `master` branches.**

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -19,6 +19,7 @@ package io.confluent.connect.elasticsearch;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Importance;
+import org.apache.kafka.common.config.ConfigDef.Range;
 import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
 
@@ -35,7 +36,8 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   public static final String MAX_RETRIES_CONFIG = "max.retries";
   public static final String TYPE_NAME_CONFIG = "type.name";
   public static final String TOPIC_INDEX_MAP_CONFIG = "topic.index.map";
-  public static final String RETRY_BACKOFF_MS_CONFIG = "retry.backoff.ms";
+  public static final String MIN_RETRY_BACKOFF_MS_CONFIG = "retry.backoff.ms";
+  public static final String MAX_RETRY_BACKOFF_MS_CONFIG = "max.retry.backoff.ms";
   public static final String KEY_IGNORE_CONFIG = "key.ignore";
   public static final String TOPIC_KEY_IGNORE_CONFIG = "topic.key.ignore";
   public static final String SCHEMA_IGNORE_CONFIG = "schema.ignore";
@@ -77,10 +79,16 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
                   "The maximum number of retries that are allowed for failed indexing requests. "
                   + "If the retry attempts are exhausted the task will fail.",
                   group, ++order, Width.SHORT, "Max Retries")
-          .define(RETRY_BACKOFF_MS_CONFIG, Type.LONG, 100L, Importance.LOW,
-                  "How long to wait in milliseconds before attempting to retry a failed indexing request. "
-                  + "This avoids retrying in a tight loop under failure scenarios.",
-                  group, ++order, Width.SHORT, "Retry Backoff (ms)");
+          .define(MIN_RETRY_BACKOFF_MS_CONFIG, Type.LONG, 100L, Range.atLeast(0), Importance.LOW,
+                  "The minimum time to wait in milliseconds before attempting to retry a failed indexing request. "
+                  + "Actual time to wait is computed using exponential backoff with jitter within minimum and maximum values. "
+                  + "This avoids retrying in a tight loop under failure scenarios and avoids thundering herd problems with the source.",
+                  group, ++order, Width.SHORT, "Retry Backoff (ms)")
+          .define(MAX_RETRY_BACKOFF_MS_CONFIG, Type.LONG, 10000L, Range.atLeast(0), Importance.LOW,
+                  "The maximum time to wait in milliseconds before attempting to retry a failed indexing request. "
+                  + "Actual time to wait is computed using exponential backoff with jitter within minimum and maximum values. "
+                  + "This avoids retrying in a tight loop under failure scenarios and avoids thundering herd problems with the source.",
+                  group, ++order, Width.SHORT, "Max Retry Backoff (ms)");
     }
 
     {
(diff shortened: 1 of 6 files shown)


src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -365,6 +383,31 @@ private BulkResponse execute() throws Exception {
...
+   * @param minRetryBackoffMs the minimum time to wait before retrying; assumed to be 0 if value is negative
+   * @param maxRetryBackoffMs the maximum amount of time to wait before retrying
+   * @return the time in milliseconds to wait before the next retry attempt, in the range {@code minRetryBackoffMs}
+   * and {@code maxRetryBackoffMs} (inclusive), or {@code minRetryBackoffMs} if {@code minRangeBackoffMs}
+   * is greater than or equal to {@code maxRetryBackoffMs}
+   */
+  protected static long computeRetryWaitTimeInMillis(int retryAttempt, long minRetryBackoffMs, long maxRetryBackoffMs) {
+    if (minRetryBackoffMs >= maxRetryBackoffMs) return minRetryBackoffMs; // this was the original value and likely what they're setting
+    if (minRetryBackoffMs < 0) minRetryBackoffMs = 0;
+    long nextMaxTimeMs = Math.max(minRetryBackoffMs + 1, Math.min(maxRetryBackoffMs, minRetryBackoffMs * 2 ^ retryAttempt));
```
kkonstantine: In java `^` corresponds to XOR and not `Math.pow`. This makes me think we probably need a test to confirm that the actual waiting time corresponds to exponential backoff. 

Since we are raising at a power of 2, I'd suggest bit shifting, such as: `minRetryBackoffMs << retryAttempt;` (haven't tried it, please confirm, we'll be doing something similar in the S3 connector).

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -35,7 +36,8 @@
   public static final String MAX_RETRIES_CONFIG = "max.retries";
   public static final String TYPE_NAME_CONFIG = "type.name";
   public static final String TOPIC_INDEX_MAP_CONFIG = "topic.index.map";
-  public static final String RETRY_BACKOFF_MS_CONFIG = "retry.backoff.ms";
+  public static final String MIN_RETRY_BACKOFF_MS_CONFIG = "retry.backoff.ms";
+  public static final String MAX_RETRY_BACKOFF_MS_CONFIG = "max.retry.backoff.ms";
```
kkonstantine: I feel that we might not strike the best compromise by introducing another config property and also not renaming the old one (wouldn't be great the break compatibility of course). 

Is `max.retry.backoff.ms` all that necessary? The users will have to figure out that this is a global max (including the backoffs with the given number of retries). And maybe that's not obvious or it becomes a bit tedious to get it right. 

Why not stay with the maximum that is implied by the `retry.backoff.ms`, `max.retries` and the fact that the backoff is exponential (at a power of 2)? Also max can't have an adjustable default (relative to what min will be). Thus it's destined to require adjustment every time someone changes min. 

I feel we don't gain much by introducing it. Just a thought

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -341,22 +345,36 @@ private BulkResponse execute() throws Exception {
...
           throw new ConnectException("Bulk request failed: " + bulkRsp.getErrorInfo());
         } catch (Exception e) {
-          if (retriable && remainingRetries > 0) {
-            log.warn("Failed to execute batch {} of {} records, retrying after {} ms", batchId, batch.size(), retryBackoffMs, e);
-            time.sleep(retryBackoffMs);
+          if (retriable && retryAttempt < maxRetries) {
+            ++retryAttempt;
+            long sleepTimeMs = computeRetryWaitTimeInMillis(retryAttempt, minRetryBackoffMs, maxRetryBackoffMs);
+            if (log.isDebugEnabled()) {
+              log.debug("Failed to execute batch {} of {} records with attempt {}/{}, will attempt retry after {} ms",
```
kkonstantine: Feels a bit odd that, depending on the log level, we change what this message is: a WARN or a DEBUG message. Shouldn't it be a WARN always? Should we just print the stack track in a separate debug message?

src/test/java/io/confluent/connect/elasticsearch/bulk/BulkProcessorTest.java
```diff
@@ -268,4 +278,54 @@ public void unretriableErrors() throws InterruptedException {
...
+      long result = BulkProcessor.computeRetryWaitTimeInMillis(retryAttempt, minRetryBackoffMs, maxRetryBackoffMs);
+      assertTrue(result >= maxRetryBackoffMs);
+    }
+  }
+
+  protected void assertComputeRetryInRange(int retryAttempts, long minRetryBackoffMs, long maxRetryBackoffMs) {
+    for (int retryAttempt=1; retryAttempt < retryAttempts; ++retryAttempt) {
+      long result = BulkProcessor.computeRetryWaitTimeInMillis(retryAttempt, minRetryBackoffMs, maxRetryBackoffMs);
+      if (maxRetryBackoffMs > 0) assertTrue(result <= maxRetryBackoffMs);
+      assertTrue(result >= minRetryBackoffMs);
```
kkonstantine: I believe it's somewhere here that we should check that the wait is truly exponential.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
rhauch: Note that currently the log message outputs the stack trace every time the connector fails and then retries. This was an attempt to change the behavior, but it's more complex than it needs to be.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
rhauch: Yeah, not sure what I was thinking. Thank goodness for reviews!

src/test/java/io/confluent/connect/elasticsearch/bulk/BulkProcessorTest.java
```diff
@@ -268,4 +278,54 @@ public void unretriableErrors() throws InterruptedException {
...
+    long minRetryBackoffMs = 100L;
+    long maxRetryBackoffMs = 10L;
+    for (int retryAttempt=1; retryAttempt < 100; ++retryAttempt) {
+      long result = BulkProcessor.computeRetryWaitTimeInMillis(retryAttempt, minRetryBackoffMs, maxRetryBackoffMs);
+      assertTrue(result >= maxRetryBackoffMs);
+    }
+  }
+
+  protected void assertComputeRetryInRange(int retryAttempts, long minRetryBackoffMs, long maxRetryBackoffMs) {
+    for (int retryAttempt=1; retryAttempt < retryAttempts; ++retryAttempt) {
```
kkonstantine: That's the `=` that I would expect checkstyle to catch.

rhauch:
Superseded by #116

PR #115, opened 2017-09-05

rhauch:

We need at least two configuration parameters to control this approach. Since `retry.backoff.ms` already exists, we can co-opt it to define the minimum time to wait during retries, but we need another configuration property to define the maximum time to wait. (We could expose parameters to control the exponential part of the equation, but that’s unnecessarily complicated.) This new algorithm computes the normal exponential backoff based upon the `retry.backoff.ms` (the minimum value) and the `max.retry.backoff.ms` value, bounding the result to be within these two values, and then choosing a random value within that range.

Note that to maintain backward compatibility, we always wait for exactly the minimum time if it is equal to or exceeds the maximum time to wait. This might happen if an existing connector configuration defines the `retry.backoff.ms` value but doesn’t set the newer `max.retry.backoff.ms`. Note the default value of `max.retry.backoff.ms` is 10 seconds and hopefully larger than most values of `retry.backoff.ms` that might be used.

**This PR is for `3.4.x` and `master`; see #114 for the PR for `3.3.x`.**

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -19,6 +19,7 @@ package io.confluent.connect.elasticsearch;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Importance;
+import org.apache.kafka.common.config.ConfigDef.Range;
 import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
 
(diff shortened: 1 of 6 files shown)


rhauch:
I'll address your comments in both PRs. Unfortunately, a commit in 3.4.x reformatted the bulk of the `ElasticsearchSinkConnectorConfig.java`, so a single PR cannot be cleanly merged onto the 3.4.x and 3.3.x branches.

kkonstantine:
Yes, these are the changes introduced by checkstyle. Up to you. I had to adjust during rebasing in the past, but I've noticed that if the new changes are done with the new style in mind, maybe it's not too bad. The problem with the other PR is that the new additions do not conform the the new style (or that's what I thought while reviewing).

rhauch:
I can backport the formatting changes, and then try to come up with a single PR. BTW, the other PR passes the checkstyle validation.

kkonstantine:
Whatever is easier for you. Which checkstyle validation is passing? There are no rich checkstyle rules (if any) in 3.3.x. I think I noticed a case of `var=val` that if checkstyle works it should have warned about the absence of white space around `=`. (But this is also in `src/test`).

rhauch:
Superseded by #116

PR #116, opened 2017-09-06

rhauch:

This new algorithm computes the normal maximum time to wait for a particular retry attempt using exponential backoff and then choosing a random value less than that maximum time.

Since this exponential algorithm breaks down after a large number of retry attempts, rather than adding a constraint for `max.retries` this change simply uses a practical (and arbitrary) absolute upper limit on the backoff time of 24 hours, and it logs a warning if this upper limit is exceeded.

### DIFF
--- a/docs/configuration_options.rst
+++ b/docs/configuration_options.rst
@@ -55,7 +55,7 @@ Connector
   * Importance: low
 
 ``retry.backoff.ms``
-  How long to wait in milliseconds before attempting to retry a failed indexing request. This avoids retrying in a tight loop under failure scenarios.
+  How long to wait in milliseconds before attempting to retry a the first failed indexing request. This connector uses exponential backoff with jitter, which means that upon additional failures, this connector may wait up to twice as long as the previous wait, up to the maximum number of retries. This avoids retrying in a tight loop under failure scenarios.
 
   * Type: long
   * Default: 100
--- a/docs/elasticsearch_connector.rst
+++ b/docs/elasticsearch_connector.rst
@@ -241,6 +241,63 @@ The following change is not allowed:
 
 As mappings are more flexible, schema compatibility should be enforced when writing data to Kafka.
 
+Automatic Retries
+-----------------
+The Elasticsearch connector may experience problems writing to the Elasticsearch endpoint, such as when
+the Elasticsearch service is temporarily overloaded. In many cases, the connector will retry the request
+a number of times before failing. To prevent from further overloading the Elasticsearch service, the connector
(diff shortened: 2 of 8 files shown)


rhauch:
This supersedes the previous PRs. Note that this will not merge cleanly onto the `3.4.x` branch.

rhauch:
As referenced in the JavaDocs of the new `RetryUtil`, the basic logic is similar to the "full jitter" approach described in [this AWS blog post](https://www.awsarchitectureblog.com/2015/03/backoff.html). That post also describes two other jitter approaches, and how they do perform differently. Ironically, the simple "full jitter" case performed quite well and is simple and intuitive. And, as the post shows, using "full jitter" completes the work in significantly less time and with fewer total number of requests (less than half) compared to simple exponential backoff. Adding jitter to exponential backoff is clearly better than exponential backoff, but there are different ways of adding jitter and the benefits of more complex jitter approaches do not seem to outweigh the additional complexity.

So, I'm disinclined to use anything more complicated than the jitter approach used in this PR, especially if there has been no testing or comparison of the approach.

rhauch:
The current logic reuses the `retry.backoff.ms` as an initial delay, effectively computing the delay before retrying as

    random(retryBackoffMs, exponentialBackoffMs)

One variation I did consider was using 0 as the lower bound for the random calculation rather than the retry backoff time. IOW,

    random(0, exponentialBackoffMs)

The upper limit would still be the exponential backoff calculated using the `retry.backoff.ms`.

Given the use of jitter, I'm not sure the non-zero backoff is necessary, and in fact it may be counter-productive as it might artificially and unnecessarily increase the delay beyond what is needed by the ES backend. Using zero for the lower bound allows the retry logic to _sometimes_ wait for less than the `retry.backoff.ms` minimum. The big advantage of using 0 for a lower bound is that it's a bit easier to understand and a bit easier to control the backoff logic since the `retry.backoff.ms` is used only in one way rather than two.

rhauch:
I've implemented the 0 lower bound as mentioned above, and this definitely results in a more even distribution of retry backoff times. For example, when using a lower bound of `retry.backoff.ms`, the backoff time before the first retry always happened to be _a constant value_. Obviously if this happens to many different tasks at once, it might still create a thundering herd, at least for the first retry. 

However, using 0 as the lower bound means that the backoff time for the first retry is a random number in the range `[0, ${retry.backoff.ms}]`, which means even the first retry for many tasks is more evenly distributed. (Yes, many of these might still fail if the ES backend is still overloaded, but some may some may succeed and distributing the retries might prevent the ES backend from experiencing a sudden cluster of requests at the same time.)

A good improvement, IMO.

PR #126, opened 2017-10-09
CC-1191: Added feature flag to control string-keyed map entry JSON serialization

rhauch:
Prior to 3.3.0, all map entries were serialized into JSON using a less compact form like the following that worked for all types of keys:

```
"metadata" : [ 
{ 
"key" : "lastUpdated", 
"value" : "20170209-23:58:26.420" 
}, { 
"key": "isSoftDeleted", 
"value": "false" 
}
...
]
```

In 3.3.0 this was changed so that the entries of all maps with string keys were written out in a more compact and natural JSON representation:

```
"metadata" : { 
   "lastUpdated": "Tue Feb 07 20:57:05 GMT 2017", 
   "isSoftDeleted": "false", 
   ...
}
```

Now, a new `compact.map.entries` configuration property allows users to control this behavior. This configuration defaults to `true` to match the newer 3.3.0 behavior for the entries of maps with string keys. However, simply set `compact.map.entries=false` for the older, pre-3.3.0 JSON representation.

### DIFF
--- a/docs/configuration_options.rst
+++ b/docs/configuration_options.rst
@@ -88,6 +88,13 @@ Data Conversion
   * Default: false
   * Importance: low
 
+``compact.map.entries``
+  Defines how map entries with string keys within record values should be written to JSON. When this is set to ``true``, these entries are written compactly as ``"entryKey": "entryValue"``. Otherwise, map entries with string keys are written as a nested document ``{"key": "entryKey", "value": "entryValue"}``. All map entries with non-string keys are always written as nested documents. Prior to 3.3.0, this connector always wrote map entries as nested documents, so set this to ``false`` to use that older behavior.
+
+  * Type: boolean
+  * Default: true
+  * Importance: low
+
 ``topic.index.map``
   A map from Kafka topic name to the destination Elasticsearch index, represented as a list of ``topic:index`` pairs.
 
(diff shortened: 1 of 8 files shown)


src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -154,7 +160,7 @@ static Schema preProcessSchema(Schema schema) {
     }
   }
 
-  private static SchemaBuilder copySchemaBasics(Schema source, SchemaBuilder target) {
+  SchemaBuilder copySchemaBasics(Schema source, SchemaBuilder target) {
```
kkonstantine: nit: Leave a comment as below that is `//visible for testing`?

Also, up to you if you want to make it `protected` instead. (this one and the other package private method used for testing below).

docs/configuration_options.rst
```diff
@@ -88,6 +88,16 @@ Data Conversion
   * Default: false
   * Importance: low
 
+``compact.map.entries``
+  Defines how map entries with string keys within record values should be written to JSON. When this is set to ``true``, these entries are written compactly as ``\"entryKey\": \"entryValue\"``.
```
kkonstantine: in the `.rst` the double quotes `"` need to be unescaped. (same as when you run `ElasticsearchSinkConnectorConfig.main`

docs/configuration_options.rst
```diff
@@ -88,6 +88,16 @@ Data Conversion
...
+  Otherwise, map entries with string keys are written as a nested document ``{\"key\": \"entryKey\", \"value\": \"entryValue\"}``.
```
kkonstantine: same as above

docs/configuration_options.rst
rhauch: ack

docs/configuration_options.rst
rhauch: ack

rhauch:
@kkonstantine Thanks for reviewing. Latest commits address your concerns with very minor fixes.

PR #127, opened 2017-10-09
CC-1191: Added feature flag to control string-keyed map entry JSON serialization

rhauch:
Prior to 3.3.0, all map entries were serialized into JSON using a less compact form like the following that worked for all types of keys:

```
"metadata" : [ 
{ 
"key" : "lastUpdated", 
"value" : "20170209-23:58:26.420" 
}, { 
"key": "isSoftDeleted", 
"value": "false" 
}
...
]
```

In 3.3.0 this was changed so that the entries of all maps with string keys were written out in a more compact and natural JSON representation:

```
"metadata" : { 
   "lastUpdated": "Tue Feb 07 20:57:05 GMT 2017", 
   "isSoftDeleted": "false", 
   ...
}
```

Now, a new `compact.map.entries` configuration property allows users to control this behavior. This configuration defaults to `true` to match the newer 3.3.0 behavior for the entries of maps with string keys. However, simply set `compact.map.entries=false` for the older, pre-3.3.0 JSON representation.

See #126 for the equivalent PR for the earlier `3.3.x` branch.

### DIFF
--- a/docs/configuration_options.rst
+++ b/docs/configuration_options.rst
@@ -88,6 +88,13 @@ Data Conversion
   * Default: false
   * Importance: low
 
+``compact.map.entries``
+  Defines how map entries with string keys within record values should be written to JSON. When this is set to ``true``, these entries are written compactly as ``"entryKey": "entryValue"``. Otherwise, map entries with string keys are written as a nested document ``{"key": "entryKey", "value": "entryValue"}``. All map entries with non-string keys are always written as nested documents. Prior to 3.3.0, this connector always wrote map entries as nested documents, so set this to ``false`` to use that older behavior.
+
+  * Type: boolean
+  * Default: true
+  * Importance: low
+
 ``topic.index.map``
   A map from Kafka topic name to the destination Elasticsearch index, represented as a list of ``topic:index`` pairs.
 
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -52,7 +52,13 @@ public class DataConverter {
     JSON_CONVERTER.configure(Collections.singletonMap("schemas.enable", "false"), false);
   }
 
-  private static String convertKey(Schema keySchema, Object key) {
+  private final boolean useCompactMapEntries;
+
+  public DataConverter(boolean useCompactMapEntries) {
+    this.useCompactMapEntries = useCompactMapEntries;
+  }
+
+  private String convertKey(Schema keySchema, Object key) {
     if (key == null) {
       throw new ConnectException("Key is used as document id and can not be null.");
     }
@@ -83,7 +89,7 @@ public class DataConverter {
     }
   }
 
-  public static IndexableRecord convertRecord(
+  public IndexableRecord convertRecord(
       SinkRecord record,
       String index,
       String type,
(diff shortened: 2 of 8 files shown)


src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -45,6 +45,7 @@
 
 public class DataConverter {
 
+  static boolean useCompactMapEntries = true;
```
kkonstantine: So this practically means we can't have 2 connectors operating in different modes. 
And, in general, fully static classes that start to become richer in logic could potentially become a problem, both in testing and regular usage. 

This class exposes only one public method (`convertRecord`). I wonder whether it's worth turning it into non-static and instantiating based on the config or punting this issue for the near future. Up to you.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -59,40 +59,43 @@ public void start(Map<String, String> props, JestClient client) {
       log.info("Starting ElasticsearchSinkTask.");
 
       ElasticsearchSinkConnectorConfig config = new ElasticsearchSinkConnectorConfig(props);
-      String type = config.getString(ElasticsearchSinkConnectorConfig.TYPE_NAME_CONFIG);
-      boolean ignoreKey =
+      final String type = config.getString(ElasticsearchSinkConnectorConfig.TYPE_NAME_CONFIG);
```
kkonstantine: That's a considerable shift to using `final` keyword for local variables. And although this might be a lost battle, I think a discussion would be nice to take place regarding this pattern. Note: I strongly disagree with this pattern.

Any particular reason this is applied here, in this particular PR? Since it's such a massive change, probably belongs in a separate refactoring PR, if it has to be applied, anyways.

docs/configuration_options.rst
```diff
@@ -88,6 +88,16 @@ Data Conversion
   * Default: false
   * Importance: low
 
+``compact.map.entries``
+  Defines how map entries with string keys within record values should be written to JSON. When this is set to ``true``, these entries are written compactly as ``\"entryKey\": \"entryValue\"``.
+  Otherwise, map entries with string keys are written as a nested document ``{\"key\": \"entryKey\", \"value\": \"entryValue\"}``.
```
kkonstantine: Is escaping of `"` intentional in the text too? I guess it is, but I'm checking.

rhauch: yes, it is intentional.

docs/configuration_options.rst
```diff
@@ -88,6 +88,16 @@ Data Conversion
...
+  All map entries with non-string keys are always written as nested documents.
+  Prior to 3.3.0, this connector always wrote map entries as nested documents, so set this to ``false`` to use that older behavior.
```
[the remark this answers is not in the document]
rhauch: The default is true.

docs/configuration_options.rst
```diff
@@ -88,6 +88,16 @@ Data Conversion
...
+
+  * Type: boolean
+  * Default: false
```
kkonstantine: then this should change I guess

kkonstantine: To be safe, to get the docs match exactly what's in the code for configs (new and old) you might choose to run the `main` method on the config.

rhauch:
@kkonstantine latest commits address your concerns, and this reflects a direct port (with reformatting of the changed code) of #126. However, it will probably be easier to apply to just merge and fix #126. Closing without merging.

PR #129, opened 2017-10-10
Check if alias exists if a index not found

### LINKED ISSUE #64: Index Creation Attempted When Using Existing Index Via Alias
I want to use an alias to insert into elastic using the pattern described here:

https://www.elastic.co/blog/managing-time-based-indices-efficiently

However I'm finding that the connector fails on startup because it is attempting to create an index with the alias name. 

On startup, the connector checks for the existence of an index but it should also check whether an alias with the same name already exists before attempting to create a new index

liketic:
Closes #64 

I have to dump `jest-common` to `5.3.2` to use it's `AliasExists`.


livelace:
Bump

Issue #132, opened 2017-10-15
Byte type default value doesn't supported by ElasticSearch


vbeskrovnov:
When we tried to insert `null_value` for binary datatype we get the following message from ElasticSearch: 
```
{
  "error": {
    "root_cause": [
      {
        "type": "mapper_parsing_exception",
        "reason": "Mapping definition for [bytes] has unsupported parameters:  [null_value : AQ==]"
      }
    ],
    "type": "mapper_parsing_exception",
    "reason": "Mapping definition for [bytes] has unsupported parameters:  [null_value : AQ==]"
  },
  "status": 400
}
```
It happens because of binary type in ElasticSearch has no `null_value` parameter: https://www.elastic.co/guide/en/elasticsearch/reference/current/binary.html
So, we should not set default value to this type here:
`io/confluent/connect/elasticsearch/Mapping.java:180`
```
case ElasticsearchSinkConnectorConstants.BINARY_TYPE:
  defaultValueNode = JsonNodeFactory.instance.binaryNode(bytes(defaultValue));
  break;
```

purbon:
fixed by the merge of #302

PR #133, opened 2017-10-15
Byte type default value doesn't supported by ElasticSearch

vbeskrovnov:
Problem described in issue #132. Here it was fixed and some tests for this problem were added.


vbeskrovnov:
[clabot:check]

vbeskrovnov:
[clabot:check]

PR #135, opened 2017-10-17
CC-6599: java.lang.ClassCastException: java.util.Date cannot be cast to java.lang.Long while setting default date value

### LINKED ISSUE #134: java.lang.ClassCastException: java.util.Date cannot be cast to java.lang.Long while setting default date value
If we want to set default value for date field in schema, we should use following for example:
`Date.builder().defaultValue(new java.util.Date()).build()`
We can pass only  `java.util.Date` in `Date.builder().defaultValue()`, because there is validating in `org.apache.kafka.connect.data.ConnectSchema#validateValue(java.lang.String, org.apache.kafka.connect.data.Schema, java.lang.Object)`, where we compare value type with expecting type(In this case - `java.util.Date`).
But when json is created here `io/confluent/connect/elasticsearch/Mapping.java:182`
```
case ElasticsearchSinkConnectorConstants.DATE_TYPE:
  defaultValueNode = JsonNodeFactory.instance.numberNode((long) defaultValue);
  break;
```
This date value tried to cast to long and failed with 
`java.lang.ClassCastException: java.util.Date cannot be cast to java.lang.Long`
So, we should do something like that:
`defaultValueNode = JsonNodeFactory.instance.numberNode(((java.util.Date) defaultValue).getTime());`

vbeskrovnov:
This closes #134 
More info can be found in issue.


src/test/java/io/confluent/connect/elasticsearch/MappingTest.java
```diff
@@ -75,6 +75,24 @@ public void testInferMapping() throws Exception {
...
+        .name("record")
+        .field("foo", Date.builder().defaultValue(expected).build())
+        .build();
+
+    JsonNode dateMapping = Mapping.inferMapping(client, dateSchema);
+    assertNotNull(dateMapping.get("properties").get("foo").get("null_value"));
+    assertEquals(
+        expected.getTime(),
+        dateMapping.get("properties").get("foo").get("null_value").asLong()
+    );
```
wicknicks: add a test for long default values too?

levzem: so default value for a Date can never be a Long - SchemaBuilder will throw an exception...

wicknicks: hmm, I meant to test these two lines: https://github.com/confluentinc/kafka-connect-elasticsearch/pull/135/files/e925cebfd15c8503639a1ec38a5a1fefec20ce7e#diff-9525ecd31fdc68633378d58447bea186R212-R213. if they can never be triggered, then we should not have those lines.

levzem: just future proofing :)

PR #151, opened 2017-11-20
CC-190: Deprecate the topic.index.map configuration option

wicknicks:
Signed-off-by: Arjun Satish <arjun@confluent.io>

### DIFF
--- a/docs/configuration_options.rst
+++ b/docs/configuration_options.rst
@@ -96,7 +96,7 @@ Data Conversion
   * Importance: low
 
 ``topic.index.map``
-  This option is now deprecated. A future version may remove it completely. Please use single message transforms to map topic names to index names.
+  This option is now deprecated. A future version may remove it completely. Please use single message transforms, such as RegexRouter, to map topic names to index names.
 
   A map from Kafka topic name to the destination Elasticsearch index, represented as a list of ``topic:index`` pairs.
 
--- a/docs/configuration_options.rst
+++ b/docs/configuration_options.rst
@@ -96,6 +96,8 @@ Data Conversion
   * Importance: low
 
 ``topic.index.map``
+  This option is now deprecated. A future version may remove it completely. Please use single message transforms to map topic names to index names.
+
   A map from Kafka topic name to the destination Elasticsearch index, represented as a list of ``topic:index`` pairs.
 
   * Type: list
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -71,10 +71,7 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   public static final String TYPE_NAME_CONFIG = "type.name";
   private static final String TYPE_NAME_DOC = "The Elasticsearch type name to use when indexing.";
 
-  /**
-   * @deprecated This option is now deprecated. A future version may remove it completely. Please
-   *             use single message transforms to map topic names to index names.
-   */
+  @Deprecated
   public static final String TOPIC_INDEX_MAP_CONFIG = "topic.index.map";
   private static final String TOPIC_INDEX_MAP_DOC =
       "A map from Kafka topic name to the destination Elasticsearch index, represented as a list "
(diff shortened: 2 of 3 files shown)


docs/configuration_options.rst
```diff
@@ -96,6 +96,8 @@ Data Conversion
   * Importance: low
 
 ``topic.index.map``
+  This option is now deprecated. A future version may remove it completely. Please use single message transforms to map topic names to index names.
```
kkonstantine: nit: _A future version will remove it completely_?

kkonstantine: nit: should we also offer some guidance w.r.t appropriate SMTs?
_Please use single message transforms, such as RegexRouter or others, to map topic names ..._

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -70,6 +70,11 @@
 
   public static final String TYPE_NAME_CONFIG = "type.name";
   private static final String TYPE_NAME_DOC = "The Elasticsearch type name to use when indexing.";
+
+  /**
```
kkonstantine: I believe we don't use javadoc style comments with config variables. Correct me if I'm wrong.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -46,6 +46,9 @@
   private final Set<String> ignoreKeyTopics;
   private final boolean ignoreSchema;
   private final Set<String> ignoreSchemaTopics;
+  /**
```
kkonstantine: Same comment as above.

docs/configuration_options.rst
wicknicks: Yeah, we can add the guidance. RegexRouter is the way to go. 

Yes, we should be removing this option at a later, as RegexRouter subsumes all functionality of this configuration. Unless, there is a reason to deprecate something and leave it there.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
wicknicks: This is mostly for reference. I suppose we should simply use `//` here

kkonstantine: Indeed, we use `//` everywhere besides javadoc (and copyright header)

docs/configuration_options.rst
kkonstantine: I feel that's a bit contradictory. Ones deprecates an option or a feature to signal that it will permanently be removed in the future for sure. If there's ambiguity regarding removal, then it shouldn't be marked as deprecated option.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
wicknicks: An addition to my previous comment: The `@deprecated` annotation applies to fields as well (https://docs.oracle.com/javase/tutorial/java/annotations/predefined.html). Also, if we use Javadoc, the java compiler can report a warning when we use these fields in the code.

kkonstantine: My point was that we skip javadocs for all the configs, it wouldn't make much sense to add it only for the deprecated ones. 

If you want to include the warning during build, feel free to add the `@Deprecated` annotation to the variable (outside a comment - javadoc or regular).

rhauch: We should use the `@Deprecated` annotation at least, and optionally the JavaDoc tag. In this case, I think the annotation would be sufficient.

wicknicks: Got it. Removed the javadoc, and replaced it with a `@Deprecated` annotation.

docs/configuration_options.rst
```diff
@@ -96,6 +96,8 @@ Data Conversion
   * Importance: low
 
 ``topic.index.map``
+  This option is now deprecated. A future version may remove it completely. Please use single message transforms, such as RegexRouter, to map topic names to index names.
+
```
rhauch: The goal is that these .rst files can be largely generated by running `ElasticsearchSinkConnectorConfig.main(...)` and replacing nearly all of the file with the output (except the first few lines). To do that, we'd have to add this text to the documentation in the code.

wicknicks: Thanks, @rhauch. Updated these lines using the main method.

PR #156, opened 2017-12-15
NCS-151. specify routing field

echo-xu:
@echo-xu @krzysztof-everbridge @timxi @mveitas

This pull request enables us to specify which field to be used as routing via ""routing.field.name" (it must be a leaf field, primitive, wrapper primitive or string type) config when initiate the connector in Kafka Connect. 

Note,
1. this is not a general purpose implementation thus is not ready to be contributed to the official repo, it takes more time to get deeper understanding on the schema usage to make it more generalized.
2. the official repo uses 2 spaces indentation, and it forces style check, I just followed that

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,1 @@
+@echo-xu @krzysztof-everbridge @timxi @mveitas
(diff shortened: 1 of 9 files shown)


echo-xu:
I just closed it due to the reason I mentioned above, this is created by mistake.

frankkoornstra:
Could we do something similar but then use one of the `SinkRecord` headers instead? That way we're not polluting the record and it gives flexibility to the user.

PR #158, opened 2017-12-27
CC-1372: Add configuration options for read and connect timeouts

wicknicks:
Signed-off-by: Arjun Satish <arjun@confluent.io>

### DIFF
--- a/docs/configuration_options.rst
+++ b/docs/configuration_options.rst
@@ -61,15 +61,15 @@ Connector
   * Default: 100
   * Importance: low
 
-``conn.timeout.ms``
-  How long to wait (in milliseconds) when establishing a connection to the Elasticsearch server.
+``connection.timeout.ms``
+  How long to wait in milliseconds when establishing a connection to the Elasticsearch server. The task fails if the client fails to connect to the server in this interval, and will need to be restarted.
 
   * Type: int
   * Default: 1000
   * Importance: low
 
 ``read.timeout.ms``
-  How long to wait (in milliseconds) for the Elasticsearch server to send responses.
+  How long to wait in milliseconds for the Elasticsearch server to send a response. The task fails if any read operation times out, and will need to be restarted to resume further operations.
 
   * Type: int
   * Default: 3000
--- a/docs/configuration_options.rst
+++ b/docs/configuration_options.rst
@@ -61,6 +61,20 @@ Connector
   * Default: 100
   * Importance: low
 
+``conn.timeout.ms``
+  How long to wait (in milliseconds) when establishing a connection to the Elasticsearch server.
+
+  * Type: int
+  * Default: 1000
+  * Importance: low
+
+``read.timeout.ms``
+  How long to wait (in milliseconds) for the Elasticsearch server to send responses.
+
+  * Type: int
+  * Default: 3000
+  * Importance: low
+
 Data Conversion
 ^^^^^^^^^^^^^^^
 
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -40,7 +40,7 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   public static final String TOPIC_KEY_IGNORE_CONFIG = "topic.key.ignore";
   public static final String SCHEMA_IGNORE_CONFIG = "schema.ignore";
   public static final String TOPIC_SCHEMA_IGNORE_CONFIG = "topic.schema.ignore";
-  public static final String CONNECTION_TIMEOUT_MS_CONFIG = "conn.timeout.ms";
+  public static final String CONNECTION_TIMEOUT_MS_CONFIG = "connection.timeout.ms";
   public static final String READ_TIMEOUT_MS_CONFIG = "read.timeout.ms";
 
   protected static ConfigDef baseConfigDef() {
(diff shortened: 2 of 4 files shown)


docs/configuration_options.rst
```diff
@@ -61,6 +61,20 @@ Connector
...
+  How long to wait (in milliseconds) when establishing a connection to the Elasticsearch server.
+
+  * Type: int
+  * Default: 1000
+  * Importance: low
+
+``read.timeout.ms``
+  How long to wait (in milliseconds) for the Elasticsearch server to send responses.
+
+  * Type: int
```
mikkin: should be long

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -40,6 +40,8 @@
   public static final String TOPIC_KEY_IGNORE_CONFIG = "topic.key.ignore";
   public static final String SCHEMA_IGNORE_CONFIG = "schema.ignore";
   public static final String TOPIC_SCHEMA_IGNORE_CONFIG = "topic.schema.ignore";
+  public static final String CONNECTION_TIMEOUT_CONFIG = "conn.timeout.ms";
```
mikkin: Recommend changing the variable name to `CONNECTION_TIMEOUT_MS_CONFIG` to maintain consistent naming pattern.

docs/configuration_options.rst
wicknicks: The internal client only accepts `int`s for these configs. Didn't want to cast `long`s to `int`s internally.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
wicknicks: good point. changed. thanks!

docs/configuration_options.rst
mikkin: I see. Yeah, we don't want to cast for sure. Thanks for making other naming fixes. Your changes look good to me!

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -80,7 +82,13 @@ protected static ConfigDef baseConfigDef() {
           .define(RETRY_BACKOFF_MS_CONFIG, Type.LONG, 100L, Importance.LOW,
                   "How long to wait in milliseconds before attempting to retry a failed indexing request. "
                   + "This avoids retrying in a tight loop under failure scenarios.",
-                  group, ++order, Width.SHORT, "Retry Backoff (ms)");
+                  group, ++order, Width.SHORT, "Retry Backoff (ms)")
+          .define(CONNECTION_TIMEOUT_MS_CONFIG, Type.INT, 1000, Importance.LOW, "How long to wait "
+                  + "(in milliseconds) when establishing a connection to the Elasticsearch server.",
+                  group, ++order, Width.SHORT, "Connection Timeout")
+          .define(READ_TIMEOUT_MS_CONFIG, Type.INT, 3000, Importance.LOW, "How long to wait (in "
+                  + "milliseconds) for the Elasticsearch server to send responses.",
```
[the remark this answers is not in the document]
wicknicks: Updated the docs to say what happens. Thanks, @rhauch.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfigTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ElasticsearchSinkConnectorConfigTest {
+
+  private Map<String, String> createProps() {
```
kkonstantine: Where we have used this pattern of manually calling `createProps` and `setUp` in other connectors without using the `@Before` junit annotation, we've done so because some properties in only a few test cases need to be set/customized before the `setUp` method is called. 

If there's no such requirement in a test class, I'd suggest we stick with the standard way of using the `@Before` annotation for initializations.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -73,13 +73,18 @@ public void start(Map<String, String> props, JestClient client) {
...
       if (client != null) {
         this.client = client;
       } else {
         List<String> address = config.getList(ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG);
         JestClientFactory factory = new JestClientFactory();
-        factory.setHttpClientConfig(new HttpClientConfig.Builder(address).multiThreaded(true).build());
+        factory.setHttpClientConfig(new HttpClientConfig.Builder(address)
+            .connTimeout(connTimeout)
+            .readTimeout(readTimeout)
+            .multiThreaded(true).build());
```
kkonstantine: nit: Since I left another comment, let me notice here that you probably want a method call per line on a multi-line call chain. Not big deal though.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfigTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ElasticsearchSinkConnectorConfigTest {
+
+  private Map<String, String> props;
+
+  @Before
+  public void createProps() {
```
rhauch: Minor nit: maybe name this method `setup()`?

wicknicks: Thanks guys. This is done.

PR #163, opened 2018-01-10
CC-350/CC-1097: Support optional deletion for sink records with null values

C0urante:
Adds a new configuration property named "behavior.on.null.value" (not set in stone, suggestions welcome), with three possible values:

**fail:** Throw an (informative, NOT null-pointer) exception
**ignore** (will be the new default): Skip over any messages with null values and non-null keys, logging (at debug level) that a message was skipped over.
**delete:** Delete the corresponding record in ES

Partially follows the lead of https://github.com/confluentinc/kafka-connect-elasticsearch/pull/92/files for some changes.


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -132,7 +133,10 @@ protected static ConfigDef baseConfigDef() {
                   group, ++order, Width.LONG, "Topics for 'Ignore Key' mode")
           .define(TOPIC_SCHEMA_IGNORE_CONFIG, Type.LIST, "", Importance.LOW,
                   "List of topics for which ``" + SCHEMA_IGNORE_CONFIG + "`` should be ``true``.",
-                  group, ++order, Width.LONG, "Topics for 'Ignore Schema' mode");
+                  group, ++order, Width.LONG, "Topics for 'Ignore Schema' mode")
+          .define(BEHAVIOR_ON_NULL_VALUES_CONFIG, Type.STRING, "ignore", ConfigDef.ValidString.in("ignore", "delete", "fail"), Importance.LOW,
```
mageshn: You could get the default string from Enum instead of hardcoding. The Validator list can also be constructed in enum and used here.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -242,4 +276,12 @@ Object preProcessValue(Object value, Schema schema, Schema newSchema) {
         return value;
     }
   }
+
+  public enum BehaviorOnNullValues {
```
mageshn: not : May be sounds better as NullValueBehavior or NullValueHandlingStrategy

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -96,13 +128,15 @@ public IndexableRecord convertRecord(SinkRecord record, String index, String typ
...
       schema = preProcessSchema(record.valueSchema());
-      value = preProcessValue(record.value(), record.valueSchema(), schema);
+      value = record.value() == null ? null : preProcessValue(record.value(), record.valueSchema(), schema);
     } else {
       schema = record.valueSchema();
       value = record.value();
     }
 
-    final String payload = new String(JSON_CONVERTER.fromConnectData(record.topic(), schema, value), StandardCharsets.UTF_8);
+    final String payload = value == null
```
mageshn: nit : formatting looks a little weird for readability. At first look, i read it as payload = value :)

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
C0urante: Ahhh, nifty! Will definitely change that.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: Agreed. Have refactored to `NullValueBehavior`. Intentionally leaving the original name for `BEHAVIOR_ON_NULL_VALUES_CONF` though, since it seems to better match the idiom set up by other config name variables.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: Fair enough--will replace the couple of ternary operators with vanilla if/else statements.

mageshn: ternary is fine. May be move it to next line or use parentheses or both

C0urante: Alright, will do.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
```diff
@@ -324,7 +431,9 @@ private ElasticsearchWriter initWriter(JestClient client, boolean ignoreKey, Set
         .setLingerMs(1000)
         .setRetryBackoffMs(1000)
         .setMaxRetry(3)
+        .setBehaviorOnNullValues(behavior)
         .build();
+    behaviorOnNullValues = behavior;
```
kkonstantine: This doesn't do what we want. It's set after it's used in base class's `setUp`. That's the only place where `getBehaviorOnNullValues` and setting it here it's too late. 

Which also means that the test for the non-default options (delete and fail) is not testing what we want? (can't test thoroughly now, intellij is freaking out with jvm versions)

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
kkonstantine: Not sure I agree. Use ternary operators liberally.

kkonstantine: Also, please don't use unneeded parentheses for readability.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
kkonstantine: I'm more comfortable with the enum matching the config. I'd prefer `BehaviorOnNullValues`. Also the important word is first in this version. `NullValuesBehaviour` kinda reads slowly to the actual meaning of the class. Up to you.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
```diff
@@ -298,6 +303,96 @@ public void testBytes() throws Exception {
     verifySearchResults(records, ignoreKey, ignoreSchema);
   }
 
+  @Test
+  public void testIgnoreNullValue() throws Exception {
+    final boolean ignoreKey = false;
```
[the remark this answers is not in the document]
C0urante: At least with `ignoreKey` and `ignoreSchema`, all the other tests have those local variables declared final.

src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
```diff
@@ -239,4 +245,63 @@ public void struct() {
     );
   }
 
+  @Test
+  public void ignoreNull() {
+    final String key = "key";
```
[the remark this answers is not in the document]
C0urante: Not following the static class suggestion. What would it contain and/or do?

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -242,4 +276,12 @@ Object preProcessValue(Object value, Schema schema, Schema newSchema) {
...
     }
   }
+
+  public enum BehaviorOnNullValues {
+    IGNORE,
+    DELETE,
+    FAIL;
+
+    public static final BehaviorOnNullValues DEFAULT = IGNORE;
+  }
```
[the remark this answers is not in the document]
C0urante: Not opposed to overriding `toString()` but after following your suggestion below about making the `behavior.on.null.values` property case-insensitive, is there still any reason to?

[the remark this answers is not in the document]
kkonstantine: In `enum`s like this one, you almost always want to override `toString`. If you don't you'll get by default `name()` which is the name of the enum member in upper case. 

But we are going to use lower case, so we want a way to print `BehaviorOnNullValues.IGNORE` as `ignore`. That's regardless of the suggestion in config. 

Now, depending what `define` expects in `ConfigDef` you should use `BehaviorOnNullValues.DEFAULT.toString()` or `BehaviorOnNullValues.DEFAULT` instead of hardcoding `"ignore"` at least in the default value for define. Similar elsewhere. Hopefully, we won't have to hardcode a value in any place.

src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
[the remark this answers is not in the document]
kkonstantine: I was referring to just a POJO so you won't make all these local variables, private class members. But you can. Same thing, up to you.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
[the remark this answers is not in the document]
kkonstantine: Up to you when to comply with what's there for the sake of uniformity and when to diverge or fix things to follow our code style. But we don't follow the pattern of declaring everything `final` by default in connectors, at least for now.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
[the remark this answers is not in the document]
C0urante: > But we are going to use lower case, so we want a way to print `BehaviorOnNullValues.IGNORE` as `ignore`.
Right, so the question I had was that if we end up making the field case-insensitive (so that `ignore`, `Ignore`, `IGNORE`, and even `iGnOrE`) are all accepted, does it actually matter that the enums appear as all lowercase when referenced to the user?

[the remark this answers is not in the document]
kkonstantine: yes, we should be a bit flexible w.r.t what we accept, but specific to our suggestion (and consistent to what we print in log messages). If we advice users to give lower case values in our docs (and I agree with your initial suggestion), that's what we should print also. 

Again, this is very common for such enums.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
rhauch: +1 for liberal use of ternary and avoiding unnecessary parens.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
rhauch: +1 for `BehaviorOnNullValues`

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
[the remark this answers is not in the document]
rhauch: Have a look at [ConnectorType](https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/rest/entities/ConnectorType.java) for an example of adding a static lookup method to the enum to handle the case-insensitive lookup. But @kkonstantine is right that toString and docs should use lowercase.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -85,6 +89,34 @@ private String convertKey(Schema keySchema, Object key) {
...
+          throw new DataException(String.format(
+              "Sink record with key of %s and null value encountered for topic/partition/offset %s/%s/%s "
+              + "(to ignore future records like this change the configuration property '%s' from '%s' to '%s')",
+              record.key(),
+              record.topic(),
+              record.kafkaPartition(),
+              record.kafkaOffset(),
+              ElasticsearchSinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG,
+              FAIL,
+              IGNORE
```
rhauch: +1 for adding an instructional error message.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -52,9 +54,11 @@
   }
 
   private final boolean useCompactMapEntries;
+  private final BehaviorOnNullValues behaviorOnNullValues;
 
-  public DataConverter(boolean useCompactMapEntries) {
+  public DataConverter(boolean useCompactMapEntries, BehaviorOnNullValues behaviorOnNullValues) {
     this.useCompactMapEntries = useCompactMapEntries;
+    this.behaviorOnNullValues = behaviorOnNullValues;
```
rhauch: Can `behaviorOnNullValues` parameter be null? Seems like we don't really use asserts (which I like very much for stating intent), but alternatively here you could just assume the default value for safety.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -132,7 +133,10 @@ protected static ConfigDef baseConfigDef() {
...
+                  "How to handle records with non-null keys but null values. Valid options are 'ignore', 'delete', and 'fail'.",
```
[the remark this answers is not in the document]
rhauch: Nit on the grammar of "and" rather than "but": I'd suggest "How to handle records with a non-null key and null value, such as a tombstone record. Valid options ...".

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -110,13 +111,15 @@
...
     private long retryBackoffMs;
+    private DataConverter.BehaviorOnNullValues behaviorOnNullValues;
 
     public Builder(JestClient client) {
       this.client = client;
     }
 
     public Builder setType(String type) {
       this.type = type;
+      this.behaviorOnNullValues = DataConverter.BehaviorOnNullValues.DEFAULT;
```
rhauch: Why is `this.behaviorOnNullValues` changed in this setter for the type?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -177,6 +180,11 @@ public Builder setRetryBackoffMs(long retryBackoffMs) {
       return this;
     }
 
+    public Builder setBehaviorOnNullValues(DataConverter.BehaviorOnNullValues behaviorOnNullValues) {
+      this.behaviorOnNullValues = behaviorOnNullValues;
```
rhauch: Maybe just to be safe:
```
this.behaviorOnNullValues = behaviorOnNullValues != null ? behaviorOnNullValues : BehaviorOnNullValues.DEFAULT
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -68,7 +68,8 @@
       int batchSize,
       long lingerMs,
       int maxRetries,
-      long retryBackoffMs
+      long retryBackoffMs,
+      DataConverter.BehaviorOnNullValues behaviorOnNullValues
```
rhauch: I'd suggest importing `BehaviorOnNullValues` to improve readability.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
rhauch: Actually, I'd recommend using `value != null ? ...` rather than `value == null ? ...` to make it more readable and to avoid misreads like @mageshn (and me!).

src/main/java/io/confluent/connect/elasticsearch/IndexableRecord.java
```diff
@@ -30,6 +34,26 @@ public IndexableRecord(Key key, String payload, Long version) {
     this.payload = payload;
   }
 
+  public BulkableAction toBulkableAction() {
+    // Null payload is treated as a tombstone and will delete from the index.
+    if (payload == null) {
```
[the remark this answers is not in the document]
rhauch: +1

src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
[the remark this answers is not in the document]
rhauch: I'd recommend making these member variables initialized in a `@Before` method (e.g., `setup()`). Tests can use them without modification, but they can always just modify the ones that are needed. Of course, this would be different if all the tests used a different combination.

src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
```diff
@@ -239,4 +245,63 @@ public void struct() {
...
+    final String topic = "topic";
+    final int partition = 12;
+    final long offset = 26;
+    final String index = "index";
+    final String type = "type";
+
+    converter = new DataConverter(true, BehaviorOnNullValues.IGNORE);
+
+    Schema schema = SchemaBuilder.struct().name("struct").field("string", Schema.STRING_SCHEMA).build();
+    SinkRecord sinkRecord = new SinkRecord(topic, partition, Schema.STRING_SCHEMA, key, schema, null, offset);
```
rhauch: Probably could make the `schema` a static field, and could even add a method that created the `SinkRecord` instance that took just the value parameter. The test then becomes something like:

```
@Test
public void ignoreNull() {
  converter = new DataConverter(true, BehaviorOnNullValues.IGNORE);
  record = createRecordWithValue(null);
  assertNull(converter.convertRecord(record, index, type, false, false));
}
```

IMO, it's often much better to have a simple test method that is very easy to see what it is doing, especially when there are other tests doing similar things.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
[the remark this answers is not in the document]
rhauch: I agree that final variables are less useful in test cases, except where they're needed in Java 7 to pass into an anonymous inner class/function.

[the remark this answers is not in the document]
wicknicks: There is a lot of value in keeping test cases brief and terse. Since it is a commonly recurring pattern in these tests, a better way might be to predefine constants for each state of `ignoreKey` and `ignoreSchema`, and then plug them into your test. For example, 

```
  private static final boolean ENFORCE_SCHEMA = true;
  private static final boolean IGNORE_SCHEMA = false;
```

and then call initWriter as follows: 

```
  initWriter(client, ENFORCE_KEY, IGNORE_SCHEMA, BehaviorOnNullValues.IGNORE);
```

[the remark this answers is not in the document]
C0urante: I'm not opposed to rewriting the tests to reflect our code style and/or to make them brief, readable, etc., but wouldn't it be better to address that in a separate PR? Don't want to clutter this one up and/or make it take longer than necessary, since changes in style aren't required for the new feature to work properly. Only asking because I feel like this is a bit of a can of worms and might take longer than anticipated--code style is easy to have disagreements over 😉

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: As of right now, none of the instantiations of the `DataConverter` within the project are at risk of passing in `null` for the `behaviorOnNullValues` parameter. I'm hesitant to allow `null` as a value here and just implicitly replace it with `DEFAULT`; seems better to, as you pointed out, utilize an assert to ensure it isn't `null`. At the very least, if `DEFAULT` is used as a replacement for `null`, I think a Javadoc comment should be added above the class detailing this behavior just to make it easier should somebody else try to use the class. Alternatively, we could overload the constructor with a (well-documented) one-arg variant that then calls the two-arg constructor with `DEFAULT` as the second argument, which we can then safely assert as non-null.

Mostly, just don't like the idea of adding implicit behavior and want to avoid any potential future confusion to arise from that.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
C0urante: Oops!

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
C0urante: Is implicit behavior a good idea here? Seems like throwing an NPE might help avoid confusion down the road--if `null` is intentionally being passed, a suggestion could be included in the exception message to instead pass `DEFAULT` explicitly; if `null` is accidentally being passed, it could save some headache over unexpected behavior.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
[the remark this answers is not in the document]
rhauch: Handling these as a separate issue/PR might be fine if this was non-test related code or was cleaning up existing code. However, this PR is adding additional tests that we're commenting on, and I'd argue that this goes beyond style and ventures into good coding practices. I don't think it's appropriate to forgo this an incur the debt when it's easy enough to fix now.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
rhauch: Let's not overload the constructor if we don't need it. If the goal is that the supplied parameter never is null, then let's add an assertion and JavaDoc that explains this.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
rhauch: Sure, disallowing the parameter to be null is perfectly acceptable. This is internal code that will never be null, and in this case I'd be fine with an assert to document that null is not allowed and to prevent changes in the future that might pass null. OTOH, if this were public code or even code that was more difficult to control, then we'd actually want to check the parameter and throw an IllegalArgumentException.

src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
kkonstantine: Totally up for that ^^

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: Alright, settling on `BehaviorOnNullValues` (in spite of a personal preference for `NullValueBehavior`) for now.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: I like the idea of inverting the condition to improve readability, will include that in the next commit. Let me know if readability is still lacking though, definitely not opposed to vanilla if/else if all else (pun intended) fails.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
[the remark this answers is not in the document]
C0urante: Alright, sounds good. Will adapt the `forValue()` method from the ConnectorType enum to enable case-insensitive reading (as long as all the enum values are uppercase), and override `toString()` to return `name().toLowerCase()`.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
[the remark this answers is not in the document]
C0urante: Sounds fine to me, case-insensitive support shouldn't be too hard to add.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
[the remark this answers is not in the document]
C0urante: Any thoughts on

>How to handle records with a non-null key and a null value (i.e. Kafka tombstone records). Valid options are 'ignore', 'delete', and 'fail' (case-insensitive).

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -220,7 +229,23 @@ public void write(Collection<SinkRecord> records) {
       final IndexableRecord indexableRecord = converter.convertRecord(sinkRecord, index, type,
                                                                       ignoreKey, ignoreSchema);
 
-      bulkProcessor.add(indexableRecord, flushTimeoutMs);
+      // In the event that the sink record's value was null and the data converter has been told to IGNORE
+      // null values, the returned record will be null.
+      // TODO: Discover if any of the lines above incur a significant performance penalty. If they do, bypass
```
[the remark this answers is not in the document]
C0urante: Good idea on the `TODO` 'hanging above' lines it's referencing. I'll leave it in the same place but reword to hopefully make the meaning more clear--the lines it refers to are the ones directly below it; it's just meant to the possibility of moving them to/adding something else at the top of the for-loop.

src/main/java/io/confluent/connect/elasticsearch/IndexableRecord.java
```diff
@@ -30,6 +34,26 @@ public IndexableRecord(Key key, String payload, Long version) {
     this.payload = payload;
   }
 
+  public BulkableAction toBulkableAction() {
+    // Null payload is treated as a tombstone and will delete from the index.
```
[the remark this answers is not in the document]
C0urante: What about
>If payload is null, the record was a tombstone and we should delete from the index.

src/main/java/io/confluent/connect/elasticsearch/IndexableRecord.java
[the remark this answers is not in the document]
C0urante: Good call. Will refactor to ternary operator here.

src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
```diff
@@ -174,7 +180,7 @@ public void stringKeyedMapNonCompactFormat() {
     origValue.put("field2", 2);
 
     // Use the older non-compact format for map entries with string keys
-    converter = new DataConverter(false);
+    converter = new DataConverter(false, DataConverter.BehaviorOnNullValues.DEFAULT);
```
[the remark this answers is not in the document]
C0urante: Nice catch. Will make adjustments accordingly, here and wherever else applicable.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
C0urante: Totally right. Turns out, `behaviorOnNullValues` doesn't have to be adjusted at all for the `DataConverter` used in the `ElasticsearchSinkTestBase` class and (for now) can and should just be set to `FAIL`.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: Will do!

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
C0urante: Alright, will add a call to `Objects.requireNonNull()` here and (hopefully) adequate documentation to help anyone encountering the resulting NPE.

src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
C0urante: I like that approach, will try to adopt it appropriately in the next commit.

C0urante:
@kkonstantine, @mageshn, @wicknicks, @rhauch, next commit is up--further review welcome!

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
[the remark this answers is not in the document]
kkonstantine: Sounds good to me. I wouldn't add the "(case-insensitive)" part. 

We don't care to encourage diversion here. We just want to support broadly this property so we don't have issues or questions on the mailing list.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -52,9 +54,18 @@
...
+  /**
+   * @param behaviorOnNullValues How to treat records with null values. Cannot be null; if in doubt,
+   * {@link BehaviorOnNullValues#DEFAULT} can be used.
+   */
+  public DataConverter(boolean useCompactMapEntries, BehaviorOnNullValues behaviorOnNullValues) {
     this.useCompactMapEntries = useCompactMapEntries;
+    this.behaviorOnNullValues = Objects.requireNonNull(
+        behaviorOnNullValues,
+        "behaviorOnNullValues cannot be null. Possible fix: use BehaviorOnNullValues.DEFAULT instead."
+    );
```
rhauch: IMO, the message is not necessary. This constructor is only called by the `ElasticsearchWriter` (and lots of tests), and the `behaviorOnNullValue` should be null only when a developer introduced a bug.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -52,9 +54,18 @@
   }
 
   private final boolean useCompactMapEntries;
+  private final BehaviorOnNullValues behaviorOnNullValues;
 
-  public DataConverter(boolean useCompactMapEntries) {
+  /**
+   * @param behaviorOnNullValues How to treat records with null values. Cannot be null; if in doubt,
+   * {@link BehaviorOnNullValues#DEFAULT} can be used.
```
rhauch: If you add JavaDoc to any method, that JavaDoc should be complete: it should have a good description, `@param` for every parameter, `@return` if the method has a return, and any thrown exceptions. In this case, you're missing the `@param useCompactMapEntries`.

When you build, are you using `mvn clean install`? I'm surprised that checkstyle didn't catch this.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -85,6 +96,34 @@ private String convertKey(Schema keySchema, Object key) {
   }
 
   public IndexableRecord convertRecord(SinkRecord record, String index, String type, boolean ignoreKey, boolean ignoreSchema) {
+    if (record.value() == null) {
+      switch (behaviorOnNullValues) {
+        case IGNORE:
+          return null;
+        case DELETE:
+          // Will proceed as normal, ultimately creating an IndexableRecord with a null payload
+          break;
```
rhauch: Note that the behavior is described by the config description as "How to handle records with a non-null key and a null value", but this logic is not really checking whether the key is not null. Do we need to do something like this:

```
    case DELETE:
      if (!ignoreKey && record.key() == null) {
        // We should use the key but the key is also null, so ignore
        return null;
      }
      // Will proceed as normal, ultimately creating an IndexableRecord with a null payload
      break;
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -77,6 +79,7 @@ public void start(Map<String, String> props, JestClient client) {
       int maxRetry = config.getInt(ElasticsearchSinkConnectorConfig.MAX_RETRIES_CONFIG);
       int connTimeout = config.getInt(ElasticsearchSinkConnectorConfig.CONNECTION_TIMEOUT_MS_CONFIG);
       int readTimeout = config.getInt(ElasticsearchSinkConnectorConfig.READ_TIMEOUT_MS_CONFIG);
+      BehaviorOnNullValues behaviorOnNullValues = BehaviorOnNullValues.forValue(config.getString(ElasticsearchSinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG));
```
rhauch: People get sloppy and sometimes include leading or trailing whitespace. Might want to `trim()` the string in the `BehaviorOnNullValues.forValue(String)` method before uppercasing it.

C0urante: Looks like the `AbstractConfg` class's `getString()` method already strips leading and trailing whitespace; is it still worth adapting for the cases where someone explicitly invokes `BehaviorOnNullValues.forValue()` themselves?

rhauch: Ah, good catch. No need to change anything then.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
```diff
@@ -293,9 +297,84 @@ public void testBytes() throws Exception {
...
+  public void testFailOnNullValue() throws Exception {
+    Collection<SinkRecord> records = new ArrayList<>();
+    SinkRecord sinkRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, key, schema, null, 0);
+    records.add(sinkRecord);
+
+    ElasticsearchWriter writer = initWriter(client, BehaviorOnNullValues.FAIL);
+    try {
+      writeDataAndRefresh(writer, records);
+      fail("should fail because of behavior.on.null.values=fail");
+    } catch (DataException dexc) {
```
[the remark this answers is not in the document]
C0urante: Ahh, okay. Was assuming checkstyle would complain but if I can get away with it `e` is definitely a preferable name for exceptions.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -220,7 +239,22 @@ public void write(Collection<SinkRecord> records) {
       final IndexableRecord indexableRecord = converter.convertRecord(sinkRecord, index, type,
                                                                       ignoreKey, ignoreSchema);
 
-      bulkProcessor.add(indexableRecord, flushTimeoutMs);
+      // In the event that the sink record's value was null and the data converter has been told to IGNORE
+      // null values, the returned record will be null.
+      // TODO: If necessary, move the check for null-valued records to the top of the for-loop to avoid potential
+      //       performance penalties. Leaving the check here since it relegates all three kinds of null-record-handling
```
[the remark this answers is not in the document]
rhauch: I think "relegates" is more accurate but less common. Personally, I'd avoid this altogether with:

> Leaving the check here since DataConverter handles all of null-record behaviors

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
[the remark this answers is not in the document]
rhauch: +1

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -242,4 +283,45 @@ Object preProcessValue(Object value, Schema schema, Schema newSchema) {
...
+    public static final BehaviorOnNullValues DEFAULT = IGNORE;
+
+    // Want values for "behavior.on.null.values" property to be case-insensitive
+    public static final ConfigDef.Validator VALIDATOR = new ConfigDef.Validator() {
+      private final ConfigDef.ValidString validator = ConfigDef.ValidString.in(names());
+
+      @Override
+      public void ensureValid(String name, Object value) {
+        if (value instanceof String) {
+          value = ((String) value).toLowerCase();
```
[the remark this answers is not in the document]
C0urante: Gotcha, will add.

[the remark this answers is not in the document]
C0urante: Any idea why `mvn clean install` isn't failing my build already with stuff like that present?

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: I am using `mvn clean install` to build--should checkstyle be refusing to build for me when encountering errors like this, or do I have to look through the build output for warnings and/or through the reports directory for results files?

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: Should they be ignored even when `ignoreKey` is `true`, then?

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -96,13 +135,15 @@ public IndexableRecord convertRecord(SinkRecord record, String index, String typ
     final Object value;
     if (!ignoreSchema) {
```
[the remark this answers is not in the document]
C0urante: In that case, how about we leave `schema` and `value` inside the scope of the `if` statement and do

```
    final String payload;
    if (record.value() != null) {
      Schema schema = ignoreSchema ? record.valueSchema() : preProcessSchema(record.valueSchema());
      Object value = ignoreSchema ? record.value() : preProcessValue(record.value(), record.valueSchema(), schema);
      payload = new String(JSON_CONVERTER.fromConnectData(record.topic(), schema, value), StandardCharsets.UTF_8);
    } else {
      payload = null;
    }
```

instead?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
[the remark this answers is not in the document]
C0urante: AFAIU, `relegates` basically means `delegates to a lower level`, which is a useful clarification to make as for why delegating the handling of `null`-valued records to the `DataConverter` class is useful. The `ElasticsearchWriter` is high-level enough that it should be mostly concerned with just writing data, instead of deciding what to write (which is beneath it, and arguably more the responsibility of the `DataConverter`).

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
```diff
@@ -41,70 +43,89 @@
 
 import io.searchbox.client.JestClient;
 
+import static io.confluent.connect.elasticsearch.DataConverter.BehaviorOnNullValues;
+
 @ThreadLeakScope(ThreadLeakScope.Scope.NONE)
 public class ElasticsearchWriterTest extends ElasticsearchSinkTestBase {
 
+  // To be used as the ignoreKey paramter when calling initWriter
+  private static final boolean ENFORCE_KEY = false;
```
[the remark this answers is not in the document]
C0urante: Yeah, as the refactoring continued that started to seem less and less necessary. Will remove.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: Makes sense, will shorten the message to `"behaviorOnNullValues cannot be null"` (unless that's still too much and no message is preferable).

rhauch:
@C0urante, one more thing. Please run the [ElasticsearchSinkConnectorConfig.main()](https://github.com/confluentinc/kafka-connect-elasticsearch/blob/master/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java#L329-L331) method, and take the output from it and update https://github.com/confluentinc/kafka-connect-elasticsearch/blob/master/docs/configuration_options.rst in this PR. This is a manual process to update the docs, but I'd prefer to do it as part of the changes rather than later on just prior to the release (when it's easy to forget). Thanks!

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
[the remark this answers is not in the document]
rhauch: We recently added more stringent checkstyle rules to 4.0, so they're not in the `3.3.x` branch. See @kkonstantine's suggestion for temporarily rebasing on `4.0.x` to get the better checkstyle rules, then rebasing on `3.3.x` before pushing to the PR branch.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
rhauch: Is that even required? Why not just `Objects.requireNonNull(behaviorOnNullValues)`?

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
rhauch: No, they should not be ignored when `ignoreKey == true` because then there is a key created from the topic partition and offset information.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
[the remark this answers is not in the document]
rhauch: That seems good.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
kkonstantine: @C0urante `getString` does not trim the value. It returns what's in the map of parsed values. 

The string value is parsed when it's added to the map as part of the overall parsing of the original properties.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: It isn't, just seems like it might make life slightly easier for anyone running into the NPE in the future. Message can definitely be removed as an argument if that's better.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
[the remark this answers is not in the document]
kkonstantine: Sure

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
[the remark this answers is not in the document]
kkonstantine: According to my dictionary, `relegate` seems a bit too accurate, because it's interpretation as `downgrading` I don't think is particularly useful. I'd agree with @rhauch's suggestion to remove it altogether.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: Aren't t/p/o guaranteed to be unique per message?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
C0urante: @kkonstantine right, point still stands though--what `getString()` returns is the trimmed version of what the `AbstractConfig` was given to begin with.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
rhauch: Good point. If a record's key and value are both null, then DELETE behavior can't do anything with the null and DELETE means nothing for a generated key. Just add a comment to clarify this.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -52,9 +54,18 @@
...
-  public DataConverter(boolean useCompactMapEntries) {
+  /**
+   * @param behaviorOnNullValues How to treat records with null values. Cannot be null; if in doubt,
+   * {@link BehaviorOnNullValues#DEFAULT} can be used.
+   */
+  public DataConverter(boolean useCompactMapEntries, BehaviorOnNullValues behaviorOnNullValues) {
     this.useCompactMapEntries = useCompactMapEntries;
+    this.behaviorOnNullValues = Objects.requireNonNull(
+        behaviorOnNullValues,
+        "behaviorOnNullValues cannot be null. Possible fix: use BehaviorOnNullValues.DEFAULT instead."
```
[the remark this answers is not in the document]
C0urante: Wouldn't making the Checkstyle build pass with the 4.0.x branch's standards mean fixing every Checkstyle error in every file, not just the errors present in the changes I made?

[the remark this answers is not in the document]
kkonstantine: The lines you are not adding have been fixed in `4.0.x`. 

The goal is make the new lines pass the checks (I've given several examples where they won't) and also minimize conflicts from 3.3.x to 4.0.x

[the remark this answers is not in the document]
rhauch: The `4.0.x` branch as it currently stands should already pass the checkstyle rules. So temporarily rebasing against this branch (if you can do this) would apply your changes on top of what's currently in `4.0.x`, so checkstyle will make sure that _your changes_ (plus all non-changed lines) satisfy the rules. When you take these commits and then rebase them on top of `3.3.x`, your changes should satisfy the 4.0.x checkstyle rules.

This may not be that straightforward, though, and it may depend on how cleanly git merge can do the rebase. You'd have to fix the checkstyle rules anyway for the `4.0.x` branch as a separate merge process after this PR is merged to the `3.3.x` branch, so IMO it's worth a shot doing it now.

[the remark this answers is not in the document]
C0urante: There are also features on the `4.0.x` branch that aren't included on the `3.3.x` line (e.g., `drop.invalid.message`; should those be merged in as well?

C0urante:
@kkonstantine, @mageshn, @wicknicks, @rhauch, next round of commits is up and ready for further review!

rhauch:
@C0urante: when we discussed above temporarily rebasing on `4.0.x`, I was not expecting to bring in all of the reformatting changes that were made on the `4.0.x` branch. Doing that is a mistake IMO because it obscures all of the changes for this issue.

I created an alternative PR (#165) by starting with your branch, squashing your commits into a single commit, and then creating a new topic branch off of the `3.3.x` branch, cherry-picking that single squashed commit, and cleaning up a few merge conflicts. (Squashing made it easier to cherry-pick and clean up the merge conflicts.) I also found a mistake in the commits related to CC-1059 and corrected that with a second commit. Let me know what you think.

C0urante:
Sorry for making such a mess of things and thanks for stepping in and giving me a hand with getting them back to normal! Looking at the new PR now.

rhauch:
@C0urante, no worries. It may be that #165 won't cleanly merge onto `4.0.x`, but it shouldn't take much to fix it manually and then continue merging on that and `master.

rhauch:
Closing in favor of #165.

PR #165, opened 2018-01-17
CC-350, CC-1097: Added support for delete and null handling

rhauch:
Added a configuration to control how records with null values (e.g., tombstone messages) are handled, with three options: `IGNORE`, `DELETE`, and `FAIL`.

The `FAIL` option is identical to the previous behavior. The `DELETE` option generates a corresponding delete operation in the Elasticsearch system only when the record has an explicit key; if `key.ignore` is true, the keys will be generated and delete does not make sense. And finally the `IGNORE` option tells the connector to quietly skip over any records that have a null value.

Although the previous behavior is equivalent to `FAIL`, that cause the connector for fail and stop whenever it came across a null value. A more sensible default is `IGNORE`.

NOTE: This is an alternative PR with nearly the same changes as #163 but with fewer commits and without the other formatting/style changes.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -16,6 +16,7 @@
 
 package io.confluent.connect.elasticsearch;
 
+import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.connect.data.ConnectSchema;
 import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
@@ -38,7 +39,9 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConstants.MAP_KEY;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConstants.MAP_VALUE;
@@ -52,9 +55,23 @@ public class DataConverter {
   }
 
   private final boolean useCompactMapEntries;
+  private final BehaviorOnNullValues behaviorOnNullValues;
 
-  public DataConverter(boolean useCompactMapEntries) {
+  /**
+   * Create a DataConverter, specifying how map entries with string keys within record
+   * values should be written to JSON. Compact map entries are written as
+   * <code>"entryKey": "entryValue"</code>, while the non-compact form are written as a nested
+   * document such as <code>{"key": "entryKey", "value": "entryValue"}</code>. All map entries
+   * with non-string keys are always written as nested documents.
+   *
+   * @param useCompactMapEntries true for compact map entries with string keys, or false for
+   *                             the nested document form.
+   * @param behaviorOnNullValues behavior for handling records with null values; may not be null
+   */
+  public DataConverter(boolean useCompactMapEntries, BehaviorOnNullValues behaviorOnNullValues) {
     this.useCompactMapEntries = useCompactMapEntries;
+    this.behaviorOnNullValues =
+        Objects.requireNonNull(behaviorOnNullValues, "behaviorOnNullValues cannot be null.");
   }
 
   private String convertKey(Schema keySchema, Object key) {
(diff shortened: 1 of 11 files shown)


src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -242,4 +303,51 @@ Object preProcessValue(Object value, Schema schema, Schema newSchema) {
         return value;
     }
   }
+  public enum BehaviorOnNullValues {
+    IGNORE,
+    DELETE,
+    FAIL;
+
+    public static final BehaviorOnNullValues DEFAULT = IGNORE;
```
C0urante: If we're aiming for backwards compatibility ("The FAIL option is identical to the previous behavior, so this option is the default to ensure backward compatibility for existing connector deployments that are upgrading") this should be changed to `DEFAULT = FAIL;`.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -220,7 +238,19 @@ public void write(Collection<SinkRecord> records) {
       final IndexableRecord indexableRecord = converter.convertRecord(sinkRecord, index, type,
                                                                       ignoreKey, ignoreSchema);
 
-      bulkProcessor.add(indexableRecord, flushTimeoutMs);
+      // In the event that the sink record's value was null and the data converter has been told to
+      // ignore null values, the returned record will be null.
```
C0urante: Had a `TODO` here just in case it turned out it might be advantageous for the `ElasticsearchWriter` to intercept null-valued records before this point when `behavior.on.null.values` is set to `IGNORE`, in order to avoid possible performance penalties and/or other issues arising from the `if (!ignoreSchema && !existingMappings.contains(index))` block. Should that be put back in or was it removed intentionally?

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
kkonstantine: I think it's safe to enable a sane config as the default (here `IGNORE`). Backwards compatibility, as I see it here, means that if people have built software around the assumption that the connector fails, they can still enable this behavior. 

But we should have the connector, starting with this version, not failing when it encounters records with non-null key/null value.

src/test/java/io/confluent/connect/elasticsearch/MappingTest.java
```diff
@@ -109,7 +111,7 @@ private void verifyMapping(Schema schema, JsonObject mapping) throws Exception {
       }
     }
 
-    DataConverter converter = new DataConverter(true);
+    DataConverter converter = new DataConverter(true, BehaviorOnNullValues.FAIL);
```
kkonstantine: Do we want the current default here (`IGNORE`)?

C0urante: No, this should be `FAIL` since there aren't any apparent cases for passing a null-valued record to the converter in this test.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
rhauch: I'm fine with changing the documentation, since the previous behavior is not really a useful default. @C0urante, do you want to make those changes and push them to this branch?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
rhauch: I removed it intentionally. We can change the code now, but I don't think the speculative TODO was useful.

rhauch:
Correctly based on `3.3.x` rather than `master`.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
C0urante: Which documentation changes are you referring to? `configuration_options.rst` already lists `ignore` as the default, not sure where else would need to be changed that hasn't already been.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
C0urante: Alright, fair enough. Guess it can't hurt to change it now--as conceptually pleasing as it is to [dr]elegate (😜) handling of null-valued records to the `DataConverter` class I see some calls to the `JestClient` inside the for-loop above that would probably be beneficial to avoid.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
rhauch: Sorry, that comment was in the commit. Let me amend the commit message.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
rhauch: Let me know if you have problems pushing to this branch.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -218,6 +220,11 @@ public ElasticsearchWriter build() {
 
   public void write(Collection<SinkRecord> records) {
     for (SinkRecord sinkRecord : records) {
+      // Preemptively skip records with null values if they're going to be ignored anyways
+      if (sinkRecord.value() == null && behaviorOnNullValues.equals(BehaviorOnNullValues.IGNORE)) {
```
rhauch: How about `behaviorOnNullValues == BehaviorOnNullValues.IGNORE`? Enums are more like primitives than like objects.

rhauch:
LGTM, but I cannot approve the PR since I created it. @C0urante  and @kkonstantine, please approve or comment.

src/test/java/io/confluent/connect/elasticsearch/MappingTest.java
kkonstantine: I agree. `null` value is not tested here. Still, since we need to give an argument there, we should rather give the default, than one that would be used less frequently

rhauch: The `DataConverterTest` verifies that `DataConverter` behaves correctly. This test is merely using a `DataConverter` to check the mapping, and it seems like we'd want the test to fail if there are any unexpected problems. How would using a different `BehaviorOnNullValue` affect this test?

kkonstantine: I don't think it would affect the test, since this test doesn't seem to check the case of record with a `null` value. My point is that we should not depend on testing something "by accident". If we need a test case for `null` we should add one.

PR #169, opened 2018-01-29
CC-1385:  Enhance connector to use text type with ES 5+

rayokota:
The ES connector now queries for the version.  The version is stored in
an instance of ElasticsearchClient, which is passed around so that
it can be queried at the appropriate times (such as when inferring
schema mappings).

As part of this change, the old Jest client is now wrapped by a
higher-level client object.  Also, all the Jest dependencies have been
isolated to the io.confluent.connect.elasticsearch.jest package.  This
makes the third-party dependencies more clear and will facilitate moving
to a different ES client library in the future should we choose to do
so.

These changes have been tested with ES 2.x, 5.x, and 6.x.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,16 +39,16 @@
     </scm>
 
     <properties>
+        <!--
         <es.version>6.0.0</es.version>
         <lucene.version>7.0.1</lucene.version>
-        <!--
         <es.version>5.0.0</es.version>
         <lucene.version>6.2.0</lucene.version>
+        -->
         <es.version>2.4.1</es.version>
         <lucene.version>5.5.2</lucene.version>
-        -->
         <hamcrest.version>1.3</hamcrest.version>
-        <jest.version>5.3.3</jest.version>
+        <jest.version>2.4.0</jest.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
     </properties>
 
@@ -113,7 +113,6 @@
             <scope>test</scope>
         </dependency>
         <!-- For ES 2.x -->
-        <!--
         <dependency>
             <groupId>org.elasticsearch</groupId>
             <artifactId>elasticsearch</artifactId>
@@ -121,7 +120,8 @@
             <scope>test</scope>
             <type>test-jar</type>
         </dependency>
-        -->
+        <!-- For ES 5.x (requires Java 8) -->
+        <!--
         <dependency>
             <groupId>org.elasticsearch.test</groupId>
             <artifactId>framework</artifactId>
@@ -129,14 +129,15 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.elasticsearch</groupId>
-            <artifactId>elasticsearch</artifactId>
+            <groupId>org.elasticsearch.plugin</groupId>
+            <artifactId>transport-netty4-client</artifactId>
             <version>${es.version}</version>
             <scope>test</scope>
         </dependency>
+        -->
         <dependency>
-            <groupId>org.elasticsearch.plugin</groupId>
-            <artifactId>transport-netty4-client</artifactId>
+            <groupId>org.elasticsearch</groupId>
+            <artifactId>elasticsearch</artifactId>
             <version>${es.version}</version>
             <scope>test</scope>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -39,11 +39,16 @@
     </scm>
 
     <properties>
+        <es.version>6.0.0</es.version>
+        <lucene.version>7.0.1</lucene.version>
+        <!--
+        <es.version>5.0.0</es.version>
+        <lucene.version>6.2.0</lucene.version>
         <es.version>2.4.1</es.version>
         <lucene.version>5.5.2</lucene.version>
-        <jna.version>4.2.1</jna.version>
-        <hamcrest.version>2.0.0.0</hamcrest.version>
-        <jest.version>2.0.0</jest.version>
+        -->
+        <hamcrest.version>1.3</hamcrest.version>
+        <jest.version>5.3.3</jest.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
     </properties>
 
@@ -72,14 +77,8 @@
             <version>${jest.version}</version>
         </dependency>
         <dependency>
-            <groupId>net.java.dev.jna</groupId>
-            <artifactId>jna</artifactId>
-            <version>${jna.version}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-junit</artifactId>
+            <artifactId>hamcrest-all</artifactId>
             <version>${hamcrest.version}</version>
             <scope>test</scope>
         </dependency>
@@ -102,15 +101,42 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+            <version>2.7</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>2.7</version>
+            <scope>test</scope>
+        </dependency>
+        <!-- For ES 2.x -->
+        <!--
+        <dependency>
             <groupId>org.elasticsearch</groupId>
             <artifactId>elasticsearch</artifactId>
             <version>${es.version}</version>
             <scope>test</scope>
             <type>test-jar</type>
         </dependency>
+        -->
+        <dependency>
+            <groupId>org.elasticsearch.test</groupId>
+            <artifactId>framework</artifactId>
+            <version>${es.version}</version>
+            <scope>test</scope>
+        </dependency>
         <dependency>
             <groupId>org.elasticsearch</groupId>
             <artifactId>elasticsearch</artifactId>
+            <version>${es.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.elasticsearch.plugin</groupId>
+            <artifactId>transport-netty4-client</artifactId>
             <version>${es.version}</version>
             <scope>test</scope>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -39,15 +39,10 @@
     </scm>
 
     <properties>
-        <!--
-        <es.version>6.0.0</es.version>
-        <lucene.version>7.0.1</lucene.version>
-        <es.version>5.0.0</es.version>
-        <lucene.version>6.2.0</lucene.version>
-        -->
         <es.version>2.4.1</es.version>
         <lucene.version>5.5.2</lucene.version>
         <hamcrest.version>1.3</hamcrest.version>
+        <mockito.version>2.13.0</mockito.version>
         <jest.version>2.4.0</jest.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
     </properties>
@@ -83,6 +78,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>${mockito.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>${junit.version}</version>
@@ -100,7 +101,6 @@
             <version>${lucene.version}</version>
             <scope>test</scope>
         </dependency>
-        <!-- For ES 2.x -->
         <dependency>
             <groupId>org.elasticsearch</groupId>
             <artifactId>elasticsearch</artifactId>
@@ -108,33 +108,6 @@
             <scope>test</scope>
             <type>test-jar</type>
         </dependency>
-        <!-- For ES 5.x (requires Java 8) -->
-        <!--
-        <dependency>
-            <groupId>org.elasticsearch.test</groupId>
-            <artifactId>framework</artifactId>
-            <version>${es.version}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.elasticsearch.plugin</groupId>
-            <artifactId>transport-netty4-client</artifactId>
-            <version>${es.version}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-api</artifactId>
-            <version>2.7</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-core</artifactId>
-            <version>2.7</version>
-            <scope>test</scope>
-        </dependency>
-        -->
         <dependency>
             <groupId>org.elasticsearch</groupId>
             <artifactId>elasticsearch</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -100,18 +100,6 @@
             <version>${lucene.version}</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-api</artifactId>
-            <version>2.7</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-core</artifactId>
-            <version>2.7</version>
-            <scope>test</scope>
-        </dependency>
         <!-- For ES 2.x -->
         <dependency>
             <groupId>org.elasticsearch</groupId>
@@ -132,6 +120,18 @@
             <groupId>org.elasticsearch.plugin</groupId>
             <artifactId>transport-netty4-client</artifactId>
             <version>${es.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+            <version>2.7</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>2.7</version>
             <scope>test</scope>
         </dependency>
         -->
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -1,0 +1,97 @@
+/**
+ * Copyright 2018 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ **/
+
+package io.confluent.connect.elasticsearch;
+
+import com.google.gson.JsonObject;
+import io.confluent.connect.elasticsearch.bulk.BulkRequest;
+import io.confluent.connect.elasticsearch.bulk.BulkResponse;
+import org.apache.kafka.connect.data.Schema;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Set;
+
+public interface ElasticsearchClient {
+
+  enum Version {
+    ONE, TWO, FIVE, SIX
+  }
+
+  /**
+   * Gets the Elasticsearch version.
+   *
+   * @return the version
+   */
+  Version getVersion();
+
+  /**
+   * Creates indices.
+   *
+   * @param indices the set of index names to create
+   */
+  void createIndices(Set<String> indices);
+
+  /**
+   * Creates an explicit mapping.
+   *
+   * @param index the index to write
+   * @param type the type for which to create the mapping
+   * @param schema the schema used to infer the mapping
+   * @throws IOException from underlying client
+   */
+  void createMapping(String index, String type, Schema schema) throws IOException;
+
+  /**
+   * Gets the JSON mapping for the given index and type. Returns {@code null} if it does not exist.
+   *
+   * @param index the index
+   * @param type the type
+   * @throws IOException from underlying client
+   */
+  JsonObject getMapping(String index, String type) throws IOException;
+
+  /**
+   * Creates a bulk request for the list of {@link IndexableRecord} records.
+   *
+   * @param batch the list of records
+   * @return the bulk request
+   */
+  BulkRequest createBulkRequest(List<IndexableRecord> batch);
+
+  /**
+   * Executes a bulk action.
+   *
+   * @param bulk the bulk request
+   * @return the bulk response
+   */
+  BulkResponse executeBulk(BulkRequest bulk) throws IOException;
+
+  /**
+   * Executes a search.
+   *
+   * @param query the search query
+   * @param index the index to search
+   * @param type the type to search
+   * @return the search result
+   */
+  JsonObject search(String query, String index, String type) throws IOException;
+
+  /**
+   * Shuts down the client.
+   */
+  void shutdown();
+}
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -25,23 +25,23 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Set;
 
-public interface ElasticsearchClient {
+public interface ElasticsearchClient extends AutoCloseable {
 
   enum Version {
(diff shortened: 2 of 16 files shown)


src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -0,0 +1,317 @@
...
+public class JestElasticsearchClient implements ElasticsearchClient {
+  private static final Logger LOG = LoggerFactory.getLogger(JestElasticsearchClient.class);
+
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  private final JestClient client;
+  private final Version version;
+
+  @VisibleForTesting
+  public JestElasticsearchClient(String address) {
```
rhauch: Can this be package-level scope?

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
```diff
@@ -108,24 +92,58 @@ public static JsonNode inferMapping(Schema schema) {
...
         return properties;
       default:
-        String esType = ElasticsearchSinkConnectorConstants.TYPES.get(schemaType);
+        String esType = getElasticsearchType(client, schemaType);
         return inferPrimitive(esType, schema.defaultValue());
     }
   }
 
+  @VisibleForTesting
+  public static String getElasticsearchType(ElasticsearchClient client, Schema.Type schemaType) {
```
rhauch: Can this be package-level scope?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -0,0 +1,317 @@
...
+      throw new ConnectException(
+              "Couldn't start ElasticsearchSinkTask due to configuration error:",
+              e
+      );
+    }
+  }
+
+  private Version getServerVersion() throws IOException {
+    // Default to newest version for forward compatibility
+    Version defaultVersion = Version.SIX;
```
rhauch: IIUC, ES 6 removes the old string type that was deprecated in 5. If that's true, then how is defaulting to 6 forward compatible?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -0,0 +1,317 @@
...
+      return defaultVersion;
+    }
+
+    JsonObject nodeRoot = nodesRoot.entrySet().iterator().next().getValue().getAsJsonObject();
+    if (nodeRoot == null) {
+      LOG.warn("Couldn't get Elasticsearch version");
+      return defaultVersion;
+    }
+
+    String esVersion = nodeRoot.get("version").getAsString();
```
rhauch: Does this *should* work for all of the ES versions and the above logic is really handling exceptional cases? If so, then perhaps this method should have doc that explains that. And, the log messages should maybe be a bit more descriptive so that we can more easily tell which path the logic takes.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -0,0 +1,97 @@
...
+public interface ElasticsearchClient {
+
+  enum Version {
+    ONE, TWO, FIVE, SIX
+  }
+
+  /**
+   * Gets the Elasticsearch version.
+   *
+   * @return the version
```
rhauch: Never null?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -0,0 +1,97 @@
...
+   */
+  Version getVersion();
+
+  /**
+   * Creates indices.
+   *
+   * @param indices the set of index names to create
```
rhauch: can `indices` be null?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -0,0 +1,97 @@
...
+   * @param batch the list of records
+   * @return the bulk request
+   */
+  BulkRequest createBulkRequest(List<IndexableRecord> batch);
+
+  /**
+   * Executes a bulk action.
+   *
+   * @param bulk the bulk request
+   * @return the bulk response
```
rhauch: Missing `@throws IOException if the client cannot execute the request`

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -0,0 +1,97 @@
...
+   */
+  BulkResponse executeBulk(BulkRequest bulk) throws IOException;
+
+  /**
+   * Executes a search.
+   *
+   * @param query the search query
+   * @param index the index to search
+   * @param type the type to search
+   * @return the search result
```
rhauch: Missing `@throws IOException if the client cannot execute the request`

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -0,0 +1,97 @@
...
+   */
+  JsonObject search(String query, String index, String type) throws IOException;
+
+  /**
+   * Shuts down the client.
+   */
+  void shutdown();
```
rhauch: Have you considered extending `AutoCloseable` and then overriding the `close()` method here to throw no exceptions?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -0,0 +1,317 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class JestElasticsearchClient implements ElasticsearchClient {
```
rhauch: Can we add unit tests for this? Might have to mock the JestClient, in which case we'd want a constructor that takes the JestClient so we can pass in the mock. This class' logic is not trivial, and it seems like it'd be better to test this logic in isolation. Thoughts?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -0,0 +1,317 @@
...
+  public void createIndices(Set<String> indices) {
+    for (String index : indices) {
+      if (!indexExists(index)) {
+        CreateIndex createIndex = new CreateIndex.Builder(index).build();
+        try {
+          JestResult result = client.execute(createIndex);
+          if (!result.isSucceeded()) {
+            String msg = result.getErrorMessage() != null ? ": " + result.getErrorMessage() : "";
+            throw new ConnectException("Could not create index '" + index + "'" + msg);
+          }
```
rhauch: Is it possible that two tasks attempt to look for the same index, neither finds it, and so both attempt to create the same index? If so, would it make sense if the create request fails to first check and see if it exists (perhaps again) and to fail only if the index does not already exist?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -0,0 +1,317 @@
...
+    final SearchResult result = client.execute(search.build());
+
+    return result.getJsonObject();
+  }
+
+  public void shutdown() {
+    try {
+      client.close();
+    } catch (IOException e) {
+      LOG.warn("Could not close client");
```
rhauch: How about including at least the exception message in the log message?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -56,52 +51,53 @@ public void start(Map<String, String> props) {
...
   // public for testing
-  public void start(Map<String, String> props, JestClient client) {
+  public void start(Map<String, String> props, ElasticsearchClient client) {
     try {
       log.info("Starting ElasticsearchSinkTask.");
 
       ElasticsearchSinkConnectorConfig config = new ElasticsearchSinkConnectorConfig(props);
-      String type = config.getString(ElasticsearchSinkConnectorConfig.TYPE_NAME_CONFIG);
-      boolean ignoreKey =
+      final String type = config.getString(ElasticsearchSinkConnectorConfig.TYPE_NAME_CONFIG);
```
kkonstantine: Were all these `final`s required by checkstyle? Otherwise it's a lot of noise that doesn't correspond to actual changes. 

Additionally to that, the connectors style does not require to use `final` keyword for local variables.

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
```diff
@@ -56,43 +56,27 @@
    * @param schema The schema used to infer mapping.
    * @throws IOException from underlying JestClient
    */
-  public static void createMapping(JestClient client, String index, String type, Schema schema)
+  public static void createMapping(ElasticsearchClient client,
+                                   String index,
```
kkonstantine: indentation style is not the one we use in connectors.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
rayokota: Yes, I'll add a comment

pom.xml
```diff
@@ -101,13 +100,41 @@
...
+        <!-- For ES 2.x -->
         <dependency>
             <groupId>org.elasticsearch</groupId>
             <artifactId>elasticsearch</artifactId>
             <version>${es.version}</version>
             <scope>test</scope>
             <type>test-jar</type>
         </dependency>
+        <!-- For ES 5.x (requires Java 8) -->
+        <!--
```
kkonstantine: nit: I'm assuming the commented out section will be removed before merging.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
rayokota: No cannot be null, I'll add a comment and a check.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
rayokota: I'll add it.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
rayokota: I'll add it.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
rayokota: Good idea, I'll do this.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -16,13 +16,15 @@
...
 
+import io.confluent.connect.elasticsearch.bulk.BulkProcessor;
 import org.apache.kafka.common.utils.SystemTime;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+
```
kkonstantine: nit: I'm guessing unneeded but formatting with our style will tell.

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
```diff
@@ -16,12 +16,11 @@
 
 package io.confluent.connect.elasticsearch;
 
-import com.google.gson.JsonObject;
-
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-
+import com.google.common.annotations.VisibleForTesting;
```
kkonstantine: Although guava is generally available via other dependencies we import, we avoid importing it explicitly as a dependency in connectors. We might change this in the future, but that's the current status. I'm not feeling very strong about this, but probably a decision to use it should be based on a more essential use case than this annotation.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.kafka.connect.data.Schema;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Set;
+
+public interface ElasticsearchClient {
+
+  enum Version {
+    ONE, TWO, FIVE, SIX
```
kkonstantine: When imported statically (see below) it's not immediately obvious that this numbering is referring to ES versions. 

Not sure what the code style says about number in variable, but maybe something like: 
`ES_V1, ES_V2, ES_V5, ES_V6`
or
`ES1, ES2, ES5, ES6` wouldn't be too terrible. 

Not crazy about any of the alternatives. Up to you.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkRequest.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ **/
+
+package io.confluent.connect.elasticsearch.bulk;
+
+public interface BulkRequest {
```
kkonstantine: How about a javadoc that this is a marker interface, and what is it for?

src/main/java/io/confluent/connect/elasticsearch/jest/JestBulkRequest.java
```diff
@@ -0,0 +1,33 @@
...
+ **/
+
+package io.confluent.connect.elasticsearch.jest;
+
+import io.confluent.connect.elasticsearch.bulk.BulkRequest;
+import io.searchbox.core.Bulk;
+
+public class JestBulkRequest implements BulkRequest {
+
+  private Bulk bulk;
```
kkonstantine: Should this be `final`?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
kkonstantine: I tend to prefer `protected` unless I have a good reason to leave package-private. Several refactorings in hdfs and storage-common showed that something might need to be used that way for testing but not reside exactly in the same package (for various reasons). Not feeling too strong about this though.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -0,0 +1,313 @@
...
+      throw new ConnectException(
+              "Couldn't start ElasticsearchSinkTask due to configuration error:",
+              e
+      );
+    }
+  }
+
+  private Version getServerVersion() throws IOException {
+    // Default to newest version for forward compatibility
+    Version defaultVersion = Version.SIX;
```
kkonstantine: Another example where I'm not 100% immediately we are referring to ES version.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -0,0 +1,313 @@
...
+
+    NodesInfo info = new NodesInfo.Builder().addCleanApiParameter("version").build();
+    JsonObject result = this.client.execute(info).getJsonObject();
```
kkonstantine: Again stylistic: Not sure if this was there already, but we tend to use `this` mainly within constructors.

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
rayokota: Yes, I'll change it

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
rayokota: I'll look into adding some tests for this.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
rayokota: Not currently.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
rayokota: It's forward compatible for ES 7.x+, meaning that the ES 6 functionality will be retained (using text instead of string, which only works for ES 2 and 5).

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
rayokota: Yes, it should work for all versions.  I'll make the errors more descriptive.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
rayokota: This is the current behavior, but yes, I can add an additional check before throwing the exception.

pom.xml
```diff
@@ -101,19 +100,46 @@
             <version>${lucene.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
```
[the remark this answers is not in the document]
rayokota: This is test scope and was necessary for working with the new ES test libs.  However, now that I've downgraded the ES test lib I will remove this.

pom.xml
rayokota: I was going to leave this in as when we move to Java 8 these will come back.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
rayokota: Ok, I don't have a strong preference so I'll change to ES_V1, etc.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
rayokota: These finals were recommended by checkstyle at one point but I think they are not needed any longer so I can remove.

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
rayokota: Ok, I can replace with a comment.

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
rayokota: Hmm, reformatting code with new coding style left this as is.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
rayokota: I will change it

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
kkonstantine: Frequently, to apply the actual style, you need to put in one line, select the whole line, and press Alt+Cmd+L

rayokota: When I do that, it just leaves it all on one line.  What formatting should it have?

pom.xml
rhauch: Yeah, considering we'll want to change this in a few months, I think it's fine to keep as-is.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConstants.java
```diff
@@ -34,19 +29,6 @@
   public static final String FLOAT_TYPE = "float";
   public static final String DOUBLE_TYPE = "double";
   public static final String STRING_TYPE = "string";
+  public static final String TEXT_TYPE = "text";
```
wicknicks: ES also has a `keyword` type. Should we support it?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -318,31 +313,11 @@ public void stop() {
...
-          if (!result.isSucceeded()) {
-            String msg = result.getErrorMessage() != null ? ": " + result.getErrorMessage() : "";
-            throw new ConnectException("Could not create index '" + index + "'" + msg);
-          }
-        } catch (IOException e) {
-          throw new ConnectException(e);
-        }
-      }
+    if (assignedTopics == null) {
+      throw new NullPointerException();
```
wicknicks: Can be replaced with a call to `java.util.Objects#requireNonNull(obj)`

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkRequest.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.List;
+
+/**
+ * BulkRequest is a marker interface for use with
+ * {@link io.confluent.connect.elasticsearch.ElasticsearchClient#createBulkRequest(List)} and
+ * {@link io.confluent.connect.elasticsearch.ElasticsearchClient#executeBulk(BulkRequest)}.
+ * Implementations will typically hold state comprised of instances of classes that are
+ * specific to the client library.
+ */
+public interface BulkRequest {
```
wicknicks: Should we call this `ConnectBulkRequest` to make it easier to read whenever it is used in the same class with `org.elasticsearch.action.bulk.BulkRequest`?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -0,0 +1,319 @@
...
+public class JestElasticsearchClient implements ElasticsearchClient {
+  private static final Logger LOG = LoggerFactory.getLogger(JestElasticsearchClient.class);
+
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  private final JestClient client;
+  private final Version version;
+
+  // visible for testing
+  public JestElasticsearchClient(String address) {
```
wicknicks: should we have a `@VisibleForTesting` constructor which accepts a JestClient as an argument. That way the test can construct the `JestClient` and can be mocked appropriately.

pom.xml
rayokota: Actually, these are the only changes required for Java 8.  However, I'll remove these comments for now.

pom.xml
```diff
@@ -39,11 +39,16 @@
     </scm>
 
     <properties>
+        <!--
+        <es.version>6.0.0</es.version>
+        <lucene.version>7.0.1</lucene.version>
+        <es.version>5.0.0</es.version>
+        <lucene.version>6.2.0</lucene.version>
+        -->
```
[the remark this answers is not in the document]
rayokota: I will remove.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -0,0 +1,99 @@
...
+   * @param type the type to search
+   * @return the search result
+   * @throws IOException if the client cannot execute the request
+   */
+  JsonObject search(String query, String index, String type) throws IOException;
+
+  /**
+   * Shuts down the client.
+   */
+  void close();
```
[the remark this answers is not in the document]
rayokota: I overrode it to not throw Exception

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConstants.java
rayokota: I can add it now as a placeholder.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
rayokota: Ok, I'll change it.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkRequest.java
rayokota: I'm going to leave this for now as all the other classes in this package start with Bulk..(BulkClient, BulkProcessor, BulkResponse).

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
rayokota: Yes, I'll add this as part of some unit tests for the Jest client.

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
kkonstantine: In IntelliJ, checking the settings under: _Preferences -> Code Style -> Java_
tab _Wrapping and Braces_ 
boxes for _Method alignment parameters_ and _Method call arguments_
and make sure _New line after '('_ is ticked, _Place ')' on new line_ is ticked (when that's the case I believe _Align when multiline_ does not matter, but better let it unticked too).

rayokota: Got it, thanks

rayokota:
I've added unit tests for the Jest client.

rayokota:
Thanks for the review, @rhauch , @kkonstantine , and @wicknicks !

PR #173, opened 2018-02-08
added a table of supported versions

mbannour:
Just I added a table in which reader can found the supported versions by  kafka-connect-elasticsearch .


med-ali-bannour:
[clabot:check]

rhauch:
This is a good thing, and in fact it'd be nice to actually add to the RST documentation as well. Plus, @rayokota has done more testing, so it'd be good to get all of the supported versions (see #169).

PR #177, opened 2018-02-23
CC-1550: Clarify ES 6 compatibility

rayokota:
Add note to state that `schema.ignore=true` can be added for ES6


rayokota:
Yes, so run pint and then hand-merge on 4.1.x like you described?

PR #187, opened 2018-03-23
Add the possibility to use secure es with basic auth

elarib:
There is some use cases of secure ES (for example multitenancy), so the Jest client should authenticate.
This PR add the possibility to sink data to a secure ES with basic auth.

✅Code changed & tested
✅Doc changed

FYI : There are many tools used to secure Elasticsearch, for example

xPack https://www.elastic.co/guide/en/x-pack/current/security-getting-started.html
Searchguard https://docs.search-guard.com/latest/index

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
@@ -115,12 +115,18 @@ public class JestElasticsearchClient implements ElasticsearchClient {
       final int readTimeout = config.getInt(
           ElasticsearchSinkConnectorConfig.READ_TIMEOUT_MS_CONFIG);
 
+      String username =
+              config.getString(ElasticsearchSinkConnectorConfig.CONNECTION_USERNAME_CONFIG);
+      String password =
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -139,6 +154,21 @@ public JestElasticsearchClient(Map<String, String> props) {
...
+   * This method generate HttpHost for each elastic url, so we can permit it for the jest client
+   */
+  private void updateHttpHosts(List<String> esUrls) {
+    for (String esUrl: esUrls) {
+      try {
+        URL url = new URL(esUrl);
+        HttpHost host = new HttpHost(url.getHost(), url.getPort(), url.getProtocol());
+        hostSet.add(host);
+      } catch (MalformedURLException e) {
+        e.printStackTrace();
```
mwl: Print exception together with an error message rather than just dumping it?
```java
log.warn("Failed to parse malformed url", e);
```

fmeeuw:
Would be nice to add support for basic auth to elasticsearch by finishing this PR! (or one of the others) Can we expect it anytime soon?

rhauch:
Sorry, it seems like we've had a number of duplicates. 

**Basic auth has already been merged via #216**, so I'm closing this issue.

PR #112, opened 2017-08-23
Secure Elastic Authentication

jdbranham:

Issue #45, opened 2016-10-27
Shield Auth Support


michaeltravisuk:
Is there support for HTTP authentication with an Elasticsearch cluster protected by Shield?

If not, is there a way of handling the authentication elsewhere, i.e. an Elasticsearch client node or similar?

shikhar:
I'm afraid we don't support Shield authentication or HTTPS currently from the connector itself. We'd definitely like to add that.

> If not, is there a way of handling the authentication elsewhere, i.e. an Elasticsearch client node or similar?

It's possible to have a plain->secure HTTP proxy in the middle, I haven't looked into the mechanics or whether such a setup can be made secure.

ave19:
Just dropping a note, +1.  Our environment requires authentication and HTTPS.  Organizations that use the Elastic stack operationally will buy support, probably, and enable the X Pack plugins which require these features to connect.  (Also, Elastic Cloud Enterprise.)  We have developers in house who could work on something like this.  Did you have any plan for implementation?  I know adding PKI bits is a lot of infrastructure.

sscarduzio:
+1 

You mentioned you want to add this, but Is there a timeline?

PrimeAC:
+1

ZeluisFS:
+1

sscarduzio:
I ended up starting to implement this myself. See #111

rayokota:
Fixed by https://github.com/confluentinc/kafka-connect-elasticsearch/pull/215

PR #215, opened 2018-05-30
CC-374 Support basic auth with secure Elasticsearch

rayokota:
This provides the ability to configure username and password for accessing a secure Elasticsearch cluster.


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -33,6 +33,16 @@
   private static final String CONNECTION_URL_DOC =
       "List of Elasticsearch HTTP connection URLs e.g. ``http://eshost1:9200,"
       + "http://eshost2:9200``.";
+  public static final String CONNECTION_USERNAME_CONFIG = "connection.username";
+  private static final String CONNECTION_USERNAME_DOC =
+      "The username used to authenticate with Elasticsearch. "
+      + "The default is the empty string, and authentication will only be performed if "
+      + " both the username and password are non-empty.";
```
rhauch: We don't want to authenticate when the username is non-null, regardless of the password?

rayokota: You mean allow for an empty username or password?  It's a bad security practice.  Do you think some customers will want it?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -108,22 +110,36 @@ public JestElasticsearchClient(String address) {
...
-      JestClientFactory factory = new JestClientFactory();
-      factory.setHttpClientConfig(new HttpClientConfig.Builder(address)
+      HttpClientConfig.Builder builder = new HttpClientConfig.Builder(address)
           .connTimeout(connTimeout)
           .readTimeout(readTimeout)
-          .multiThreaded(true)
-          .build()
-      );
+          .multiThreaded(true);
+      if (username != null && !username.isEmpty() && password != null && !password.isEmpty()) {
```
mageshn: If we are doing only Basic Auth should we be we using preemptive auth?

rayokota: Yes, it is using preemptive auth on line 138.

rayokota:
@rhauch , I made a change to allow for username and/or password to be empty

rhauch:
@rayokota please be sure to update #45

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -158,6 +168,26 @@ private static void addConnectorConfigs(ConfigDef configDef) {
...
+        null,
+        Importance.MEDIUM,
+        CONNECTION_USERNAME_DOC,
+        group,
+        ++order,
+        Width.SHORT,
+        "Connection Username"
+    ).define(
+        CONNECTION_PASSWORD_CONFIG,
+        Type.STRING,
```
kkonstantine: This needs to change to `Type.PASSWORD` so it won't be leaked in configs.

kkonstantine:
Type for field password is wrong. PR needs to be amended with a follow-up PR.

fmeeuw:
Great that this pr has been merged! I was wondering if its possible to backport this pr (and the other dependent ones) to a version 4.1.x release?

rhauch:
@fmeeuw we generally try to avoid backporting features to earlier versions. Is there a reason you wouldn't want to upgrade to 5.0.0 when it is out? Other than dropping support for JDK 7, this is so far the only change to the ES connector since 4.1.0.

cricket007:
Close #112 ?

PR #216, opened 2018-05-30
CC-2037 Support keyword subfields for text types

rayokota:
Enable string fields for aggregation in ES 6 using keyword subfield. This mimics the default functionality of ES 6.


rmoff:
Yay! \o/

PR #220, opened 2018-06-30
CC-7146 : Introduce an option to control HTTP-level compression

catap:
This code uses jest-http client that's supported Gzip compression over HTTP and this commit introduced an option to control this compression.

By default compression is disabled and this doesn't change any behaviour until it is enabled by hand.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -142,6 +142,13 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
       + "if any read operation times out, and will need to be restarted to resume "
       + "further operations.";
 
+  public static final String CONNECTION_COMPRESSION_CONFIG = "connection.compression";
+  private static final String CONNECTION_COMPRESSION_DOC = "Whether to use GZip compression on "
+          + "HTTP connection to ElasticSearch. Valid options are ``true`` and ``false``. "
+          + "Default is ``false``. To make this setting to work "
+          + "the ``http.compression`` setting also needs to be enabled at the Elasticsearch nodes "
+          + "or the load-balancer before using it.";
+
   public static final String BEHAVIOR_ON_NULL_VALUES_CONFIG = "behavior.on.null.values";
   private static final String BEHAVIOR_ON_NULL_VALUES_DOC = "How to handle records with a "
       + "non-null key and a null value (i.e. Kafka tombstone records). Valid options are "
@@ -310,6 +317,16 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         ++order,
         Width.SHORT,
         "Retry Backoff (ms)"
+      ).define(
+        CONNECTION_COMPRESSION_CONFIG,
+        Type.BOOLEAN,
+        false,
+        Importance.LOW,
+        CONNECTION_COMPRESSION_DOC,
+        group,
+        ++order,
+        Width.SHORT,
+        "Compression"
       ).define(
         CONNECTION_TIMEOUT_MS_CONFIG, 
         Type.INT, 
--- a/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
@@ -169,10 +169,14 @@ public class JestElasticsearchClient implements ElasticsearchClient {
     List<String> address = config.getList(
         ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG);
 
+    final boolean compressionEnabled = config.getBoolean(
+            ElasticsearchSinkConnectorConfig.CONNECTION_COMPRESSION_CONFIG);
+
     HttpClientConfig.Builder builder =
         new HttpClientConfig.Builder(address)
             .connTimeout(connTimeout)
             .readTimeout(readTimeout)
+            .requestCompressionEnabled(compressionEnabled)
             .multiThreaded(true);
     if (username != null && password != null) {
       builder.defaultCredentials(username, password.value())
--- a/src/test/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClientTest.java
+++ b/src/test/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClientTest.java
@@ -72,6 +72,7 @@ import static org.hamcrest.Matchers.equalTo;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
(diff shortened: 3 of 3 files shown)


src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -128,11 +128,15 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
       final Password password = config.getPassword(
           ElasticsearchSinkConnectorConfig.CONNECTION_PASSWORD_CONFIG);
 
+      final boolean compression = config.getBoolean(
```
stanislavkozlovski: nit: Maybe change the variable name to `compressionEnabled` for better clarity, although it's questionable if it's worth it as it's a short-lived variable

catap: @stanislavkozlovski, I guess that correct name is `compressionIsEnabled` because it is should be a passive voice because an user is enable an compression.

But I don't think that it is important for this case ;)

catap: @stanislavkozlovski I agree with you because fiend name is `requestCompressionEnabled` and it sounds logical to rename this variable.

stanislavkozlovski: yep absolutely, `compressionIsEnabled` is better

catap: @stanislavkozlovski but it requests also to rename a field and getter/setter at jest-http client that looks overengineering for this goal and that also broke back compatibility.

So, I prefere to use incorrect value but similar to name of setter ;)

purbon:
@catap @stanislavkozlovski, anything else pending here to move this PR? the use of compression is really interesting.

@catap  you might need to rebase this PR as it has now come conflicts on the documentation side.

purbon:
@catap I am wrong, or the test for this feature, as well as others, does not check if the indexing with compression actually works? would you mind as well writing an integration test for this is possible? that  would be great.

catap:
@purbon it is compression at HTTP level. I don't think it is easy to test.

anyway, this code works for 6 month without any issues ;)

BTW I will rebase it on this week

catap:
@purbon rebased.

I removed docs part because someone moved it to somewhere else at this commit https://github.com/confluentinc/kafka-connect-elasticsearch/commit/2a2908db8d43d5eb9e5f182d4d345c4826f042dd

But I'd like to keep old docs here in the case if you can add it to the new place:
```
``connection.compression``
   Defines to use GZip compression on HTTP connection to ElasticSearch. Valid options are ``true`` and ``false``.

   * Type: boolean
   * Default: false
   * Importance: low


```

yzhangdemon:
Any updates on if this PR can be updated and merged?
We're really hoping to see this feature get released.

catap:
@yzhangdemon let me rebase it one more time :)

catap:
@yzhangdemon it is possible to merge without conflict now.

yzhangdemon:
Thanks @catap .
@purbon @stanislavkozlovski Would you able to give it one more review and merge if it looks good on your end?

catap:
@yzhangdemon any news?

We're using it in production more for a year now, maybe let it merge into master ;) ?

stanislavkozlovski:
@catap sorry for the delay here. I am not the best-suited person to review this. Let me bring it up with the appropriate team

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -142,6 +142,11 @@
       + "if any read operation times out, and will need to be restarted to resume "
       + "further operations.";
 
+  public static final String CONNECTION_COMPRESSION_CONFIG = "connection.compression";
+  private static final String CONNECTION_COMPRESSION_DOC = "Defines to use GZip compression on "
```
levzem:
```suggestion
  private static final String CONNECTION_COMPRESSION_DOC = "Whether to use GZip compression on "
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -142,6 +142,11 @@
...
+          + "HTTP connection to ElasticSearch. Valid options are ``true`` and ``false``. "
+          + "Default is ``false``";
```
levzem: /**
```suggestion
          + "Default is ``false``. The ``http.compression`` setting also needs to be set on the Elasticsearch nodes for this to work."; 
```

src/test/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClientTest.java
```diff
@@ -125,6 +125,27 @@ public void connectsSecurely() {
     assertEquals(HttpHost.create("http://localhost:9200"), preemptiveAuthTargetHosts.iterator().next());
   }
 
+  @Test
+  public void compressedConnectsSecurely() {
```
levzem: this mimics all the other tests - let's at least check that the compression was set

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
catap: @levzem or at load-balancer behind the ES node.

catap: Also base on https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-http.html it is enabled by default and I suggest to don't add this line.

Are you agree?

catap:
@levzem I've fixed your comments. Tests was rewritten and the branch was rebased to the last master to fix all conflicts.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
levzem: I think it's safer to document this just in case.

levzem:
@catap looks good but we should rebase to `5.0.x` after that LGTM

catap:
@levzem I've rebased it to master. It isn't enough?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
catap: it's already done.

levzem:
> @levzem I've rebased it to master. It isn't enough?

@catap we would like to merge this into `5.0.x` so that we can backport it and then propogate it all the way to `master`. I can do it for you if you like. Should be relatively painless however

catap:
@levzem it won't be an easy because master contains https://github.com/confluentinc/kafka-connect-elasticsearch/commit/7ad54d8847f61b2eb862750750eda263efb8f9da

and 5.3.x+ also contains it.

And this commit is root cause of conflicts that I've rebased.

So. I can rebase it to 5.0.x and you can easy migrate it to 5.1.x and 5.2.x; but at 5.3.x+ you will have a conflict :)

What should we do?

levzem:
@catap in that case, master should be sufficient. Thanks for your contribution!

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -142,6 +142,13 @@
       + "if any read operation times out, and will need to be restarted to resume "
       + "further operations.";
 
+  public static final String CONNECTION_COMPRESSION_CONFIG = "connection.compression";
+  private static final String CONNECTION_COMPRESSION_DOC = "Whether to use GZip compression on "
+          + "HTTP connection to ElasticSearch. Valid options are ``true`` and ``false``. "
+          + "Default is ``false``. To make this setting to work "
+          + "the ``http.compression`` setting also needs to be enabled at the Elasticsearch nodes "
+          + "or the load-balancer before the Elasticsearch if you use it.";
```
levzem:
```suggestion
          + "or the load-balancer before using it";
```

catap: done

levzem:
@catap merged, thank you for your contribution!

PR #233, opened 2018-08-27
DOCS-618 - Refactor Connect docs

joel-hamill:
- simplifies doc design by moving the child elasticsearch_connector.rst page contents to top-level index.rst
- adds text string replace for common terms (e.g. Kafka Connect)
- changes titles to match naming shown at https://www.confluent.io/hub/#
- redirect added here https://github.com/confluentinc/docs/pull/1146/files#diff-fb0d0d7ff27f33697ba3b787ec894677R448


docs/changelog.rst
```diff
@@ -19,7 +19,7 @@ No changes
 Version 4.1.1
 -------------
 
-* `PR-191 <https://github.com/confluentinc/kafka-connect-elasticsearch/pull/191>`_ - CC-918 Add Kafka Connect Maven plugin to build
+* `PR-191 <https://github.com/confluentinc/kafka-connect-elasticsearch/pull/191>`_ - CC-918 Add |kconnect| Maven plugin to build
```
kkonstantine: My previous comment was lost. Re-writing here: 

These are mostly autogenerated entries from the PR titles. Adding the aliases instead will require always a second pass (unless it's integrated to the changelog command). And this seems that will incur extra burden. 

We only change the changelog to fix typos or improve the title significantly when needed. 
If feel we should not use aliases here. Just a thought.

joel-hamill: okay then feel free to ignore

PR #250, opened 2018-10-23
Fix bug when topic names are not in all lowercase

purbon:
As described in (https://github.com/confluentinc/kafka-connect-elasticsearch/issues/246#issuecomment-431794980), Elasticsearch can only create indexes in lowercase letters, this PR fix this by making sure that all indexes are created that way.

Please forgive me If I miss a necessary check before pushing the PR, is the first time I contribute a change. 

Fix partially #246

Issue #246, opened 2018-10-12
Regex router not applied


gunnarmorling:
I'm trying to use the regex router to lower-case the index name for a given topic. But the router isn't applied, causing an exception as the original topic name isn't a valid index name:

```
Invalid index name [dbserver1.inventory.Hike], must be lowercase\",\"index_uuid\":\"_na_\",\"index\":\"dbserver1.inventory.Hike\"} at 
io.confluent.connect.elasticsearch.jest.JestElasticsearchClient.createIndices(JestElasticsearchClient.java:238) at 
io.confluent.connect.elasticsearch.ElasticsearchWriter.createIndicesForTopics(ElasticsearchWriter.java:330) at 
io.confluent.connect.elasticsearch.ElasticsearchSinkTask.open(ElasticsearchSinkTask.java:157)
```

That's the connector config:

```
{
    "name": "elastic-sink",
    "config": {
        "connector.class": "io.confluent.connect.elasticsearch.ElasticsearchSinkConnector",
        "tasks.max": "1",
        "topics": "Customers",
        "connection.url": "http://elastic:9200",
        "transforms": "unwrap,key,route",
        "transforms.unwrap.type": "io.debezium.transforms.UnwrapFromEnvelope",
        "transforms.key.type": "org.apache.kafka.connect.transforms.ExtractField$Key",
        "transforms.key.field": "id",
        "transforms.route.type": "org.apache.kafka.connect.transforms.RegexRouter",
        "transforms.route.regex": "Customers",
        "transforms.route.replacement": "customers",
        "key.ignore": "false",
        "type.name": "customer"
    }
}
```

I have also an executable reproducer, if you'd like to take a look: https://github.com/gunnarmorling/debezium-examples/tree/kafka-connect-elasticsearch-99/unwrap-smt

Check out that branch and follow the steps in https://github.com/gunnarmorling/debezium-examples/blob/kafka-connect-elasticsearch-99/unwrap-smt/reproducer.md. Registering the sink connector fails due to an upper-case index name, which should have been transformed by means of the regex router. This is with Kafka 2.0 and ES sink 5.0.0.

purbon:
Hi, thanks for reporting this @gunnarmorling , certainly this is a bug, however a quick test where I was able to verify the problem, I am not sure this is a problem of transformations. From my understanding the transformations are used to change event per event, however the plugin try to create the indexes at startup time, this is what I understand from:

```java
[...]

PR #251, opened 2018-10-30
[Backport ready] Fix bug when topic names are not in all lowercase

purbon:
As described in (https://github.com/confluentinc/kafka-connect-elasticsearch/issues/246#issuecomment-431794980), Elasticsearch can only create indexes in lowercase letters, this PR fix this by making sure that all indexes are created that way.

Please forgive me If I miss a necessary check before pushing the PR, is the first time I contribute a change. 

Fix partially #246 

Backport of #250 as requested in https://github.com/confluentinc/kafka-connect-elasticsearch/pull/250#pullrequestreview-167503044

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
@@ -189,7 +189,6 @@ public class ElasticsearchWriter {
   }
 
   public void write(Collection<SinkRecord> records) {
-
     for (SinkRecord sinkRecord : records) {
       final String index = convertTopicToIndexName(sinkRecord.topic());
       final boolean ignoreKey = ignoreKeyTopics.contains(sinkRecord.topic()) || this.ignoreKey;
@@ -211,9 +210,7 @@ public class ElasticsearchWriter {
 
       bulkProcessor.add(indexableRecord, flushTimeoutMs);
     }
-
   }
-
 
   /**
    * Return the expected index name for a given topic, using the configured mapping or the topic name. Elasticsearch
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
@@ -214,18 +214,15 @@ public class ElasticsearchWriter {
 
   }
 
+
   /**
-   * Return the expected index name for a given topic. It makes sure to lowercase the topic name
-   * as elasticsearch require all index names to be lowercase.
-   * See https://github.com/confluentinc/kafka-connect-elasticsearch/issues/246 for connector details.
-   * See https://github.com/elastic/elasticsearch/issues/29420 for elasticsearch details.
-   * @param topic The topic name being proceed.
-   * @return String A valid elasticsearch index name
+   * Return the expected index name for a given topic, using the configured mapping or the topic name. Elasticsearch
+   * <a href="https://github.com/elastic/elasticsearch/issues/29420">accepts only lowercase index names</a>.
    */
   private String convertTopicToIndexName(String topic) {
     final String indexOverride = topicToIndexMap.get(topic);
     String index = indexOverride != null ? indexOverride : topic.toLowerCase();
-    log.debug("Topic " + topic + " was translated as index name " + index);
+    log.debug("Topic '{}' was translated as index '{}'", topic, index);
     return index;
   }
 
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
@@ -189,9 +189,9 @@ public class ElasticsearchWriter {
   }
 
   public void write(Collection<SinkRecord> records) {
+
     for (SinkRecord sinkRecord : records) {
-      final String indexOverride = topicToIndexMap.get(sinkRecord.topic());
-      final String index = indexOverride != null ? indexOverride : sinkRecord.topic();
+      final String index = convertTopicToIndexName(sinkRecord.topic());
       final boolean ignoreKey = ignoreKeyTopics.contains(sinkRecord.topic()) || this.ignoreKey;
       final boolean ignoreSchema = ignoreSchemaTopics.contains(sinkRecord.topic()) || this.ignoreSchema;
 
@@ -211,6 +211,22 @@ public class ElasticsearchWriter {
 
       bulkProcessor.add(indexableRecord, flushTimeoutMs);
     }
+
+  }
+
+  /**
+   * Return the expected index name for a given topic. It makes sure to lowercase the topic name
+   * as elasticsearch require all index names to be lowercase.
+   * See https://github.com/confluentinc/kafka-connect-elasticsearch/issues/246 for connector details.
+   * See https://github.com/elastic/elasticsearch/issues/29420 for elasticsearch details.
+   * @param topic The topic name being proceed.
+   * @return String A valid elasticsearch index name
+   */
+  private String convertTopicToIndexName(String topic) {
+    final String indexOverride = topicToIndexMap.get(topic);
+    String index = indexOverride != null ? indexOverride : topic.toLowerCase();
+    log.debug("Topic " + topic + " was translated as index name " + index);
+    return index;
   }
 
   public void flush() {
@@ -261,12 +277,7 @@ public class ElasticsearchWriter {
   private Set<String> indicesForTopics(Set<String> assignedTopics) {
     final Set<String> indices = new HashSet<>();
     for (String topic : assignedTopics) {
-      final String index = topicToIndexMap.get(topic);
-      if (index != null) {
-        indices.add(index);
-      } else {
-        indices.add(topic);
-      }
+      indices.add(convertTopicToIndexName(topic));
     }
     return indices;
   }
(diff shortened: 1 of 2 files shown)


src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTaskTest.java
```diff
@@ -72,4 +79,39 @@ public void testPutAndFlush() throws Exception {
...
+            key,
+            schema,
+            record,
+            0 );
+
+    try {
+      task.start(props, client);
+      task.open(new HashSet<>(Collections.singletonList(TOPIC_IN_CAPS_PARTITION)));
+      task.put(Collections.singleton(sinkRecord));
+      assertTrue("A topic name not in lowercase was created in Elasticsearch", true);
```
rhauch: This assertion doesn't really check anything, since you're always passing `true` into the method. Is this intentional?

Don't we really want to check whether the index was created in the ES cluster? Without that, how are we sure that this functionality works?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -211,6 +211,22 @@ public void write(Collection<SinkRecord> records) {
...
       bulkProcessor.add(indexableRecord, flushTimeoutMs);
     }
+
+  }
+
+  /**
+   * Return the expected index name for a given topic. It makes sure to lowercase the topic name
+   * as elasticsearch require all index names to be lowercase.
+   * See https://github.com/confluentinc/kafka-connect-elasticsearch/issues/246 for connector details.
+   * See https://github.com/elastic/elasticsearch/issues/29420 for elasticsearch details.
```
rhauch: Minor: How about
```
   * Return the expected index name for a given topic, using the configured mapping or the topic name. Elasticsearch
   * <a href="https://github.com/elastic/elasticsearch/issues/29420">accepts only lowercase index names</a>.
 ```

File: src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java [hunk 3]
```diff
@@ -211,6 +211,22 @@ public void write(Collection<SinkRecord> records) {
...
+   * @param topic The topic name being proceed.
+   * @return String A valid elasticsearch index name
+   */
+  private String convertTopicToIndexName(String topic) {
+    final String indexOverride = topicToIndexMap.get(topic);
+    String index = indexOverride != null ? indexOverride : topic.toLowerCase();
```
rhauch:
```suggestion
    String index = indexOverride != null ? indexOverride : topic.toLowerCase(StandardCharsets.UTF_8);
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -211,6 +211,22 @@ public void write(Collection<SinkRecord> records) {
...
+    log.debug("Topic " + topic + " was translated as index name " + index);
```
rhauch: The following will avoid having to create the string when debug is not enabled:
```suggestion
    log.debug("Topic '{}' was translated as index '{}'", topic, index);
```

The challenge, of course, is that this log message is written out for every record. What do you think about keeping another cache (map) of the topic names to index name that were used so far, and then logging this message at INFO level when the topic had not yet been seen? It's a bit more complicated, but it would be much better to have that INFO level log message. (The cache really is more for bookkeeping than for performance.)

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTaskTest.java
purbon: If you try to create an index with caps an exception will be throw, should look something like:

```java
org.apache.kafka.connect.errors.ConnectException: Could not create index 'AnotherTopicInCaps': {"root_cause":[{"type":"invalid_index_name_exception","reason":"Invalid index name [AnotherTopicInCaps], must be lowercase","index":"AnotherTopicInCaps"}],"type":"invalid_index_name_exception","reason":"Invalid index name [AnotherTopicInCaps], must be lowercase","index":"AnotherTopicInCaps"}
	at __randomizedtesting.SeedInfo.seed([A1BAC01989C87C:D2B7A48645E95AFE]:0)
	at io.confluent.connect.elasticsearch.ElasticsearchWriter.createIndicesForTopics(ElasticsearchWriter.java:268)
	at io.confluent.connect.elasticsearch.ElasticsearchSinkTask.open(ElasticsearchSinkTask.java:120)
	at io.confluent.connect.elasticsearch.ElasticsearchSinkTaskTest.testPutWithTopicInCaps(ElasticsearchSinkTaskTest.java:106)
```
I've learn this pattern over the year, probably contaminated with ruby and other languages, so yes to answer your question. I now the assertTrue does nothing, is mostly to show the message, the real check for this to work in on the fail part.

I am happy to write it more politely and learn, please let me know how it should look? should I just remove the assert?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
purbon: fair for me.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
purbon: lowercase accepts locale, but not Charset, right? we might like to add this kind of encoding validation before we reach here, isn't?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
purbon: re the cache, something to think about it, certainly a good idea, but I think for this case and how we create indexes right now (only at bootstrap time), we are good the way we are. what do you think?

I am thinking for when we allow for an unseen index/topic to be created at the ES side.

purbon: thanks for the suggestion, happy to learn better ways of doing work.

purbon:
Hi @rhauch, did you had any chance to take a look to the changes I recently introduced to this PR? Looking forward to know your thoughts.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
rhauch: Oh, sorry. My bad. And this is actually not required, so it's okay to not change this line.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTaskTest.java
purbon: Very useful feedback @rhauch, pushing again my java testing skills 😄 forward, love that!

purbon:
@rhauch updated the PR with your last feedaback, looking forward to green light from you 😄

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -189,9 +189,9 @@ public ElasticsearchWriter build() {
   }
 
   public void write(Collection<SinkRecord> records) {
+
```
rhauch: We generally try to avoid unnecessary changes like adding/removing lines so that the diff is smaller and more focused.
```suggestion
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -211,6 +211,19 @@ public void write(Collection<SinkRecord> records) {
 
       bulkProcessor.add(indexableRecord, flushTimeoutMs);
     }
+
```
rhauch: Let's not introduce this blank line unnecessarily.
```suggestion
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -211,6 +211,19 @@ public void write(Collection<SinkRecord> records) {
...
+  }
+
```
rhauch: Let's not introduce this extra blank line unnecessarily. We prefer 1 blank line between methods.
```suggestion
```

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTaskTest.java
```diff
@@ -72,4 +79,38 @@ public void testPutAndFlush() throws Exception {
...
+    try {
+      task.start(props, client);
+      task.open(new HashSet<>(Collections.singletonList(TOPIC_IN_CAPS_PARTITION)));
+      task.put(Collections.singleton(sinkRecord));
+    } catch (Exception ex) {
+      fail("A topic name not in lowercase can not be used as index name in Elasticsearch");
+    } finally {
+      task.stop();
+    }
+
```
rhauch: Again, let's remove unnecessary blank lines.
```suggestion
```

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTaskTest.java
```diff
@@ -72,4 +79,38 @@ public void testPutAndFlush() throws Exception {
...
+  }
+
```
rhauch: One more: let's remove unnecessary blank lines.
```suggestion
```

purbon:
@rhauch applyed your last request of changes at 72f996ec3090fb1078cd8913bd9acf90fd04b901

PR #254, opened 2018-11-12
inherit test artifact versions from common pom

xvrl:


xvrl:
@C0urante sure happy to remove them. Different teams tend to have different philosophies around this, and it seemed you were using specifying properties elsewhere, so I didn't touch this.

PR #270, opened 2019-01-09
CC-2127: Partial prototype for using TLS/SSL with Elasticsearch (WIP)

rhauch:
Very rough sketch of enabling support for TLS/SSL connections to Elasticsearch. This does not use a key store (which we should do) or verify hostnames


cyrusv:
Resolved in https://github.com/confluentinc/kafka-connect-elasticsearch/pull/278

PR #274, opened 2019-01-31
Disable index creation at bootstrap time

purbon:
This PR include changes enhance the experience creating indexes in Elasticsearch based on the preferred index names and not only on the topic name.

Before this PR the connector create all indexes at start time, using the topic names as a basis for that. However there are situations when the users need to change the expected index name using an SMT. This PR include as well a backwards compatibility flag, so the users can decide if they aim to keep the initial bootstrap index creation time or disable it and rely only in the record write time operation.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -147,6 +147,11 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
       + " mapping conflict or a field name containing illegal characters. Valid options are "
       + "'ignore', 'warn', and 'fail'.";
 
+  public static final String CREATE_INDEX_AT_START_CONFIG = "create.index.at.start";
+  private static final String CREATE_INDEX_AT_START__DOC = "Create the Elasticsearch indexes at "
+      + " bootstrap time. This is useful when the indexes are a direct mapping "
+      + " of the Kafka topics.";
+
   protected static ConfigDef baseConfigDef() {
     final ConfigDef configDef = new ConfigDef();
     addConnectorConfigs(configDef);
@@ -272,10 +277,20 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         3000, 
         Importance.LOW, 
         READ_TIMEOUT_MS_CONFIG_DOC,
-        group, 
-        ++order, 
-        Width.SHORT, 
-        "Read Timeout");
+        group,
+        ++order,
+        Width.SHORT,
+        "Read Timeout"
+    ).define(
+        CREATE_INDEX_AT_START_CONFIG,
+        Type.BOOLEAN,
+        true,
+        Importance.LOW,
+        CREATE_INDEX_AT_START__DOC,
+        group,
+        ++order,
+        Width.SHORT,
+        "Create indexes at bootstrap time");
   }
 
   private static void addConversionConfigs(ConfigDef configDef) {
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -147,6 +147,11 @@
       + " mapping conflict or a field name containing illegal characters. Valid options are "
       + "'ignore', 'warn', and 'fail'.";
 
+  public static final String CREATE_INDEX_AT_START_CONFIG = "create.index.at.start";
+  private static final String CREATE_INDEX_AT_START__DOC = "Create the Elasticsearch indexes at "
```
wicknicks: extra underscore before `DOC`.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -147,6 +147,11 @@
       + " mapping conflict or a field name containing illegal characters. Valid options are "
       + "'ignore', 'warn', and 'fail'.";
 
+  public static final String CREATE_INDEX_AT_START_CONFIG = "create.index.at.start";
```
wicknicks: should this be called `auto.create.indices` and document that these indices are created when the connector is created?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
purbon: oh, shit. Good catch @wicknicks

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
purbon: works for me.

purbon:
@wicknicks resolved your comments.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -147,6 +147,11 @@
       + " mapping conflict or a field name containing illegal characters. Valid options are "
       + "'ignore', 'warn', and 'fail'.";
 
+  public static final String AUTO_CREATE_INDEX_AT_START_CONFIG = "auto.create.index.at.start";
+  private static final String AUTO_CREATE_INDEX_AT_START_DOC = "Auto create the Elasticsearch"
+      + " indexes at bootstrap time. This is useful when the indexes are a direct mapping "
```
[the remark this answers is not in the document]
purbon: works for me as well. @rayokota closing this PR now and starting a new one based on 4.1.x as per discussion with @wicknicks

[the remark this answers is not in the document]
wicknicks: that's what I was trying to get to with my previous comment as well. `auto.create.index` sounds like you are creating a single index.

purbon:
closing this PR in favour of https://github.com/confluentinc/kafka-connect-elasticsearch/pull/275 as per discussion with @wicknicks this change should be rebased on 4.1.x

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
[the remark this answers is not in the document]
purbon: fair point, updating this in #275

PR #261, opened 2018-12-05
Create indices before writing records

Mongey:
Currently, indices are only created when the connector is opened. If the
destination index is changed by a Single Message Transform, the index
may not exist when the record is being written, and as such will fail
to create a mapping, as the index does not exist.

This patch creates the index, after the message has been transformed.

Fixes #99.

PR #275, opened 2019-01-31
Disable index creation at bootstrap time

purbon:
This PR include changes enhance the experience creating indexes in Elasticsearch based on the preferred index names and not only on the topic name.

Before this PR the connector create all indexes at start time, using the topic names as a basis for that. However there are situations when the users need to change the expected index name using an SMT. This PR include as well a backwards compatibility flag, so the users can decide if they aim to keep the initial bootstrap index creation time or disable it and rely only in the record write time operation.

Proposed flow:

1. By default: create always an initial topic to index during start time. This could be deactivated if necessary.
2. Make sure that if a SMT is applied, the index is created anyway. (this does not happen in the current way of operation).

This is a backport PR from https://github.com/confluentinc/kafka-connect-elasticsearch/pull/274 as per discussion with @wicknicks

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -141,7 +141,7 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
 
   public static final String AUTO_CREATE_INDICES_AT_START_CONFIG = "auto.create.indices.at.start";
   private static final String AUTO_CREATE_INDICES_AT_START_DOC = "Auto create the Elasticsearch"
-      + " indices at bootstrap time. This is useful when the indices are a direct mapping "
+      + " indices at startup. This is useful when the indices are a direct mapping "
       + " of the Kafka topics.";
 
   protected static ConfigDef baseConfigDef() {
@@ -262,7 +262,7 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         group,
         ++order,
         Width.SHORT,
-        "Create indices at bootstrap time"
+        "Create indices at startup"
       );
   }
 
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -139,6 +139,11 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
       + " mapping conflict or a field name containing illegal characters. Valid options are "
       + "'ignore', 'warn', and 'fail'.";
 
+  public static final String CREATE_INDEX_AT_START_CONFIG = "create.index.at.start";
+  private static final String CREATE_INDEX_AT_START__DOC = "Create the Elasticsearch indexes at "
+      + " bootstrap time. This is useful when the indexes are a direct mapping "
+      + " of the Kafka topics.";
+
   protected static ConfigDef baseConfigDef() {
     final ConfigDef configDef = new ConfigDef();
     addConnectorConfigs(configDef);
@@ -244,10 +249,20 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         3000, 
         Importance.LOW, 
         READ_TIMEOUT_MS_CONFIG_DOC,
-        group, 
-        ++order, 
-        Width.SHORT, 
-        "Read Timeout");
+        group,
+        ++order,
+        Width.SHORT,
+        "Read Timeout"
+    ).define(
+        CREATE_INDEX_AT_START_CONFIG,
+        Type.BOOLEAN,
+        true,
+        Importance.LOW,
+        CREATE_INDEX_AT_START__DOC,
+        group,
+        ++order,
+        Width.SHORT,
+        "Create indexes at bootstrap time");
   }
 
   private static void addConversionConfigs(ConfigDef configDef) {
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -139,9 +139,9 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
       + " mapping conflict or a field name containing illegal characters. Valid options are "
       + "'ignore', 'warn', and 'fail'.";
 
-  public static final String CREATE_INDEX_AT_START_CONFIG = "create.index.at.start";
-  private static final String CREATE_INDEX_AT_START__DOC = "Create the Elasticsearch indexes at "
-      + " bootstrap time. This is useful when the indexes are a direct mapping "
+  public static final String AUTO_CREATE_INDICES_AT_START_CONFIG = "auto.create.indices.at.start";
+  private static final String AUTO_CREATE_INDICES_AT_START_DOC = "Auto create the Elasticsearch"
+      + " indices at bootstrap time. This is useful when the indices are a direct mapping "
       + " of the Kafka topics.";
 
   protected static ConfigDef baseConfigDef() {
@@ -254,15 +254,16 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         Width.SHORT,
         "Read Timeout"
     ).define(
-        CREATE_INDEX_AT_START_CONFIG,
+        AUTO_CREATE_INDICES_AT_START_CONFIG,
         Type.BOOLEAN,
         true,
         Importance.LOW,
-        CREATE_INDEX_AT_START__DOC,
+        AUTO_CREATE_INDICES_AT_START_DOC,
         group,
         ++order,
         Width.SHORT,
-        "Create indexes at bootstrap time");
+        "Create indices at bootstrap time"
+      );
   }
 
   private static void addConversionConfigs(ConfigDef configDef) {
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -253,6 +253,8 @@ public void write(Collection<SinkRecord> records) {
       final boolean ignoreSchema =
           ignoreSchemaTopics.contains(sinkRecord.topic()) || this.ignoreSchema;
 
+      client.createIndices(Collections.singleton(index));
```
wicknicks: just saw this one. I think this violates the `on.start` in the documentation.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -244,10 +249,20 @@ private static void addConnectorConfigs(ConfigDef configDef) {
...
+    ).define(
+        AUTO_CREATE_INDICES_AT_START_CONFIG,
+        Type.BOOLEAN,
+        true,
+        Importance.LOW,
+        AUTO_CREATE_INDICES_AT_START_DOC,
+        group,
+        ++order,
+        Width.SHORT,
+        "Create indexes at bootstrap time");
```
wicknicks: indexes -> indicies.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
purbon: I don't think so, this makes sure that indices are created during the write operations. Imagine the situation where you are using a regexp SMT to rewrite the topicToIndexName, this indices would not be created at start time.

Actually this made me realise the position of this call is wrong, it should always be done after the mapping is created, otherwise elasticsearch will create them "automatically".

Happy to collaborate with you.

purbon: To extend this, I'm proposing this flow with this PR.

1. By default: create always an initial topic to index during start time. This could be deactivated if necessary.
2. Make sure that if a SMT is applied, the index is created anyway. (this does not happen in the current way of operation).

rayokota:
@purbon, `testCreateAndWriteToIndexNotCreatedAtStartTime(io.confluent.connect.elasticsearch.ElasticsearchSinkTaskTest)` seems to be failing

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
wicknicks: Thanks @purbon for describing the motivation, but here are a couple of things to consider:

1. This call forces an index creation request to ES for every message consumed by every sink task. That's going to be expensive, and will reduce connector throughput.
2. If we are going to create indices per record, then we don't really ever need to create them at bootstrap time. One of those is redundant. 

IMHO, for this PR, we should punt the index-creates-at-write-time to a future PR since that might need deeper changes (or follow up with PR https://github.com/confluentinc/kafka-connect-elasticsearch/pull/261).

purbon:
@rayokota sad thing, this is what happen when you do a change after a long flight and jet lag. I apologies for the issue.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
purbon: Hi @wicknicks,
  speaking about 1, yes this is inevitable, it will be an extra over head, the only thing we can do here is to manage our internal index name cache, as proposed in #261, however this will as well include more extra challenges.  
  i included 2 as a backwards compatible feature. we can decide to don't do it, but this was my intention.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTaskTest.java
```diff
@@ -94,7 +99,38 @@ public void testCreateAndWriteToIndexForTopicWithUppercaseCharacters() {
...
+    } finally {
+      task.stop();
+    }
+  }
+
+  @Test
+  public void testCreateAndWriteToIndexNotCreatedAtStartTime() {
+    InternalTestCluster cluster = ESIntegTestCase.internalCluster();
+    cluster.ensureAtLeastNumDataNodes(3);
+    Map<String, String> props = createProps();
```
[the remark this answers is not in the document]
purbon: It is triggering the same behaviour, but good catch! I added it.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTaskTest.java
```diff
@@ -106,7 +142,7 @@ public void testCreateAndWriteToIndexForTopicWithUppercaseCharacters() {
       task.open(new HashSet<>(Collections.singletonList(TOPIC_IN_CAPS_PARTITION)));
       task.put(Collections.singleton(sinkRecord));
     } catch (Exception ex) {
-      fail("A topic name not in lowercase can not be used as index name in Elasticsearch");
+      fail("Record could not be written to elasticsearch due to non existing index");
```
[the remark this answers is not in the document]
purbon: fair point, apologies for it.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -187,13 +190,16 @@ public Version getVersion() {
   }
 
   private boolean indexExists(String index) {
-    Action action = new IndicesExists.Builder(index).build();
-    try {
-      JestResult result = client.execute(action);
-      return result.isSucceeded();
-    } catch (IOException e) {
-      throw new ConnectException(e);
+    if (!indexCache.keySet().contains(index)) {
```
wicknicks: couple of points here:

1. Since indexCache is only tested for the presence of a value, we probably only need a HashSet<> here. 
2. Making a keySet everytime is expensive. We can check for membership with `containsKey(key)`.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
wicknicks: This is good now. We have a cache that keeps track of created indices.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -71,6 +72,8 @@
   private final JestClient client;
   private final Version version;
 
+  private final Map<String, Boolean> indexCache = new ConcurrentHashMap<>();
```
wicknicks: since this is local to each task and read and updated in a single method, we could get away with just `HashMap<>()`?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -209,6 +215,7 @@ public void createIndices(Set<String> indices) {
               throw new ConnectException("Could not create index '" + index + "'" + msg);
             }
           }
+          indexCache.put(index, true);
```
wicknicks: we should check if `indexCache` contains any or all of indices before executing operations on the ES cluster.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
purbon: Good, happy to be of help.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
purbon: fair point, thanks @wicknicks here are my rusty java skills at display.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
purbon: fair point, I somehow put the concurrent behviour as there is the public method:

```java
  public void start(Map<String, String> props, ElasticsearchClient client) {
```

apologies for that.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
purbon: 👍

purbon:
Finally time to get the PR updates, thank @wicknicks for your feedback.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -187,13 +189,16 @@ public Version getVersion() {
...
-    } catch (IOException e) {
-      throw new ConnectException(e);
+    if (!indexCache.contains(index)) {
+      Action action = new IndicesExists.Builder(index).build();
+      try {
+        JestResult result = client.execute(action);
+        return result.isSucceeded();
+      } catch (IOException e) {
+        throw new ConnectException(e);
+      }
```
wicknicks: nit: I'd write this code as:

```
if (indexCache.contains(index)) {
  return true;
} 

Action action = .... // the old code continues from here.
```

It's easier to read, and makes the changes look more succinct.

purbon:
@wicknicks I think I fixed all your last concerns, hopefully is good to go now. 😄

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -244,10 +249,21 @@ private static void addConnectorConfigs(ConfigDef configDef) {
...
+    ).define(
+        AUTO_CREATE_INDICES_AT_START_CONFIG,
+        Type.BOOLEAN,
+        true,
+        Importance.LOW,
+        AUTO_CREATE_INDICES_AT_START_DOC,
+        group,
+        ++order,
+        Width.SHORT,
+        "Create indices at bootstrap time"
```
rhauch: As a user, I might be unsure what "bootstrap" means, since it's not a term that Connect really uses. Can we just use "Create indices at startup"?

purbon: done, sorry for the confusion.

wicknicks:
merging this PR.

PR #277, opened 2019-02-02
CC-7644: Add support for http(s) proxy 

sslavian812:
One can specify a proxy with `connection.proxy` to connect to elastic-search endpoint through it.


sslavian812:
[clabot:check]

> It looks like @sslavian812 hasn't signed our **C**ontributor **L**icense **A**greement, yet.
> 
> > The purpose of a CLA is to ensure that the guardian of a project's outputs has the necessary ownership or grants of rights over all contributions to allow them to distribute under the chosen licence.
> > [Wikipedia](http://en.wikipedia.org/wiki/Contributor_License_Agreement)
> 
> You can read and sign our full Contributor License Agreement [here](http://clabot.confluent.io/cla).
> 
> Once you've signed reply with `[clabot:check]` to prove it.
> 
> Appreciation of efforts,
> 
> clabot

[clabot:check]

sslavian812:
Hi @C0urante , @rayokota ,
Would you be so kind to take a look at this PR?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -137,6 +137,22 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
             .preemptiveAuthTargetHosts(address.stream()
                 .map(addr -> HttpHost.create(addr)).collect(Collectors.toSet()));
       }
+
+      final String proxy =
+              config.getString(ElasticsearchSinkConnectorConfig.CONNECTION_PROXY_CONFIG);
+      if (proxy != null && !proxy.isEmpty()) {
```
purbon: this looks certainly good to me, might be wrap this into their own function? this method starts to be quite big, right?

sslavian812: Sure. Moved proxy configuration to a separate method.

C0urante:
@cyrusv do you want to take a look at this? IIRC this relates to work you're planning on the connector.

@sslavian812 thanks for the PR! As you can see I've roped in a colleague who may be in a better position to review; if I'm wrong about that I can take a look myself.

cyrusv:
Thank you for the contribution, @sslavian812, and the note @C0urante. I'm double-checking potential conflicts with other in-flight PRs, but don't anticipate any issue.

cyrusv:
@sslavian812, how did you test this besides the unit test? I'm specifically curious about HTTPS proxy, and how keystore/truststore/certauth get passed through. In other Kafka clients, we use the standard config properties https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/config/SslConfigs.java

I want to be sure we have confidence that `HttpClientConfig.Builder` respects the certificates we are passing in. It would be better if we used the standard Apache Kafka config method than an undocumented, untested method to pass these properties into the builder.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -154,6 +157,24 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
...
+                              HttpClientConfig.Builder builder) {
+    final String proxy =
+            config.getString(ElasticsearchSinkConnectorConfig.CONNECTION_PROXY_CONFIG);
+    if (proxy != null && !proxy.isEmpty()) {
+      try {
+        String protocol = proxy.split("://")[0];
+        String host = proxy.split("://")[1].split(":")[0];
+        int port = Integer.parseInt(proxy.split(":")[2]);
+
+        builder.proxy(new HttpHost(host, port, protocol));
```
purbon: quick question, why are you using as well the authentication system for the proxy? some proxies I've been using require you some sort of it as well. what do you think?

sslavian812:
@cyrusv 
We've been using this build in productoin for a couple of weeks already. 
We have a kubernetes cluster, and ElasticSearch can only be accessible through `http` proxy from it.

I didn't have a chance to test it with `https` proxies, but since the underlying client supports it, it should be ok.
`Http` proxy supports comes in handy, though. 

If you want me to deal with certificates in this PD as well, I'll need some help here.
Can change this PR to be "Support for `http` proxy only" and move `https` support to another.
What do you think?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
cyrusv: @purbon, not sure what you mean?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -154,6 +157,24 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
     }
   }
 
+  private void configureProxy(ElasticsearchSinkConnectorConfig config,
+                              HttpClientConfig.Builder builder) {
+    final String proxy =
+            config.getString(ElasticsearchSinkConnectorConfig.CONNECTION_PROXY_CONFIG);
+    if (proxy != null && !proxy.isEmpty()) {
+      try {
+        String protocol = proxy.split("://")[0];
```
cyrusv: Let's use the `URL.getProtocol`, `.getHost`, `getPort`. https://docs.oracle.com/javase/7/docs/api/java/net/URL.html

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
kkonstantine: I think @purbon refers to the `user:pass` part that is common to proxies. Can be supplied within the proxy url, but as I mention in my example below, I think it's better if it's supported as a separate config property like we do in the S3 connector.

levzem:
closed in lieu of #373

PR #278, opened 2019-02-12
Add HTTPS support to Elasticsearch connector

cyrusv:
Builds on https://github.com/confluentinc/kafka-connect-elasticsearch/pull/270

Uses common Apache Kafka configs

Has a very basic unit test, but integration test framework here is for Elastic 2.x, which needs to be updated. The automated test here is taking a lot of time, and I'm trying to learn if integration test in this repo is better or system test external to this repo.

This has been manually tested. Automated integration or system tests will come, but Jest client does not integrate with elastic's provided integration test framework for elastic 6.x (required for xpack), so that would require an overhaul on the client. Could be done but is out of scope

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -150,9 +150,15 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
 
   public static final String CONNECTION_SSL_CONFIG = "elastic.https";
   private static final String CONNECTION_SSL_DOC = "Override to specify whether to use SSL "
-      + "connection despite URL protocol. By default, the connector will use SSL when at "
-      + "least one URL begins with 'https:'.";
+      + "connection in all cases. By default, the connector will use SSL when at "
+      + "least one URL begins with 'https:'. This config allows users to specify SSL even if "
+      + "host is specified without a protocol.";
   public static final String CONNECTION_SSL_DEFAULT = "false";
+
+  public static final String CLIENT_AUTH_REQ_CONF = "elastic.https.client.auth.required";
+  private static final String CLIENT_AUTH_REQ_DOC = "Specify whether the client (connector) will "
+      + "be required to authenticate itself for SSL communication.";
+  public static final String CLIENT_AUTH_REQ_DEFAULT = "false";
 
   protected static ConfigDef baseConfigDef() {
     final ConfigDef configDef = new ConfigDef();
@@ -299,6 +305,17 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         ++order,
         Width.LONG,
         "Use Secure Connection"
+    )
+    .define(
+        CLIENT_AUTH_REQ_CONF,
+        Type.BOOLEAN,
+        CLIENT_AUTH_REQ_DEFAULT,
+        Importance.HIGH,
+        CLIENT_AUTH_REQ_DOC,
+        group,
+        ++order,
+        Width.SHORT,
+        "Client authentication Required"
     );
   }
 
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -34,7 +34,8 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   private static final String CONNECTION_URL_DOC =
       "List of Elasticsearch HTTP connection URLs e.g. ``http://eshost1:9200,"
       + "http://eshost2:9200``, or ``https://eshost3:9200``. If any of the URLs specifies "
-      + "``https:`` protocol, https will be used for all connections.";
+      + "``https:`` protocol, https will be used for all connections. A URL with no protocol will "
+      + "be treated as ``http``.";
   public static final String CONNECTION_USERNAME_CONFIG = "connection.username";
   private static final String CONNECTION_USERNAME_DOC =
       "The username used to authenticate with Elasticsearch. "
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -21,6 +21,7 @@ import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
 import org.apache.kafka.common.config.internals.BrokerSecurityConfigs;
 
+import java.util.List;
 import java.util.Map;
 
 import static io.confluent.connect.elasticsearch.DataConverter.BehaviorOnNullValues;
@@ -32,7 +33,8 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   public static final String CONNECTION_URL_CONFIG = "connection.url";
   private static final String CONNECTION_URL_DOC =
       "List of Elasticsearch HTTP connection URLs e.g. ``http://eshost1:9200,"
-      + "http://eshost2:9200``.";
+      + "http://eshost2:9200``, or ``https://eshost3:9200``. If any of the URLs specifies "
+      + "\"https:\" protocol, https will be used for all connections.";
   public static final String CONNECTION_USERNAME_CONFIG = "connection.username";
   private static final String CONNECTION_USERNAME_DOC =
       "The username used to authenticate with Elasticsearch. "
@@ -150,11 +152,6 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
       + "'ignore', 'warn', and 'fail'.";
 
   public static final String CONNECTION_SSL_CONFIG = "elastic.https";
-  private static final String CONNECTION_SSL_DOC = "Override to specify whether to use SSL "
-      + "connection in all cases. By default, the connector will use SSL when at "
-      + "least one URL begins with 'https:'. This config allows users to specify SSL even if "
-      + "host is specified without a protocol.";
-  public static final String CONNECTION_SSL_DEFAULT = "false";
 
   public static final String CLIENT_AUTH_CONF = BrokerSecurityConfigs.SSL_CLIENT_AUTH_CONFIG;
   private static final String CLIENT_AUTH_DOC = BrokerSecurityConfigs.SSL_CLIENT_AUTH_DOC;
@@ -295,16 +292,6 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         ++order, 
         Width.SHORT, 
         "Read Timeout"
-    ).define(
-        CONNECTION_SSL_CONFIG,
-        Type.BOOLEAN,
-        CONNECTION_SSL_DEFAULT,
-        Importance.HIGH,
-        CONNECTION_SSL_DOC,
-        group,
-        ++order,
-        Width.LONG,
-        "Use Secure Connection"
     ).define(CLIENT_AUTH_CONF,
         Type.STRING,
         CLIENT_AUTH_DEFAULT,
@@ -315,6 +302,11 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         Width.SHORT,
         "Client Authentication Required"
     );
+  }
+
+  public boolean secured() {
+    List<String> address = getList(ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG);
+    return address.stream().anyMatch(a -> a.startsWith("https:"));
   }
 
   private static void addConversionConfigs(ConfigDef configDef) {
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -24,6 +24,7 @@ import java.util.Map;
 
 import static io.confluent.connect.elasticsearch.DataConverter.BehaviorOnNullValues;
 import static io.confluent.connect.elasticsearch.bulk.BulkProcessor.BehaviorOnMalformedDoc;
+import static org.apache.kafka.common.config.SslConfigs.addClientSslSupport;
 
 public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
 
@@ -147,10 +148,22 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
       + " mapping conflict or a field name containing illegal characters. Valid options are "
       + "'ignore', 'warn', and 'fail'.";
 
+  public static final String CONNECTION_SSL_CONFIG = "elastic.https";
+  private static final String CONNECTION_SSL_DOC = "Override to specify whether to use SSL "
+      + "connection despite URL protocol. By default, the connector will use SSL when at "
+      + "least one URL begins with 'https:'.";
+  public static final String CONNECTION_SSL_DEFAULT = "false";
+
   protected static ConfigDef baseConfigDef() {
     final ConfigDef configDef = new ConfigDef();
     addConnectorConfigs(configDef);
     addConversionConfigs(configDef);
+    ConfigDef sslConfigDef = new ConfigDef();
+    addClientSslSupport(sslConfigDef);
+    configDef.embed(
+        CONNECTION_SSL_CONFIG + ".", "HTTPS",
+        configDef.configKeys().size() + 1, sslConfigDef);
+
     return configDef;
   }
 
@@ -275,7 +288,18 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         group, 
         ++order, 
         Width.SHORT, 
-        "Read Timeout");
+        "Read Timeout"
+    ).define(
+        CONNECTION_SSL_CONFIG,
+        Type.BOOLEAN,
+        CONNECTION_SSL_DEFAULT,
+        Importance.HIGH,
+        CONNECTION_SSL_DOC,
+        group,
+        ++order,
+        Width.LONG,
+        "Use Secure Connection"
+    );
   }
 
   private static void addConversionConfigs(ConfigDef configDef) {
@@ -388,6 +412,12 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
 
   public ElasticsearchSinkConnectorConfig(Map<String, String> props) {
     super(CONFIG, props);
+  }
+
+  public Map<String, Object> sslConfigs() {
+    ConfigDef sslConfigDef = new ConfigDef();
+    addClientSslSupport(sslConfigDef);
+    return sslConfigDef.parse(originalsWithPrefix(CONNECTION_SSL_CONFIG + "."));
   }
 
   public static void main(String[] args) {
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -19,7 +19,6 @@ import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Importance;
 import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
-import org.apache.kafka.common.config.internals.BrokerSecurityConfigs;
 
 import java.util.List;
 import java.util.Map;
@@ -34,8 +33,8 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   public static final String CONNECTION_URL_CONFIG = "connection.url";
   private static final String CONNECTION_URL_DOC =
       "The comma-separated list of one or more Elasticsearch URLs, such as ``http://eshost1:9200,"
-      + "http://eshost2:9200`` or ``https://eshost3:9200``. HTTPS will be used for all connections "
-      + "if any of the URLs starts with ``https:``. A URL without a protocol will be treated as "
+      + "http://eshost2:9200`` or ``https://eshost3:9200``. HTTPS is used for all connections "
+      + "if any of the URLs starts with ``https:``. A URL without a protocol is treated as "
       + "``http``.";
   public static final String CONNECTION_USERNAME_CONFIG = "connection.username";
   private static final String CONNECTION_USERNAME_DOC =
@@ -113,7 +112,7 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
       "Whether to ignore schemas during indexing. When this is set to ``true``, the record "
       + "schema will be ignored for the purpose of registering an Elasticsearch mapping. "
       + "Elasticsearch will infer the mapping from the data (dynamic mapping needs to be enabled "
-      + "by the user).\n Note that this is a global config that applies to all topics, use ``"
+      + "by the user).\n Note that this is a global config that applies to all topics. Use ``"
       + TOPIC_SCHEMA_IGNORE_CONFIG + "`` to override as ``true`` for specific topics.";
   private static final String TOPIC_SCHEMA_IGNORE_DOC =
       "List of topics for which ``" + SCHEMA_IGNORE_CONFIG + "`` should be ``true``.";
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -125,9 +130,16 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
           ElasticsearchSinkConnectorConfig.CONNECTION_USERNAME_CONFIG);
       final Password password = config.getPassword(
           ElasticsearchSinkConnectorConfig.CONNECTION_PASSWORD_CONFIG);
+      Boolean alwaysSecured = config.getBoolean(
+          ElasticsearchSinkConnectorConfig.CONNECTION_SSL_CONFIG);
+      alwaysSecured = alwaysSecured == null ? false : alwaysSecured;
```
kkonstantine: it can't be `null` (you may confirm with the code in `AbstractConfig` and `ConfigDef`). There's no value you can give to your properties that will result in a `null` value. I think the only way to have `null` here, would be to set `null` as your default (again, I'd have to remember if that's valid), which however you don't do above.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -137,6 +149,19 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
             .preemptiveAuthTargetHosts(address.stream()
                 .map(addr -> HttpHost.create(addr)).collect(Collectors.toSet()));
       }
+
+      if (secured) {
+        log.info("Using secured connection");
+        SslFactory kafkaSslFactory = new SslFactory(Mode.CLIENT, "nonev", false);
```
kkonstantine: is there a typo in `nonev`? I only see `required` / `requested` in the factory's code. Do you mind setting this as a static variable so typos get detected more easily?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -147,10 +148,22 @@
       + " mapping conflict or a field name containing illegal characters. Valid options are "
       + "'ignore', 'warn', and 'fail'.";
 
+  public static final String CONNECTION_SSL_CONFIG = "elastic.https";
```
kkonstantine: Would you mind adding some background/context on why this config is needed and protocol in url is not enough ? (basically some motivation at the initial commit message of this PR)

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -137,6 +149,19 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
...
+        kafkaSslFactory.configure(config.sslConfigs());
+        SSLContext sslContext = kafkaSslFactory.sslContext();
+        SSLConnectionSocketFactory sslSocketFactory = new SSLConnectionSocketFactory(
+            sslContext, SSLConnectionSocketFactory.getDefaultHostnameVerifier());
+
+        builder.sslSocketFactory(sslSocketFactory); // this only affects sync calls
+        // TODO builder.httpsIOSessionStrategy(httpsIOSessionStrategy);this only affects async calls
```
kkonstantine: nit: Using //TODO: highlights the first line of the code (does it highlight like that too?)

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -137,6 +149,19 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
...
+
+      if (secured) {
+        log.info("Using secured connection");
+        SslFactory kafkaSslFactory = new SslFactory(Mode.CLIENT, "nonev", false);
+        kafkaSslFactory.configure(config.sslConfigs());
+        SSLContext sslContext = kafkaSslFactory.sslContext();
+        SSLConnectionSocketFactory sslSocketFactory = new SSLConnectionSocketFactory(
+            sslContext, SSLConnectionSocketFactory.getDefaultHostnameVerifier());
+
+        builder.sslSocketFactory(sslSocketFactory); // this only affects sync calls
```
kkonstantine: nit: comments on separate line are better to distinguish between code changes and comment changes on git history

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfigTest.java
```diff
@@ -46,4 +47,20 @@ public void testSetHttpTimeoutsConfig() {
         (Integer) 15000
     );
   }
+
+  @Test
+  public void testSslConfigs() {
```
kkonstantine: should we add a test case for the default too? (which is `false`). Do we have a case for `https` in the URL? Or both these cases are covered from existing tests?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
cyrusv: I reread the code and I agree!

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
cyrusv: The original thinking was that offering an explicit, documented option gives less mysterious reasons for misconfiguration. Another motivation is today we allow ambiguous protocols in the host (e.g. `localhost` instead of `http://localhost`). However I can be convinced that this is unnecessary and we can leave it out, and rely on the rule of "if any of the connection URLs starts with HTTPS, we will use HTTPS across the board"). 

This config does raise one ambiguity I don't consider: What if user specifies `https://localhost` and sets `elastic.https=false`? Maybe we should throw an error.

I can be convinced either way, but now i'm leaning slightly towards removing the option. Thoughts?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
cyrusv: Great point -- After discussion with @NathanNam, I made this a configurable option since there is enough demand for 2-way in various connectors.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfigTest.java
cyrusv: Will address after resolving https://github.com/confluentinc/kafka-connect-elasticsearch/pull/278#discussion_r256499119

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
rhauch: I agree that it'd be better to specify it in one place, preferably in the URL protocol. Might be good to make the `elastic.https.*` configs dependent upon the URL config, such that they only show up when the protocol is `https`. Thoughts?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -147,10 +148,28 @@
...
+  public static final String CLIENT_AUTH_REQ_DEFAULT = "false";
+
   protected static ConfigDef baseConfigDef() {
     final ConfigDef configDef = new ConfigDef();
     addConnectorConfigs(configDef);
     addConversionConfigs(configDef);
+    ConfigDef sslConfigDef = new ConfigDef();
+    addClientSslSupport(sslConfigDef);
+    configDef.embed(
+        CONNECTION_SSL_CONFIG + ".", "HTTPS",
```
rhauch: Nit: Using "HTTPS" for a group name seems less useful than something like "Security" or "Secure Connections"

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -275,7 +294,28 @@ private static void addConnectorConfigs(ConfigDef configDef) {
...
+    ).define(
+        CLIENT_AUTH_REQ_CONF,
+        Type.BOOLEAN,
+        CLIENT_AUTH_REQ_DEFAULT,
+        Importance.HIGH,
+        CLIENT_AUTH_REQ_DOC,
+        group,
+        ++order,
+        Width.SHORT,
+        "Client authentication Required"
```
rhauch: Nit: capitalization should be consistent.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfigTest.java
```diff
@@ -46,4 +47,20 @@ public void testSetHttpTimeoutsConfig() {
...
+    props.put("elastic.https.ssl.keystore.password", "opensesame");
+    props.put("elastic.https.ssl.truststore.location", "/path2");
+    props.put("elastic.https.ssl.truststore.password", "opensesame2");
+    ElasticsearchSinkConnectorConfig config = new ElasticsearchSinkConnectorConfig(props);
+    Map<String, Object> sslConfigs = config.sslConfigs();
+    Assert.assertTrue(sslConfigs.size() > 0);
+    Assert.assertEquals(
+        sslConfigs.get("ssl.keystore.password"),
+        new Password("opensesame")
+    );
```
rhauch: Should this check all of the other SSL properties as well?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -390,6 +430,12 @@ public ElasticsearchSinkConnectorConfig(Map<String, String> props) {
     super(CONFIG, props);
   }
 
+  public Map<String, Object> sslConfigs() {
+    ConfigDef sslConfigDef = new ConfigDef();
+    addClientSslSupport(sslConfigDef);
+    return sslConfigDef.parse(originalsWithPrefix(CONNECTION_SSL_CONFIG + "."));
```
rhauch: It's not clear to me why this approach is used, rather than something like:
```
    return originalsWithPrefix(CONNECTION_SSL_CONFIG + ".");
```
The ConfigDef for the class already contains these, so why do we need to create a new ConfigDef to parses the originals?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -147,10 +148,28 @@
...
+  private static final String CONNECTION_SSL_DOC = "Override to specify whether to use SSL "
+      + "connection in all cases. By default, the connector will use SSL when at "
+      + "least one URL begins with 'https:'. This config allows users to specify SSL even if "
+      + "host is specified without a protocol.";
+  public static final String CONNECTION_SSL_DEFAULT = "false";
+
+  public static final String CLIENT_AUTH_REQ_CONF = "elastic.https.client.auth.required";
+  private static final String CLIENT_AUTH_REQ_DOC = "Specify whether the client (connector) will "
+      + "be required to authenticate itself for SSL communication.";
+  public static final String CLIENT_AUTH_REQ_DEFAULT = "false";
```
rhauch: This is an example of a config that I think might be best not typed as a boolean but rather as an enumerated list of values (e.g., `required` and `requested`). With the boolean, it's not clear that setting this to false corresponds to `requested`. IIUC, what's really happening is that this controls whether to "Require client certificate authentication, or terminate the connection if no suitable client certificate is found". If this is `false`, it still will sent a client certificate if one is found that is trusted by the server's CA, right? Is that really clear?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
cyrusv: Perhaps if the UI disables the fields and gives the user some documentation on how to use it, then I could get behind hiding the fields, but I think hiding them completely by default would be confusing. I don't know enough about our GUIs to make the call here what is possible.

If we are thinking "cloud-first" then we should probably ask explicitly a boolean question ("Should use SSL?" and have everything else be subfields. In this case I think the best config-file experience is if this config completely controls ssl decision, and use a validator/recommender. I can also add a log warning around the security code if they try to use a keystore but don't enable the boolean.

I'm fairly confident that parsing URLs is not the right move here after this discussion, and we should stick to boolean only. Thoughts?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
cyrusv: Using just `originalsWithPrevis(...)`,  not parse the values into the correct type or use the default value (it pulls "originals" but we want "values"). There are two functions to pull values-with-prefix in AK, but neither is really suited for this case because they both assume that the sub-field is also defined on the main ConfigDef, which we don't want to do in this connector since it will collide with names for connecting to the rest of CP. 

I wasn't thrilled about this strategy, and I think the cleanest way to handle without adding another `withPrefix`-parsing function to AK, which we can't rely on.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
cyrusv: I made this match AK config for this now :+1: 

Originally I was shooting for consistency with our Confluent Rest Utils, which uses a boolean.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
cyrusv: Thanks for the discussion, @kkonstantine , @rhauch , if you guys agree with the decision here then I can push what should be a final version

kkonstantine: Hmm, I'm not convinced the config is absolutely necessary. Checking the protocol is pretty common approach. And in the absence of protocol a sensible default may be used. 

We already do that elsewhere: 
https://github.com/confluentinc/kafka-connect-storage-cloud/blob/master/kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ProxyConfig.java#L62-L67

I think my suggestion would be to assume `https` by default (if protocol is absent) and `http` if explicitly present. WDYT?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -113,7 +119,7 @@ public JestElasticsearchClient(Map<String, String> props) {
   }
 
   // visible for testing
-  protected JestElasticsearchClient(Map<String, String> props, JestClientFactory factory) {
+  public JestElasticsearchClient(Map<String, String> props, JestClientFactory factory) {
```
kkonstantine: I don't see how this change is used in this PR. Did I miss a change?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
NathanNam: @rhauch @kkonstantine @cyrusv I will rely on your technical expertise. One thing I'd like to make a point here is what's best for users? Can we make TLS/SSL configuration consistent across connectors? It seems Replicator and S3 have their own way to handle it. How do other connectors support TLS/SSL?

cyrusv: Replicator and MQTT proxy hav an enum option for "kafka.security.protocol" and you can pick PLAINTEXT, SSL, or SASL, and get the relevant subfields. S3 proxy does not have any subfields for passing certificates and passwords around, so not sure it's the best analogy here. If anything i would upgrade this to an enum for PLAINTEXT or SSL. not sure if sasl or other mechanisms might be used in the future. I can read more of our connectors to see if there is more to the pattern.

cyrusv: After discussion w @kkonstantine, since Elastic is only HTTP(S) it is fine to go with the simpler pattern we use in s3 and determine protocol from the URL

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
cyrusv: Residual from a different strategy -- thanks for catching

cyrusv:
@kkonstantine, @rhauch I addressed all the issues discussed so far!

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -19,18 +19,22 @@
...
+import static org.apache.kafka.common.config.SslConfigs.addClientSslSupport;
 
 public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
 
   public static final String CONNECTION_URL_CONFIG = "connection.url";
   private static final String CONNECTION_URL_DOC =
       "List of Elasticsearch HTTP connection URLs e.g. ``http://eshost1:9200,"
-      + "http://eshost2:9200``.";
+      + "http://eshost2:9200``, or ``https://eshost3:9200``. If any of the URLs specifies "
+      + "``https:`` protocol, https will be used for all connections.";
```
kkonstantine: Are we accepting urls without protocol? If so, it'd be good to mention the default here, if protocol is absent.

cyrusv: Yes they will be treated as `http` to match behavior of previous versions. The text is now updated and covered in a new unit test

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
rhauch: Ok, that's fair. Another approach would be to find the configs with the prefix, get the value for each of those configs, and add to a new map without the prefixed name. Is it really simpler, clearer and better to reparse and revalidating the original strings?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -19,18 +19,23 @@
...
 
 public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
 
   public static final String CONNECTION_URL_CONFIG = "connection.url";
   private static final String CONNECTION_URL_DOC =
       "List of Elasticsearch HTTP connection URLs e.g. ``http://eshost1:9200,"
-      + "http://eshost2:9200``.";
+      + "http://eshost2:9200``, or ``https://eshost3:9200``. If any of the URLs specifies "
+      + "``https:`` protocol, https will be used for all connections. A URL with no protocol will "
+      + "be treated as ``http``.";
```
rhauch: Nit: the original doc string wasn't terribly useful, but this could still be cleaned up a bit:
```
The comma-separated list of one or more Elasticsearch URLs, such as
``http://eshost1:9200,http://eshost2:9200`` or ``https://eshost3:9200``. 
HTTPS will be used for all connections if any of the URLs start with ``https:``. 
A URL without a protocol will be treated as ``http``.
```

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfigTest.java
```diff
@@ -46,4 +50,39 @@ public void testSetHttpTimeoutsConfig() {
...
+    props.put("elastic.https.ssl.keystore.location", "/path");
+    props.put("elastic.https.ssl.keystore.password", "opensesame");
+    props.put("elastic.https.ssl.truststore.location", "/path2");
+    props.put("elastic.https.ssl.truststore.password", "opensesame2");
+    ElasticsearchSinkConnectorConfig config = new ElasticsearchSinkConnectorConfig(props);
+    Map<String, Object> sslConfigs = config.sslConfigs();
+    Assert.assertTrue(sslConfigs.size() > 0);
+    Assert.assertEquals(
+        new Password("opensesame"),
+        sslConfigs.get("ssl.keystore.password"));
```
rhauch: Nit: this line and several others doesn't match our style of putting the closing `);` on the next line.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -147,10 +152,22 @@
...
+
   protected static ConfigDef baseConfigDef() {
     final ConfigDef configDef = new ConfigDef();
     addConnectorConfigs(configDef);
     addConversionConfigs(configDef);
+    ConfigDef sslConfigDef = new ConfigDef();
+    addClientSslSupport(sslConfigDef);
+    configDef.embed(
+        CONNECTION_SSL_CONFIG + ".", "Security",
+        configDef.configKeys().size() + 1, sslConfigDef);
```
rhauch: Nit: let's match the style used in the rest of the code, so one parameter on a line, and put the closing `);` on a separate line.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -137,6 +145,27 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
             .preemptiveAuthTargetHosts(address.stream()
                 .map(addr -> HttpHost.create(addr)).collect(Collectors.toSet()));
       }
+
+      if (secured) {
```
rhauch: How about moving the `secured` variable definition closer to where it's used, or maybe even use:
```
   if (config.secured()) {
```

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -137,6 +145,27 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
...
+        log.info("Using secured connection");
```
rhauch: Do we want to include the URLs in this log message? Seems like it'd be more useful, especially when using multiple ES connectors in the same Connect cluster.

```suggestion
        log.info("Using secured connection to {}", config.urls());
```

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -137,6 +145,27 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
...
+        SSLConnectionSocketFactory sslSocketFactory = new SSLConnectionSocketFactory(
+            sslContext, SSLConnectionSocketFactory.getDefaultHostnameVerifier());
+        builder.sslSocketFactory(sslSocketFactory);
+
+        // Async calls
+        SSLIOSessionStrategy sessionStrategy = new SSLIOSessionStrategy(
+            sslContext, SSLConnectionSocketFactory.getDefaultHostnameVerifier());
+        builder.httpsIOSessionStrategy(sessionStrategy);
+      } else {
+        log.info("Using unsecured connection");
```
rhauch: Same comment as above:

```suggestion
        log.info("Using unsecured connection to {}", config.urls());
```

cyrusv:
Re-targeting to 5.2.x

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
cyrusv: I think this line is very readable: `return sslConfigDef.parse(originalsWithPrefix(CONNECTION_SSL_CONFIG_PREFIX));`. It is perhaps redundant to parse it twice, but it should be a fast operation that only happens when connector task starts. To make this clearer, I would rather make a static property of `sslConfigDef` so it's not goofily created multiple times. Here we're relying on AK abstractions and intentions instead of prescriptively repeating work. WDYT?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -275,7 +295,22 @@ private static void addConnectorConfigs(ConfigDef configDef) {
...
+    ).define(CLIENT_AUTH_CONF,
+        Type.STRING,
+        CLIENT_AUTH_DEFAULT,
+        Importance.MEDIUM,
+        CLIENT_AUTH_DOC,
+        SSL_GROUP,
+        ++order,
+        Width.SHORT,
+        "Client Authentication Required"
+    );
```
kkonstantine: Follow up comment I forgot. I thought we wanted to add an enum recommender here. `required` is the default. Which are the other options. Probably need a validator too.

cyrusv: Actually reading this more closely, I think it should be removed entirely. The connector is always the client, so it is Elastic's responsibility to ask for 2-way SSL. This config only makes sense for brokers, IIUC. Agree? thanks for the catch here, this one was fishy

kkonstantine: Sounds good removing, since it's not used!

PR #281, opened 2019-03-06
CC-3863 Integration test for HTTPS connection on Docker

cyrusv:
Use AK integration test framework for Connect
Use docker image to spin up Elastic 6 since it comes default with xpack.
Use fabric8 plugin handle docker spin-up/teardown


pom.xml
```diff
@@ -119,10 +120,101 @@
...
+                        <phase>pre-integration-test</phase>
+                        <goals>
+                            <goal>start</goal>
+                        </goals>
+                    </execution>
+                    <execution>
+                        <id>stop</id>
+                        <phase>post-integration-test</phase>
+                        <goals>
+                            <goal>stop</goal>
```
wicknicks: should we add a "remove" step after this?

src/test/java/io/confluent/connect/elasticsearch/integration/SecurityIT.java
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  @After
+  public void close() {
+    connect.stop();
+  }
+
+  private void waitForTasks(int minTasks) throws InterruptedException, TimeoutException {
+    long startTime = System.currentTimeMillis();
+    for (; ; ) {
```
wicknicks: use TestUtils.waitForCondition instead of this for loop with sleeps. it'll make the code more readable.

src/test/java/io/confluent/connect/elasticsearch/integration/SecurityIT.java
```diff
@@ -0,0 +1,161 @@
...
+
+    factory.setHttpClientConfig(builder.build());
+    return factory.getObject();
+  }
+
+  private void verify(JestClient client) throws Throwable {
+    // Read the message out of elastic directly via Jest
+    final Get get = new Get.Builder(KAFKA_TOPIC, MESSAGE_KEY).type(TYPE_NAME).build();
+
+    for (int i=0; i<30; i++) {
```
wicknicks: make a constant out of this `30`?

pom.xml
cyrusv: The default behavior for "stop" is to "stop and remove", and an explicit "remove" is usually only used for more granular, nondefault behavior in the plugin. So when you run `mvn verify`, then check `docker ps -a`, you won't see any containers related to this run with this current default configuration

cyrusv:
Thanks for the comments, @wicknicks! I believe they are addressed. There have been some updates to get the build working on Jenkins!

src/test/java/io/confluent/connect/elasticsearch/integration/SecurityIT.java
```diff
@@ -0,0 +1,119 @@
...
+    JestClientFactory factory = new JestClientFactory();
+    factory.setHttpClientConfig(getClientConfig(props));
+    return factory.getObject();
+  }
+
+  private void verify(JestClient client) throws Throwable {
+    // Read the message out of elastic directly via Jest
+    final Get get = new Get.Builder(KAFKA_TOPIC, MESSAGE_KEY).type(TYPE_NAME).build();
+
+    for (int i=0; i<RETRY__VERIFY_SECONDS; i++) {
```
wicknicks: nit: are there two underscores here?

pom.xml
```diff
@@ -119,10 +120,105 @@
...
+                            <run>
+                                <hostname>elastichost</hostname>
+                                <ports>
+                                    <port>9200:9200</port>
+                                </ports>
+                                <wait>
+                                    <tcp>
+                                        <host>localhost</host>
+                                        <ports>
+                                            <port>9200</port>
```
wicknicks: with the hardcoded port, this test might fail if that port is not available. would it be possible to look up the port number in the test? that way, the container can pick up any available port, and the test can discover it at runtime. this also opens up opportunities for running parallel tests.

pom.xml
```diff
@@ -119,10 +120,105 @@
             <version>${lucene.version}</version>
             <scope>test</scope>
         </dependency>
+        <!--Require newer version than inherited-->
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>27.0.1-jre</version>
```
wicknicks: is this needed for tests only?

src/test/java/io/confluent/connect/elasticsearch/integration/README.md
```diff
@@ -0,0 +1,12 @@
+The Integration tests are built to run with `maven-surefire-plugin` in conjuction with 
+`docker-maven-plugin` by `io.fabric8` to set up docker containers for Elasticsearch 
+(embedded nodes are no longer supported or recommended by Elasticsearch).
+
+To run the tests in their entirety, use `mvn verify`.
```
wicknicks: wouldn't `mvn integration-test` work too?

src/test/java/io/confluent/connect/elasticsearch/integration/SecurityIT.java
```diff
@@ -0,0 +1,119 @@
...
+      try {
+        JsonObject result = client.execute(get).getJsonObject();
+        Boolean success = Boolean.parseBoolean(result.get("found").getAsString());
+        if (success) return;
+      } catch (Exception e) {
```
wicknicks: also: this loop can be replaced with a `waitForCondition` as you have done above.

src/test/java/io/confluent/connect/elasticsearch/integration/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+To run the test manually (e.g. in Intellij IDEA),
+* Set up container with `mvn pre-integration-test`.
+* Run the tests manually
+* Tear down container with `mvn post-integration-test`.
+
+See the plugin specification in `pom.xml` for more details.
```
wicknicks: do you want to describe how you created the certs here? would be good to keep the knowledge written down somewhere in case we want to regenerate them, or create new ones.

src/test/java/io/confluent/connect/elasticsearch/integration/SecurityIT.java
```diff
@@ -0,0 +1,119 @@
...
+    props.put("elastic.https.ssl.key.password", "asdfasdf");
+    props.put("elastic.https.ssl.truststore.location", "./src/test/resources/certs/truststore.jks");
+    props.put("elastic.https.ssl.truststore.password", "asdfasdf");
+    connect.configureConnector(CONNECTOR_NAME, props);
+    waitForCondition(() -> {
+      ConnectorStateInfo info = connect.connectorStatus(CONNECTOR_NAME);
+      return info != null && info.tasks() != null && info.tasks().size() == 1;
+    }, "Timed out waiting for connector task to start");
+
+    connect.kafka().produce(KAFKA_TOPIC, MESSAGE_KEY, MESSAGE_VAL);
```
wicknicks: it might be good to produce and sink a few more records (2000-5000?).

pom.xml
cyrusv: This is a good point, but I cannot figure out how to do it, and in Debezium project they also use hardcoded values. I'm not sure it's worth the effort to figure this out

src/test/java/io/confluent/connect/elasticsearch/integration/README.md
cyrusv: According to the failsafe docs, "NOTE: when running integration tests, you should invoke Maven with the (shorter to type too) `mvn verify` rather than trying to invoke the integration-test phase directly, as otherwise the post-integration-test phase will not be executed." I can add this language to the README

src/test/java/io/confluent/connect/elasticsearch/integration/README.md
cyrusv: My thinking is that this README would be for integration tests in general. I have made it easy to recreate the certs with the bash script `generate_certificates.sh` -- if you think it helps with clarity then I'll link it from the HTTPS integration test docs! If you feel strongly, then I can also reference it here

Jenkinsfile
```diff
@@ -1,5 +1,55 @@
...
+
+def defaultConfig = [
+        nodeLabel       : defaultNodeLabel(),
+        cron            : '@weekly',
+        testResultSpecs : ['junit': '**/*-reports/TEST-*.xml'],
+        slackChannel    : '#connect-eng',
+        upstreamProjects:'confluentinc/common'
+]
+
+def config = jobConfig({}, defaultConfig)
```
maxzheng: The defaultConfig dict is meant to be used inside of jenkins-common only when creating a new type, like "common java" vs "common Python" jobs. 

For job specific configs, it is better to pass it as a code like:
``` 
def config = jobConfig {
    nodeLabel = ...
}
```

I guess you intended to merge this back into jenkins-common, so maybe it's ok. Either case, it's minor.

Jenkinsfile
```diff
@@ -1,5 +1,55 @@
 #!/usr/bin/env groovy
-common {
-  slackChannel = '#connect-eng'
-  upstreamProjects = 'confluentinc/common'
+
+/*
+Copied from jenkins-common/var/common.groovy
```
maxzheng: what's the reason that you can't use `common` or `dockerfile` step? It would be good to list it out here for context and future readers. 

Maintaining this here is ok if this is a unique build workflow/steps pattern that nowhere else would use -- or rarely.

Jenkinsfile
cyrusv: This is a good point, but as you suggest I will keep it similar to the original so it's easy to merge back

Jenkinsfile
cyrusv: @maxzheng, we actually expect this to be a common pattern for dozens of connectors. I will open a parallel PR into jenkins-common to add a new groovy function to handle this case. Just wanted to put it all here first to make sure it runs!

src/test/java/io/confluent/connect/elasticsearch/integration/README.md
```diff
@@ -0,0 +1,13 @@
+The Integration tests are built to run with `maven-surefire-plugin` in conjuction with 
+`docker-maven-plugin` by `io.fabric8` to set up docker containers for Elasticsearch 
+(embedded nodes are no longer supported or recommended by Elasticsearch).
+
+To run the tests in their entirety, use `mvn verify`. This will setup, run and tear
+down integration tests correctly (which `mvn integration-test` will not do).
```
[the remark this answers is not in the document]
cyrusv: So noted in f41bea86259eee5c833a93ff1ff810e838869915

pom.xml
```diff
@@ -119,10 +120,106 @@
...
+                <configuration>
+                    <watchInterval>500</watchInterval>
+                    <logDate>default</logDate>
+                    <verbose>true</verbose>
+                    <images>
+                        <image>
+                            <alias>elastic-integration-test</alias>
+                            <!--"name" is in local images if "build" is defined, else will look
+                            for an image in a remote repo-->
+                            <name>confluentinc/elastic-integration</name>
```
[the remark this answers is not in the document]
cyrusv: Addressed in c939f6712c014c9c1796199e289faff17f137c44. The commit message will have sample output.

src/test/java/io/confluent/connect/elasticsearch/integration/README.md
cyrusv: Added in d597892bad91d91f69101c02247abc25d8ad5e95

pom.xml
```diff
@@ -119,10 +120,106 @@
...
+            <version>${kafka.version}</version>
+        </dependency>
     </dependencies>
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>io.fabric8</groupId>
+                <artifactId>docker-maven-plugin</artifactId>
+                <version>${version.docker.maven.plugin}</version>
```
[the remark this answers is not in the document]
cyrusv: I'm not convinced that the profile strategy is the best strategy here. By default, we'd want to run integration tests against all versions, and we'd want a common jenkins command to work for all connectors -- meaning ideally `mvn verify` (or `install`) would run against both es5 and es6 images. This is how Debezium project also works: all containers are brought up by default claiming different ports, and all tests run by default. Since maven already gives control over which integration test to run, I think giving our devs the granular control over which docker images to bring up and down is probably a premature optimization that adds more complexity to an already complex pom.

WDYT?

cyrusv:
@rhauch, thanks for the feedback! I've addressed all the comments, including a pushback on an open thread wrt docker profiles.

pom.xml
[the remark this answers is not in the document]
rhauch: Okay, that sounds good. I think for JDBC we'll want to allow limiting to testing against specific DBMSes, and profiles may work there. But I agree, by default we probably want to run against all of them.

PR #287, opened 2019-03-19
Merge 275 into 5.2.x

wicknicks:
Merge changes from #275 into 5.2.x


cyrusv:
Why are there 9 commits? If we squash-and-rebase the other PRs, then it will be a simple cherry-pick of one commit on these, and much safer.

Do we have to do this because pint-merge creates conflicts?

PR #292, opened 2019-03-25
ST-1492: Use predefined kafka.scala.version

ewencp:
A predefined default version is provided in the common pom so we default to using the most recent Scala version where possible and so we don't have to build all versions in some other cases (such as system tests).


ewencp:
Oh, and I should mention that I validated that without this and a clean cache, kafka-connect-elasticsearch breaks the muckrake build; with this fix it passes cleanly.

PR #305, opened 2019-05-03
SEC-168: Exclude jackson-core from packaging.

avocader:
ES has a test dependency `jackson-core:2.8.1` which has a vulnerability described here - https://nvd.nist.gov/vuln/detail/CVE-2018-19362.
This PR excludes it from packaging.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,24 +107,12 @@
             <version>${es.version}</version>
             <scope>test</scope>
             <type>test-jar</type>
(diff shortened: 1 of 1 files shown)


avocader:
It happens that `connect-runtime` was not marked as `test` scope and it was bringing lots of extra dependencies, including `jackson-core`. Since the ES test framework dependency appears earlier in the pom - its version was used.

PR #306, opened 2019-05-03
fix: set connect-runtime to test scope

schroedermatt:
Changing `connect-runtime` to `test` scope to avoid conflicts with explicit dependencies.

Per @rhauch -
> It’s possible that the transitive dependencies of the Connect runtime may be added or conflict with other explicit dependencies, potentially bringing in the wrong versions and/or adding them unnecessarily to the generated archive.


rhauch:
Actually, this was already addressed with #305.

PR #309, opened 2019-05-07
If the index is not create by JestElasticsearchClient (or it just be …

randylouws:
If the index is not create by JestElasticsearchClient (or it just be already exist) , we will not to cache it , and call Jest client to check if it exists repeat and repeat.
It will dramatically decrease the performance , in my case it will be 3000 documents per second be wrote in the elasticsearch that not cache index , and 9000 documents per seconds if we cached the index.


randylouws:
[clabot:check]

levzem:
We have applied this fix in PR #340, so that we could rebase onto 4.1.x. Thank you @randylouws for your contribution. Closing this PR as a result.

argl:
I got thit hard by this using the confluent v5.3.1 docker images. Performance crawled to a halt because the connector was making an HTTP request on _every_ record that got added. Probably not the right place to ask this, but where and when does this land on the docker releases?

levzem:
We have a v5.4 planned and it will make it on there.

argl:
Thanks!

PR #310, opened 2019-05-18
Fix compilation errors due to Kafka changes

omkreddy:


omkreddy:
cc @cmccabe @C0urante

PR #312, opened 2019-05-21
CC-4788: Added elastic.security.protocol.

avocader:
Relying on `connection.url` is not sufficient in some case, e.g. when users have configured Worker to talk to ES, which is behind proxy, and proxy terminates SSL and uses plaintext.
This PR adds an additional configuration parameter, `elastic.security.protocol`, with two possible values: `PLAINTEXT`(default) and `SSL` which should be already known to users.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -21,11 +21,11 @@ import org.apache.kafka.common.config.ConfigDef.Importance;
 import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
 
-import java.util.List;
 import java.util.Map;
 
 import static io.confluent.connect.elasticsearch.DataConverter.BehaviorOnNullValues;
 import static io.confluent.connect.elasticsearch.bulk.BulkProcessor.BehaviorOnMalformedDoc;
+import static io.confluent.connect.elasticsearch.jest.JestElasticsearchClient.WriteMethod;
 import static org.apache.kafka.common.config.SslConfigs.addClientSslSupport;
 
 public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -324,8 +343,13 @@ private static void addConnectorConfigs(ConfigDef configDef) {
   }
 
   public boolean secured() {
-    List<String> address = getList(ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG);
-    return address.stream().anyMatch(a -> a.startsWith("https:"));
+    getString(ElasticsearchSinkConnectorConfig.ELASTICSEARCH_SECURITY_PROTOCOL_CONFIG);
```
cyrusv: Was this left here by mistake?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -174,18 +173,38 @@
       + " indices at startup. This is useful when the indices are a direct mapping "
       + " of the Kafka topics.";
 
+  private static final String ELASTICSEARCH_SECURITY_PROTOCOL_CONFIG = "elastic.security.protocol";
+  private static final String ELASTICSEARCH_SECURITY_PROTOCOL_DOC =
+      "The security protocol to use when connecting to Elasticsearch. "
+          + "Values can be `PLAINTEXT` or `SSL`.";
```
cyrusv: Since the name prefix here doesn't match the other elastic config names ("elastic.https"), can we get some additional text specifying that this will determine whether those other configs are required or ignored?

cyrusv:
Also I forgot to make sure you're targetting the right version

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
avocader: Good catch, removing.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
avocader: Thanks, added a clarification.

src/test/java/io/confluent/connect/elasticsearch/integration/SecurityIT.java
```diff
@@ -82,6 +82,7 @@ public void testSecureConnection() throws Throwable {
     // Start connector
     Map<String, String> props = getProps();
     props.put("connection.url", address);
+    props.put("elastic.security.protocol", "SSL");
```
mikkin: Recommend adding integration test to validate PLAINTEXT works as expected

Issue #295, opened 2019-03-29
_version not supported in elasticsearch 6.1 onwards


babadofar:
When using the option "key.ignore":false, a **_version** field is added to the bulk index metadata.  
This will not work in Elasticsearch versions from 6.1 onwards, I tested it in Elasticsearch,  in 6.40 and 7.0-rc1.  The field should be named **version**, without the leading underscore

https://www.elastic.co/guide/en/elasticsearch/reference/6.1/docs-bulk.html#bulk-versioning

`WorkerTask.java:219)\n\tat java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)\n\tat java.util.concurrent.FutureTask.run(FutureTask.java:266)\n\tat java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)\n\tat java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)\n\tat java.lang.Thread.run(Thread.java:748)\nCaused by: org.apache.kafka.connect.errors.ConnectException: Bulk request failed: {\"root_cause\":[{\"type\":\"illegal_argument_exception\",\"reason\":\"Action/metadata line [1] contains an unknown parameter [_version]\"}],\"type\":\"illegal_argument_exception\",\"reason\":\"Action/metadata line [1] contains an unknown parameter [_version]\"}\n\tat io.confluent.connect.elasticsearch.bulk.BulkProcessor$BulkTask.handleMalformedDoc(BulkProcessor.java:428)\n\tat io.confluent.connect.elasticsearch.bulk.BulkProcessor$BulkTask.execute(BulkProcessor.java:382)\n\tat io.confluent.connect.elasticsearch.bulk.BulkProcessor$BulkTask.call(BulkProcessor.java:343)\n\tat io.confluent.connect.elasticsearch.bulk.BulkProcessor$BulkTask.call(BulkProcessor.java:329)`

creativelikeadog:
Having the same problem, let me know if there is any workaround.

jpeyper:
What version of the connector are you running?

Just did a quick test with ES 6.7.1 and KCES 5.2.1 as this would have been a huge blocker for me and everything worked fine. 

[...]

PR #322, opened 2019-07-08
Enable support for Elasticsearch 7 and include flexible test infra (with TestContainers)

### LINKED ISSUE #314: compatibility with Elasticsearch version 7
I am not totally sure but according to my observations in error log of the kafka ES sink connector, the connector is not compatible  with Elasticsearch version 7.1. This because of the required "type.name" in the configuration of the connector. This "type.name" parameter is not compatible with the ES 7.1 the incompatibility is related to: https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html

**UPDATE** I can solve it by using "schema.ignore:true", but that gives me in some topics errors, like
```
Bulk request failed: [{"type":"mapper_parsing_exception","reason":"failed to parse field [amount] of type [float] in document with id 'project_budget_contributor+0+54677'","caused_by":{"type":"number_format_exception","reason":"For input string: \"AmIL\""}}]
```
so, it seems the the ES sink connector is sending string which is not recognized as number. A decimal like 5094.61 is sent as "B8YV" and 5094.60 is sent as "B8YU"
It looks a bit like the issue about numeric.mapping doesn't work for DECIMAL fields:
https://github.com/confluentinc/kafka-connect-jdbc/issues/563

**What happened**

**Step A:**
 Before upgrade ES 7.1 and running a an ES 6.8 cluster with 60 indices coming form confluent  ES sink  kafka connector with the following configuration:

```
connector.class=io.confluent.connect.elasticsearch.ElasticsearchSinkConnector
type.name=ainsource
transforms.timestamp.timestamp.field=timestamp
tasks.max=1
topics=contract_payments
transforms=unwrap,key,timestamp
key.ignore=true
transforms.timestamp.type=org.apache.kafka.connect.transforms.InsertField$Value
transforms.key.field=id
transforms.key.type=org.apache.kafka.connect.transforms.ExtractField$Key
transforms.unwrap.type=io.debezium.transforms.UnwrapFromEnvelope
connection.url=http://elasticsearch-master:9200

```
Created in ES 6.8 a index _template/kafkakconnect:

```
{
  "kafkaconnect" : {
    "order" : 0,
    "index_patterns" : [
      "person*",
      "organisation*",
      "mbo*",
      "project*",
      "system*",
      "user*",
      "talend",
      "travel*",
      "postal*",
      "notification*",
      "mobilised*",
      "memo",
      "expert*",
      "donor*",
      "contract*",
      "advisory*",
      "address",
      "comment*",
      "contact*",
      "emergency*",
      "file",
      "goal",
      "intake*",
      "language*",
      "light*",
      "linked*"
    ],
    "settings" : {
      "index" : {
        "number_of_shards" : "1",
        "number_of_replicas" : "1"
      }
    },
    "mappings" : {
      "ainsource" : {
        "dynamic_templates" : [
          {
            "dates" : {
              "mapping" : {
                "type" : "date"
              },
              "match" : "timestamp"
            }
          },
          {
            "created" : {
              "mapping" : {
                "type" : "date"
              },
              "match" : "created_at"
            }
          },
          {
            "modified" : {
              "mapping" : {
                "type" : "date"
              },
              "match" : "modified_on"
            }
          },
          {
            "otherdates" : {
              "mapping" : {
                "type" : "date"
              },
              "match" : "date_*"
            }
          },
          {
            "duestartenddates" : {
              "mapping" : {
                "type" : "date"
              },
              "match" : "*_date"
            }
          },
          {
            "non_analysed_string_template" : {
              "mapping" : {
                "ignore_above" : 1025,
                "type" : "keyword"
              },
              "match_mapping_type" : "string",
              "match" : "*"
            }
          }
        ]
      }
    },
    "aliases" : { }
  }
}
```
Everything worked fine.

**Step B:**
Followed the instruction in ES 6.8 to check if it's possible to upgrade. After this I did the upgrade and everything seems to work. Because the"old"  indices still had the type.name "ainsource" .

**Step C:**
 When new values where steaming in from ES sink connector with above described configuration every topic with new values stopped working with the same error:

```
org.apache.kafka.connect.errors.ConnectException: Exiting WorkerSinkTask due to unrecoverable exception.
	at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:587)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:323)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:226)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:194)
	at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:175)
	at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:219)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: org.apache.kafka.connect.errors.ConnectException: Cannot create mapping {"ainsource":{"properties":{"id":{"type":"integer"},"contract_id":{"type":"integer"},"title":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}},"amount":{"type":"double"},"status":{"type":"integer"},"date_planned":{"type":"long"},"date_receipt":{"type":"long"},"date_request":{"type":"long"},"modified_by":{"type":"integer"},"modified_on":{"type":"long"},"code":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}},"date_paid":{"type":"long"},"amount_paid":{"type":"double"},"is_payment_imported":{"type":"short"},"timestamp":{"type":"date"}}}} -- {"root_cause":[{"type":"illegal_argument_exception","reason":"Types cannot be provided in put mapping requests, unless the include_type_name parameter is set to true."}],"type":"illegal_argument_exception","reason":"Types cannot be provided in put mapping requests, unless the include_type_name parameter is set to true."}
	at io.confluent.connect.elasticsearch.jest.JestElasticsearchClient.createMapping(JestElasticsearchClient.java:255)
	at io.confluent.connect.elasticsearch.Mapping.createMapping(Mapping.java:67)
	at io.confluent.connect.elasticsearch.ElasticsearchWriter.write(ElasticsearchWriter.java:260)
	at io.confluent.connect.elasticsearch.ElasticsearchSinkTask.put(ElasticsearchSinkTask.java:163)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:565)
	... 10 more

```

Question 1: in case this not a bug, can I assume the current confluent ES Sink connector is not compatible with ES 7.1 and I need to go back to ES 6.8?
Question 2: Are there any plans to make the connector compatible with ES 7.1?

BTW: The people from ES are recommending to do the following:
"In order to not be broken by this change, you should make your application pass include_type_name=true to these API calls, which is a no-op on 6.x, and will tell 7.x to make these APIs behave as they used to do in 6.x. For instance here is an index creation call with this parameter. It works the same way on 6.8 and 7.x."
So, perhaps it would be possible to build put URL with something like
`?include_type_name=true`

thx!

purbon:
This PR add changes necessary to enable Elasticsearch support as well as add a flexible test infra with test containers to allow test with each major, or minor, release of Elasticsearch in a flexible way.

## Changes introduced.

* Add portable actions to manipulate indexes and mapping that if the version detected is 7.x will include the parameter _include_type_name_ to _true_. As referenced from https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html this change will allow a backwards compatible way of operation until the usage of types and Jest are upgraded for version 8.x.

* Upgrade the Jest lib to the version 6.3.1 as this version is able to avoid problems with the __version_ as described in #295. 

* Remove the usage of _ESIntegTestCase_, in favor of [TestContainers](https://www.testcontainers.org/modules/elasticsearch/), as a base class of _ElasticsearchSinkTestBase_, following best practises recommended by the same elasticsearch developers of avoiding using embedded code for REST kind _integration_ tests.

## Notes

With the changes introduced in this PR, you need docker to execute all test. When you run the _mvn test_ command, [TestContainers](https://www.testcontainers.org/modules/elasticsearch/) will make sure to provision the necessary container for the test to function.

Because of the changes in the test, I bet this will not work in the Jenkins test system. 

# Pending work

- [x] Enable specific Elasticsearch version selection, via a system or event variable, when running test. Like this we could build a test matrix with all supported version of Elasticsearch.
- [ ] If test changes are accepted, need to work on integrating them with Jenkins.
- [x] Cleanup unused or deprecated version dependencies.

Fixed #314 #295


purbon:
Actually test are failing, according to Jenkins because of:

```
[INFO] --- docker-maven-plugin:0.28.0:start (start) @ kafka-connect-elasticsearch ---
[INFO] DOCKER> [confluentinc/elastic-integration-snapshot-190708-102858-0864:latest] "elastic-integration-test": Start container 7c6d6b908392
[ERROR] DOCKER> Error occurred during container startup, shutting down...
[ERROR] DOCKER> I/O Error [Unable to start container id [7c6d6b908392] : {"message":"driver failed programming external connectivity on endpoint elastic-integration-snapshot-190708-102858-0864-1 (e626c8dc152714c9fa5933758714490de6788daab6438fc6728787a55c400a72): Bind for 0.0.0.0:9200 failed: port is already allocated"} (Internal Server Error: 500)]
```

any thoughts on that?

babadofar:
This PR is very welcome :) 
Are you considering to get someone from @elastic in as reviewers, to help out with implementing best practices?

purbon:
Hi @babadofar , thanks for your nice comments. Lots of elasticsearch experience under my belt as well , even former elasticsearch employee here.

src/test/java/io/confluent/connect/elasticsearch/TestContainerTest.java
```diff
@@ -0,0 +1,218 @@
...
+  public void tearDown() {
+    if (client != null) {
+      client.close();
+    }
+    client = null;
+    container.stop();
+  }
+
+
+  protected ElasticsearchClient client;
```
kkonstantine: ideally all the member fields would be declared at the beginning of the class.

src/test/java/io/confluent/connect/elasticsearch/TestContainerTest.java
```diff
@@ -0,0 +1,218 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.testcontainers.elasticsearch.ElasticsearchContainer;
+
+public class TestContainerTest {
+
+  private static final String INDEX = "kafka-connect";
+  private static final String TYPE = "kafka-connect-type";
+
+
```
kkonstantine: nit: extra space

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTestBase.java
```diff
@@ -50,30 +48,41 @@
...
   protected static final TopicPartition TOPIC_PARTITION3 = new TopicPartition(TOPIC, PARTITION3);
 
+  protected static ElasticsearchContainer container;
   protected ElasticsearchClient client;
   private DataConverter converter;
 
+  @BeforeClass
+  public static void setupBeforeAll() {
+    // Relevant and available docker images for elastic can be found at https://www.docker.elastic.co
+    container = new ElasticsearchContainer("docker.elastic.co/elasticsearch/elasticsearch:7.0.0");
```
rhauch: These tests are currently unit tests, whereas this change would really warrant making them "integration" tests that require additional time to load the container. We normally run all of our integration tests during the Maven integration test phase, name them accordingly, and put them in the `integration` package. Should we make these test class integration tests instead?

My concern with doing this is that a lot of our unit tests are going to switch to be integration tests, and that our unit test coverage becomes pretty spartan.

An alternative approach is to *add* integration tests that effectively do what these tests do (and preferably more), yet keep these simple unit tests for faster compile-build-test cycles during development. Ideally we'd be able to run integration tests against several versions.

src/test/java/io/confluent/connect/elasticsearch/TestContainerTest.java
```diff
@@ -0,0 +1,218 @@
...
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.SchemaBuilder;
+import org.apache.kafka.connect.data.Time;
+import org.apache.kafka.connect.data.Timestamp;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.testcontainers.elasticsearch.ElasticsearchContainer;
+
+public class TestContainerTest {
```
rhauch: The name is initially a bit confusing. Maybe `VerifyContainerTest`, or something similar?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -57,16 +71,6 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.IOException;
```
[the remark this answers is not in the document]
purbon: auch, good catch. Sorry for this. Intellij was trying to me smarter than me :-)

src/main/java/io/confluent/connect/elasticsearch/jest/actions/PortableJestCreateIndex.java
```diff
@@ -0,0 +1,41 @@
...
+package io.confluent.connect.elasticsearch.jest.actions;
+
+import io.confluent.connect.elasticsearch.ElasticsearchClient.Version;
+import io.searchbox.indices.CreateIndex;
+
+/**
+ * Portable Jest action to create indexes, this action will support ES version 7 with type
+ * support still enabled.
+ */
+public class PortableJestCreateIndex extends CreateIndex.Builder {
```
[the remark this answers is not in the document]
purbon: fair point, thanks @kkonstantine

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -91,6 +91,26 @@
    */
   JsonObject search(String query, String index, String type) throws IOException;
 
+  /**
+   * Forces a flush of all indexes to disk
+   *
+   * @throws IOException If the client cannot execute the request
+   */
+  void flush() throws IOException;
```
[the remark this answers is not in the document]
purbon: I added it, even if is not used. For test, could be very useful in the future. Flush is a hard core refresh ;-). happy to get rid of it if you like.

src/test/java/io/confluent/connect/elasticsearch/TestContainerTest.java
purbon: this is a legacy file,  it should not be there is only experiments of mine ;-)

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -91,6 +91,26 @@
...
+
+  /**
+   * Delete all indexes in Elasticsearch (useful for test)
+   * @throws IOException if the client cannot execute the request
+   */
+  void deleteAll() throws IOException;
```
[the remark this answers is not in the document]
purbon: Sort of makes sense, but I'm confused in one thing. Can you give me an example of _client classes that might have customized the connector_ ? as of now I know envision two possible clients:

* The jest client, this is the only one we've right now.
* The upcoming client using the elastic java REST client.

Or you envision an specific client only for tests? there this methods makes more sense than in the "regular" client. I'm probably more on this, should we:

* Have an specific client for test ? 
* If yes, we should implement this methods like deleteAll, flush, etc...
* For the main clients, this operations are not that useful.

What do you think @kkonstantine ?

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTaskTest.java
```diff
@@ -62,8 +58,6 @@
 
   @Test
   public void testPutAndFlush() throws Exception {
-    InternalTestCluster cluster = ESIntegTestCase.internalCluster();
```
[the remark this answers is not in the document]
purbon: This ties with the discussion points started in @rhauch comment. This class is more an integration test than a unit test, and as well we where using, more abusing, the embedded client for test. Its is a better practise to spin a "real" Elasticsearch and run the test as is done now with test containers.

In the current shape, we only test with one Elasticsearch version, while introducing the changes proposed in this PR, we will be able to test with multiple versions, current and future. This will provide us with a better test bed, without losing anything.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTestBase.java
```diff
@@ -50,30 +48,41 @@
...
+  @AfterClass
+  public static void teardownAfterAll() {
+    container.close();
+  }
+
   @Before
   public void setUp() throws Exception {
-    super.setUp();
-    client = new JestElasticsearchClient("http://localhost:" + getPort());
+    client = new JestElasticsearchClient("http://"+container.getHttpHostAddress());
```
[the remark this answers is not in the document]
purbon: this is used in the test,, otherwise the connection won't be possible to Elasticsearch.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTestBase.java
```diff
@@ -50,30 +48,41 @@
...
     converter = new DataConverter(true, BehaviorOnNullValues.IGNORE);
   }
 
   @After
-  public void tearDown() throws Exception {
-    super.tearDown();
+  public void tearDown() {
+   try {
+     client.deleteAll();
+   } catch (Exception ex) {
```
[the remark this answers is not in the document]
purbon: makes complete sense. thanks!

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTestBase.java
purbon: Hi @rhauch this is actually a very valid question and I was expecting you to drop it ;-). I see this question in two sides, I will try to argument this here, hopefully in a clear way.

## Current state
As of now the connector test such as the ones inheriting from ElasticsearchSinkTestBase are having a direct interaction with an embedded Elasticsearch version. As you can see from the same class definition:

```java
public class ElasticsearchSinkTestBase extends ESIntegTestCase {
```

As of now this means, MappingTest, ElasticsearchSinkTaskTest and ElasticsearchWriterTest.

Taking as example _MappingTest_ for example we can see a mixture of test, for example 

```java
 public void testMapping() throws Exception {
    InternalTestCluster cluster = ESIntegTestCase.internalCluster();
    cluster.ensureAtLeastNumDataNodes(1);
```
is testing the creation and retrieval of a mapping with an embedded real Elasticsearch. This function is clearly an integration test as it will depend on the target system API and this is subject to changes across major versions.

we've as well
```java
  @Test
  @SuppressWarnings("unchecked")
  public void testStringMappingForES6() throws Exception {
    ElasticsearchClient client = mock(ElasticsearchClient.class);
    when(client.getVersion()).thenReturn(ElasticsearchClient.Version.ES_V6);

    Schema schema = SchemaBuilder.struct().name("textRecord")
            .field("string", Schema.STRING_SCHEMA)
            .build();
    ObjectNode mapping = (ObjectNode) Mapping.inferMapping(client, schema);
```

that is more a unit test per definition.

the problem with the first one is that makes part of our test dependant of ES internal versions. When code is develop like this, as is a common practise in the Elasticsearch world (aka no dependency management), plugins are released for specific versions. This is clearly not what we are aiming to do or need to do.

Other test classes mention here have a similar pattern. 

## How I envision things.

As we can't have a connector release for each Elasticsearch version, we would need to offer a more clear way to test between major versions. For this my proposal is to:

* Move as integration test, under integration test the ones that makes sense. For example test for creating and retrieving mappings as I commented earlier. For the tests move into integration test, because they are end of the day integration test, we should introduce mocks to abstract the end system call.
* Have a flexible integration test layer were we can pass a variable and pull different old and upcoming Elasticsearch versions. This will allow us to build a more resilient test matrix and detect support for upcoming or compatibility problems easier.

Does this makes sense? what do you think @rhauch and @kkonstantine ?

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTestBase.java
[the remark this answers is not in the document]
kkonstantine: I was referring to the lack of single spaces between `+`. In regular code this is not permitted by the checkstyle. 
Should be: 
`    client = new JestElasticsearchClient("http://" + container.getHttpHostAddress());`

purbon:
Sorry, I can't really answer your message from https://github.com/confluentinc/kafka-connect-elasticsearch/pull/322#discussion_r302991068

With this in mind, I do agree I like your proposal. I'll move forward with it.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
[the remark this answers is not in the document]
purbon: fair for me! 😄

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTestBase.java
purbon: Hi @kkonstantine, don't get me wrong, I'm not asking for having less unit test and more unit test. I'm actually advocating for better unit testing than what we have now.

End of the day what we've now is a mixture of integration test, "shaded" as unit test, this is because end of the day they go to a real elasticsearch, even if embedded. The way I've got to learn and apply unit test in my experience as developer, they never had interaction with an external system, this sort of defuse the idea of a unit test.

But I'm for sure open to change my mind, talking about the options presented in other messages, sorry if I'm missing any:

+ Mock out the interaction with the Elasticsearch server: This is my preferred option as it will enable proper entity test without interference of external apis. The integration with the external apis will be tested with the integration tests.
+ Work with an embed of a given version (probably latest): This will certainly safe a lot of need of mocks, but still all unit test code will be dependant on the last available version used of elasticsearch. I'm not confortable with this, but might be a solution to avoid having a ton of mocks.
+ Provide a parametrised version for an embed test:  I would certainly not enforce this, i don't even know how to start with it as the multiple versions and provided test. If possible I would discard this option.

What do you think @kkonstantine ?

purbon: Hi @kkonstantine, I created https://github.com/confluentinc/kafka-connect-elasticsearch/pull/323 this weekend, hopefully it able to match if second point is more of your preference.

If this is the direction you would recommend this PR to take, I would kindly close this PR and move all the work there. 

I gave a bit of thought and I think moving forward implementing option 2 of the previous list is the most sensible way now, but certainly i would recommend improving the way unit tests are approached because the current approach is not providing enough level of confidence. Hopefully last is corrected with better real integration test, so end of the day we could remove the need of the real ES instance for the unit test. Makes sense?

purbon: Talking about real examples of integration test that lay inside the current unit test suit, what do you think?

```java
  @Test
  public void testPutAndFlush() throws Exception {
    Map<String, String> props = createProps();

    ElasticsearchSinkTask task = new ElasticsearchSinkTask();
    task.start(props, client);
    task.open(new HashSet<>(Arrays.asList(TOPIC_PARTITION, TOPIC_PARTITION2, TOPIC_PARTITION3)));

    String key = "key";
    Schema schema = createSchema();
    Struct record = createRecord(schema);

    Collection<SinkRecord> records = new ArrayList<>();
    SinkRecord sinkRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, key, schema, record, 0);
    records.add(sinkRecord);

    sinkRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, key, schema, record, 1);
    records.add(sinkRecord);

    task.put(records);
    task.flush(null);

    client.refresh();

    verifySearchResults(records, true, false);
  }
```

the fact that we use _verifySearchResults_ it forces an interaction with the external system, at less based on my experience, a more suited unit test style would be, something on the sense of:

```java
```java
  @Test
  public void testPutAndFlush() throws Exception {
    Map<String, String> props = createProps();

    ElasticsearchSinkTask task = new ElasticsearchSinkTask();
    task.start(props, client);
    task.open(new HashSet<>(Arrays.asList(TOPIC_PARTITION, TOPIC_PARTITION2, TOPIC_PARTITION3)));

    String key = "key";
    Schema schema = createSchema();
    Struct record = createRecord(schema);

    Collection<SinkRecord> records = new ArrayList<>();
    SinkRecord sinkRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, key, schema, record, 0);
    records.add(sinkRecord);

    sinkRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, key, schema, record, 1);
    records.add(sinkRecord);

    task.put(records);
    task.flush(null);

   // in mockito or other similar framework call
  // verify executeBulk is called the exact number of times, that for this test should be 1
 
  }
```

this would be the approach I propose to take for the first point in the previous list. While the current approach "solve the problem", it makes the test a bit harder to trust as they depend on a given elasticsearch version. 

What do you think?

cyrusv:
Thanks for the hard work, @purbon! I'm wondering if you have a recommendation on which versions of ES we should go ahead and deprecate, so we can feel comfortable removing them in future connector releases?

purbon:
This is a hard question to answer, there is always going to be teams that for one reason or another they will use a very old version. But if we stick to the current Elasticsearch EOL matrix,I would recommend we drop as of upcoming releases support for anything below ES 6.0.0, if you want to extend it a bit, probably keep support until latest release of 5.6.x, but certainly I would recommend dropping everything below 5.6.x as they are really ancient history.

more info: https://www.elastic.co/support/eol

kkonstantine:
Hmm rushed it a bit. 
@purbon I see the tests fail with port collisions. We'd need to fix this before we merge.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTestBase.java
rhauch: IMO, we should not get too hung up on whether "unit" tests never hit an external system and that "integration" tests always hit an external system. IIUC, the embedded version of Elasticsearch was intended primarily for testing purposes so it becomes easier to test individual components and logic, and in that sense I don't think of the embedded Elasticsearch as a "real" external system, and is closer to a "mock" system. In fact, I think we should avoid creating our own "mock" system because isn't worth the effort or risk. Let's not be pedantic about what is a "unit" vs "integration" tests. I'd much rather be pragmatic.

I do think it's acceptable to have "unit" tests that use a single embedded version of Elasticsearch. Granted, we should probably use a more recent version than what we've been using, but the purpose of these unit tests is to verify basic function and operation _of individual components and logic_, not of the entire connector. The point is that they are quick and focused on testing *individual components*. If it happens to be easiest to verify this using a real embedded Elasticsearch instance, then so be it.

Those "unit" tests would not really cover compatibility with any versions other than the one selected version, and this is where the ITs come in with two main purposes:
a) run the whole connector via Connect distributed worker(s) to validate the lifecycle and verify proper functionality
b) do this against a series of Elasticsearch versions (e.g., the latest patch on multiple minor versions)

Docker (or Test Containers) may be the most effective way to satisfy b) while running *real* instances of Elasticsearch. One way to do this with Docker and the Fabric8 plugin is to have different profiles for different versions of Elasticsearch, such that I could use `-Pv7.2.0` (for example) run the same ITs against the 7.2.0 version (other versions could have similar profiles). ([Elastic's Docker registry](https://www.docker.elastic.co/) contains nearly every version, unlike Docker Hub's.) Locally I could then run `-Pv7.2.0,v6.8.1` to sequentially run the ITs against multiple versions in one build. Once we get ITs runnable on Jenkins, we could change the `-Pjenkins` profile to run multiple of these profiles this way.

Test Containers gives us a bit more flexibility without all the profiles, since we could create a separate test suite (classes) for each version we wish to test. A base class would contain all of the real tests, but the concrete subclasses would specify the version of ES that would be used for those tests.

rhauch: BTW, I think this also is aligned with @kkonstantine's previous second bullet, which @purbon also seemed to favor. I think this also allows us to do this in steps: first get ES 7 working with the connector, then (in a separate PR) add ways of running ITs against multiple ES versions for compatibility testing. We could improve the unit testing (as you suggested above, @purbon) to rely less upon an embedded ES server in yet another subsequent PR, though as I stated above, IMO that is lower priority.

purbon:
Hi @kkonstantine and @rhauch, as spoken with @rhauch, we agreed to close this PR in favour of #323. 

PR #323 has:

* Keep the current unit test as they are right now.
* Add the necessary new Integration test (using TestContainer) to validate the changes on this PR.
* This PR will not change current integration test framework, we can always improve this in the future.

In next iterations, I would like to improve current unit test, by relying less and less on the old _ESIntegrationTestCase_ framework. We could as well decide how to integrate either more docker-maven, fabric or TestContainers, step by step. Basically following the discussion on https://github.com/confluentinc/kafka-connect-elasticsearch/pull/322#discussion_r302189204

PR #324, opened 2019-07-18
CC-5605: Log version conflicts as warnings.

avocader:
This PR handles version conflicts for upsert use-case when a topic has multiple partitions and there is no guarantee that messages with the same ids will be written to the same partition.

We are assuming that when we use a user-provided key(`ignoreKey=false`), all messages with the same id are produced to the same partition, thus we use message's offset as version when writing a document to ES.

Example:
User has a topic `moon` with 10 partitions, and produces a message `{"entityId": "42": "color": "red"}`, it gets written to partition 1, with offset 50.
💚 ES Connector sends a document to ES: `Key:42, Version:50, Payload:{"color": "red"}`.
Then, the next message with the same id is produced `{"entityId": "42": "color": "red"}`, it gets written to partition 2, with offset 30.
🔴 ES Connector sends a document to ES: `Key:42, Version:30, Payload:{"color": "red"}`, it gets rejected with DEBUG `Ignoring version conflicts for items`.

Solution:
Use `timestamp` from a message, if it's present, failback to offset.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
@@ -379,7 +379,7 @@ public class JestElasticsearchClient implements ElasticsearchClient {
     }
 
     if (!versionConflicts.isEmpty()) {
-      LOG.debug("Ignoring version conflicts for items: {}", versionConflicts);
+      LOG.warn("Ignoring version conflicts for items: {}", versionConflicts);
       if (errors.isEmpty()) {
         // The only errors were version conflicts
         return BulkResponse.success();


src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
```diff
@@ -332,10 +334,54 @@ public void failOnNullValue() {
     }
   }
 
+  @Test
+  public void useOffsetAsVersion() {
+    converter = new DataConverter(true, BehaviorOnNullValues.FAIL);
+    Struct struct = new Struct(schema);
+    struct.put("string", "abc");
+    Long ts = System.currentTimeMillis();
```
Cyril-Engels: This is not used.

src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
```diff
@@ -332,10 +334,54 @@ public void failOnNullValue() {
...
+    SinkRecord sinkRecord = createSinkRecordWithValueAndLogAppendTs(struct, ts);
+    IndexableRecord indexableRecord = converter.convertRecord(sinkRecord, index, type, false, false);
+    assertNotNull(indexableRecord);
+    assertEquals(ts, indexableRecord.version);
+  }
+
```
Cyril-Engels: Maybe add one more test with `ignoreKey` true?

wicknicks:
@avocader should this be backported to earlier versions?

cyrusv:
I would recommend leaving the code as-is, and upgrading the `degug` to a `warn`, and elaborate in the warning that this can happen only if your kafka records are not keyed correctly (e.g. "entity_id") in your example. It can guide the user to either re-key their records to guarantee ordering in a partition, or they can configure their connector to ignore keys. But I believe switching to timestamp just changes the frequency of the error, and does not resolve their issue, or give any new information on how to resolve for themselves.

PR #327, opened 2019-07-30
Enable Elasticsearch 7 support with adding minimum  Integration test (based on TestContainer)

### LINKED ISSUE #314: compatibility with Elasticsearch version 7

purbon:
This PR add changes necessary to enable Elasticsearch support as well as add a flexible test infra with test containers to allow test with each major, or minor, release of Elasticsearch in a flexible way.

## Changes introduced.

* Add portable actions to manipulate indexes and mapping that if the version detected is 7.x will include the parameter _include_type_name_ to _true_. As referenced from https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html this change will allow a backwards compatible way of operation until the usage of types and Jest are upgraded for version 8.x.

* Upgrade the Jest lib to the version 6.3.1 as this version is able to avoid problems with the __version_ as described in #295. 

This PR contains all major commits and feedback provided during the discussion of #322, including an initial implementation fo the integration test direction, were current unit test methodology (by using an embedded Elasticsearch node is keep) and a new integration test layer is introduced by using test containers.

Out of the scope of this PR is to change the usage of maven-docker-plugin, but this should be considered in the future as using TestContainers is certainly a more flexible way of running integration test.

## Notes

To run the integration test that verify support for ES 7 you need to run something like:

```bash
mvn integration-test -Delasticsearch.version=7.0.0 -Pjenkins
```

by default elasticsearch.version=7.0.0

Fixed #314 #295


purbon:
@kkonstantine @rhauch PR is rebased and cleanup as discussed in #322 , looking forward to get your latest thoughts and hopefully green light to move forward.

Looking forward to it.

PD: Before is good to merge, let me know and I will squash to 1 commit, this will provide a more clean history.

pom.xml
```diff
@@ -121,6 +121,12 @@
             <version>${lucene.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>elasticsearch</artifactId>
+            <version>1.11.3</version>
```
rhauch: BTW, 1.11.4 is the latest patch release of TestContainers for Java.

src/main/java/io/confluent/connect/elasticsearch/jest/actions/PortableJestCreateIndexBuilder.java
```diff
@@ -0,0 +1,45 @@
...
+
+  public PortableJestCreateIndexBuilder(String index, Version version) {
+    super(index);
+    this.version = version;
+  }
+
+  @Override
+  public CreateIndex build() {
+    if (version.equals(Version.ES_V7)) {
+      setParameter(INCLUDE_TYPE_NAME_PARAM, true);
```
rhauch: This seems brittle if/when we add `Version.ES_V8`. Should we instead do something like:
```
if (version.compareTo(Version.ES_V7) >= 0) {
```
This is annoyingly vague and easy to mess up the logic, so we could simplify it to:
```
if (version.isAtLeast(Version.ES_V7)) {
```
if we added a method to the enum:
```
public boolean isAtLeast(Version other) {
  return this.compareTo(other) >= 0;
}
```
WDYT?

src/main/java/io/confluent/connect/elasticsearch/jest/actions/PortableJestGetMappingBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+
+  private final Version version;
+
+  public PortableJestGetMappingBuilder(Version version) {
+    this.version = version;
+  }
+
+  @Override
+  public GetMapping build() {
+    if (version.equals(Version.ES_V7)) {
```
rhauch: Same comment as above.

src/main/java/io/confluent/connect/elasticsearch/jest/actions/PortableJestPutMappingBuilder.java
```diff
@@ -0,0 +1,45 @@
...
+  private final Version version;
+
+  public PortableJestPutMappingBuilder(String index, String type, Object source, Version version) {
+    super(index, type, source);
+    this.version = version;
+  }
+
+  @Override
+  public PutMapping build() {
+    if (version.equals(Version.ES_V7)) {
```
rhauch: Same comment as above.

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchIntegrationTestBase.java
```diff
@@ -0,0 +1,164 @@
...
+        IndexableRecord indexableRecord = converter.convertRecord((SinkRecord) record, index, TYPE, ignoreKey, ignoreSchema);
+        assertEquals(indexableRecord.payload, hits.get(indexableRecord.key.id));
+      } else {
+        assertEquals(record, hits.get("key"));
+      }
+    }
+  }
+
+  private static String getElasticsearchContainerVersion() {
+    return System.getProperty("esVersion", DEFAULT_ES_VERSION);
```
rhauch: I just recently created https://github.com/confluentinc/connect-testcontainers for the JMS connectors (there were no existing TestContainer classes for JMS brokers), and there I read from first the system property and then a similarly-named environment property.

I don't know whether it's worth also checking an environment property here, but I would suggest that we adopt a similar naming convention for this project for the system properties. In https://github.com/confluentinc/connect-testcontainers#overriding-the-image-version I outline the approach of using a period-separated all lowercase system property (e.g., `activemq.version` or `es.version` here) because it's easily mapped to underscore-separated uppercase environment variable names (e.g., `ACTIVEMQ_VERSION` or `ES_VERSION` here). 

Thus, my suggestion is to:
```suggestion
    return System.getProperty("es.version", DEFAULT_ES_VERSION);
```

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchIntegrationTestBase.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  private static String getElasticsearchDockerImageName() {
+    return System.getProperty("esImageName", DEFAULT_DOCKER_IMAGE_NAME);
```
rhauch: Similar reasoning to my previous comment, maybe:
```suggestion
    return System.getProperty("es.image", DEFAULT_DOCKER_IMAGE_NAME);
```
See also https://github.com/confluentinc/connect-testcontainers/blob/1.0.x/activemq/src/main/java/io/confluent/testcontainers/activemq/ActiveMQContainer.java#L31-L41

src/main/java/io/confluent/connect/elasticsearch/jest/actions/PortableJestCreateIndexBuilder.java
purbon: Hi,
  this parameter is only for ES_V7, it will not work for newer versions. The only solution to have the connector support ES_V8 will be to remove types, what means:

* Code changes to remove usage of types.
* Wait until Jest support ES_V8, or patch Jest Actions Builders to not use types.
* Another option would be to use the Elasticsearch own client.

BTW, removing types is not a backwards compatible change, so we'll need to have a backwards compatible change for the connector build-in.

_Note:_ that patch Jest Action builders to not use types will be quite some work, this is why I created this PR for now and will do the work for ES_V8 in an upcoming PR. 

What do you think?

src/main/java/io/confluent/connect/elasticsearch/jest/actions/PortableJestGetMappingBuilder.java
purbon: See https://github.com/confluentinc/kafka-connect-elasticsearch/pull/327#discussion_r309052970

src/main/java/io/confluent/connect/elasticsearch/jest/actions/PortableJestPutMappingBuilder.java
purbon: See https://github.com/confluentinc/kafka-connect-elasticsearch/pull/327#discussion_r309052970

pom.xml
purbon: got catch, this one was the latest when I started this work 1m ago 😄 .

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchIntegrationTestBase.java
```diff
@@ -0,0 +1,183 @@
...
+  protected static final TopicPartition TOPIC_PARTITION = new TopicPartition(TOPIC, PARTITION);
+  protected static final TopicPartition TOPIC_PARTITION2 = new TopicPartition(TOPIC, PARTITION2);
+  protected static final TopicPartition TOPIC_PARTITION3 = new TopicPartition(TOPIC, PARTITION3);
+
+  private static final String DEFAULT_ES_VERSION = "7.0.0";
+  private static final String DEFAULT_DOCKER_IMAGE_NAME = "docker.elastic.co/elasticsearch/elasticsearch";
+  private static final String ELASTICSEARCH_VERSION_SYS_PROP = "elasticsearch.version";
+  private static final String ELASTICSEARCH_VERSION_ENV_PROP = "ELASTICSEARCH_VERSION";
+  private static final String ELASTICSEARCH_DOCKER_IMAGE_SYS_PROP = "elasticsearch.docker.image";
+  private static final String ELASTICSEARCH_DOCKER_IMAGE_ENV_PROP = "ELASTICSEARCH_DOCKER_IMAGE";
```
rhauch: Any chance we can be consistent with the other already released library and use `elasticsearch.image` and `ELASTICESEARCH_IMAGE` instead? I think we don’t need to include “Docker” in the name since that is pretty well assumed.

purbon:
I'll add this later today, thanks for your wisdom. do you mind pre
approving so i can merge after this change?

rhauch:
I've added a change that makes the names more consistent, and updated the Confluent Hub description.

I'm also thinking that maybe we need to skip the integration tests on Jenkins. We currently have a limitation our Jenkins environment where any build that uses Docker containers may not work, so we may want to skip the ITs in the `jenkins` profile. I'm going to let my previous changes build again, and see what the results are. But I may make an additional change soon, though it depends on https://github.com/confluentinc/common/pull/215 (which I hope will be approved and merged today).

NathanNam:
This is great! @purbon

PR #328, opened 2019-08-01
MINOR: Fix SecurityIT to work on different host IPs

rhauch:
The build runs locally, but fails on Jenkins because the Elasticsearch TestContainer returns the IP address for the host on Jenkins but `locahost` on a dev machine. This is likely due to how Docker is configured on Jenkins vs locally. This change always makes the test use `localhost` for the ES address.


rhauch:
Related to #334. Closing, since this approach didn't work.

PR #335, opened 2019-08-22
MINOR: Upgrade Mockito to work on Java 11

rhauch:


rhauch:
@C0urante This issue was fixed in `2.23.4`. We previously used a different version, so I'd like to merge this and then deal with being different than `common` in a different PR.

BTW, this build will fail because of the `SecurityIT` problem; see #334, which is broken until this is fixed. I wanted to keep these changes separate in case we want to ultimately revert #334 at some point.

PR #337, opened 2019-08-23
CC-4886: Fix the SecurityIT integration test

rhauch:
Previous incarnations of the test used the `org.testcontainers.elasticsearch.ElasticsearchContainer` class, but the way we were pre-generating the certificates used to enable SSL that were valid when Docker allowed connecting to the container via `localhost` (which is what the certificates used). However, on Jenkins this was not the case, and on Jenkins the generated certs were not trusted and prevented the connector from actually establishing a connection.

The current changes involve creating a new custom `ElasticsearchContainer` subclass that encapsulates the ability to optionally enable SSL. When SSL is enabled, it builds the connector differently and dynamically generates the certs using the local IP address of the running container; the test then downloads the necessary truststore and keystore so that the connector can use it.

The generation of the SSL certs uses Elasticsearch tools that come with Elasticsearch 7 in the container, using the techniques documented in https://github.com/elastic/stack-docker. However, the normal Elasticsearch client doesn’t require a keystore and truststore (just the ES certs), whereas our connector uses the Jest client and like the other Kafka components does use them.


rhauch:
@purbon, @cyrusv: I think I finally cracked the code about getting an integration test that sets up an ES sink connector to talk to a secured Elasticsearch service running in a Docker container. Basically, the certs need to be generated at runtime and need to include the Docker host's IP address (not the container's), at least on Jenkins. I also learned a lot about TestContainers and advanced configuration - way more than I wanted to learn.

The nice thing is that all of these ITs run just like JUnit tests in IntelliJ, locally via the Maven build (using any combination of tests), and finally now on Jenkins.

I ended up having to completely customize the `ElasticsearchContainer` from TestContainers, which had no support for secured Elasticsearch. The good news is that the test cases are super simple, as they should be.

src/test/resources/ssl/start-elasticsearch.sh
```diff
@@ -0,0 +1,93 @@
...
+export PATH=/usr/share/elasticsearch/jdk/bin/:$PATH
+
+if [[ -z "${IP_ADDRESS}" ]]; then
+    IP_ADDRESS=$(hostname -I)
+fi
+
+echo
+echo "Replacing the ip address in the ${ES_DIR}/config/ssl/instances.yml file with ${IP_ADDRESS}"
+sed -i "s/ipAddress/${IP_ADDRESS}/g" ${ES_DIR}/config/ssl/instances.yml
+#cat ${ES_DIR}/config/ssl/instances.yml
```
cyrusv: Cruft

src/test/resources/ssl/start-elasticsearch.sh
```diff
@@ -0,0 +1,93 @@
...
+        rm -rf ${ES_DIR}/config/ssl/cluster
+    fi
+    echo "Unzipping cluster certs zipfile..."
+    unzip ${ES_DIR}/config/ssl/cluster.zip -d ${ES_DIR}/config/ssl/cluster
+    rm -f ${ES_DIR}/config/ssl/cluster.zip
+
+    #echo "Move logstash certs to logstash config dir..."
+    #mv /config/ssl/docker-cluster/logstash/* /config/logstash/
+    #echo "Move kibana certs to kibana config dir..."
+    #mv /config/ssl/docker-cluster/kibana/* /config/kibana/
```
cyrusv: Cruft

rhauch: Removed.

src/test/resources/ssl/start-elasticsearch.sh
rhauch: Removed

PR #338, opened 2019-08-30
Add indices to the internal cache if exist and not created by the connector

### LINKED ISSUE #332: performance degradation in case of existing index
Hello.
I was able to send ~8k docs per seconds when I started 1 task. And this task initiates a new index creation. But when I add any additional number of tasks, or just a create connector which sends data to the existing index the sending rate is never cross 300-500 docs/s level.
image

purbon:
If an index exist in Elasticsearch and is used by the connector, during the index creation verification we check with the server if the required index exist.

This call is protected with an internal cache, to not overflow the elasticsearch server.  However this case has not been covered in previous improvements.

This PR add the index to the cache in that case, covering the two cases, an index created by the connector or directly in elasticsearch.

Fix #332


levzem:
Closing duplicate PR from #340 and #309. Thanks @purbon!

PR #364, opened 2019-12-09
CC-7367: Add a lot more logging

rhauch:
Some logging was added at INFO level when it would not occur under nominal operation, more at DEBUG level to identify what the connector is doing and when while still not writing individual record details, and a few more at TRACE that output which messages are being processed.

I tried to be a bit conservative in terms of the log level, so as to not overwhelm the log during normal or even debug operation.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -30,6 +30,8 @@ import org.apache.kafka.connect.errors.DataException;
 import org.apache.kafka.connect.json.JsonConverter;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.storage.Converter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.math.BigDecimal;
 import java.nio.charset.StandardCharsets;
@@ -47,6 +49,7 @@ import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConst
 
 public class DataConverter {
 
+  private static final Logger log = LoggerFactory.getLogger(DataConverter.class);
   private static final Converter JSON_CONVERTER;
 
   static {
@@ -115,6 +118,14 @@ public class DataConverter {
     if (record.value() == null) {
       switch (behaviorOnNullValues) {
         case IGNORE:
+          if (log.isTraceEnabled()) {
+            log.trace(
+                "Ignoring record with null value at topic '{}', partition {}, offset {}",
+                record.topic(),
+                record.kafkaPartition(),
+                record.kafkaOffset()
+            );
+          }
           return null;
         case DELETE:
           if (record.key() == null) {
@@ -125,9 +136,26 @@ public class DataConverter {
             // offset information for the SinkRecord. Since that information is guaranteed to be
             // unique per message, we can be confident that there wouldn't be any corresponding
             // index present in ES to delete anyways.
+            if (log.isTraceEnabled()) {
+              log.trace(
+                  "Ignoring record with null key at topic '{}', partition {}, offset {}, since "
+                  + "the record key is used as the ID of the index",
+                  record.topic(),
+                  record.kafkaPartition(),
+                  record.kafkaOffset()
+              );
+            }
             return null;
           }
           // Will proceed as normal, ultimately creating an IndexableRecord with a null payload
+          if (log.isTraceEnabled()) {
+            log.trace(
+                "Deleting from Elasticsearch record at topic '{}', partition {}, offset {}",
+                record.topic(),
+                record.kafkaPartition(),
+                record.kafkaOffset()
+            );
+          }
           break;
         case FAIL:
           throw new DataException(String.format(
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -118,14 +118,12 @@ public class DataConverter {
     if (record.value() == null) {
       switch (behaviorOnNullValues) {
         case IGNORE:
-          if (log.isTraceEnabled()) {
-            log.trace(
-                "Ignoring record with null value at topic '{}', partition {}, offset {}",
-                record.topic(),
-                record.kafkaPartition(),
-                record.kafkaOffset()
-            );
-          }
+          log.trace(
+              "Ignoring record with null value at topic '{}', partition {}, offset {}",
+              record.topic(),
+              record.kafkaPartition(),
+              record.kafkaOffset()
+          );
           return null;
         case DELETE:
           if (record.key() == null) {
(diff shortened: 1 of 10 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -143,6 +143,11 @@ public void start(Map<String, String> props, ElasticsearchClient client) {
 
       writer = builder.build();
       writer.start();
+      log.info(
```
gharris1727: "we know this value from the configs - is this superfluous" -lev

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -235,16 +235,26 @@ public ElasticsearchWriter build() {
   }
 
   public void write(Collection<SinkRecord> records) {
+    log.debug("Writing {} records to Elasticsearch", records.size());
```
gharris1727: "duplicates the put() log message considering put() calls write()" - lev

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -212,40 +212,58 @@ private Version getServerVersion() throws IOException {
...
     } else if (esVersion.startsWith("7.")) {
-      return Version.ES_V7;
+      version = Version.ES_V7;
+      log.info("Detected Elasticsearch version is {}", version);
+    } else {
+      version = defaultVersion;
+      log.info("Detected unexpected Elasticsearch version {}, using {}", esVersion, version);
     }
-    return defaultVersion;
+    return version;
```
gharris1727: I think one log message can subsume the rest of these
```suggestion
    log.info("Detected Elasticsearch version is {}", version);
    return version;
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
rhauch: If you have the code, then it is superfluous. But if you don't, it can be very confusing to know which config controls this value in the log. I've seen cases when users were trying to control things like this, but they were definitely unsure which property set it. It shouldn't take trial and error to get that to work.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
rhauch: ack

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
rhauch: Only if we don't want to know about the final `else` block case. I was trying not to restructure this code to make the logging simpler, and preferred to leave the logic untouched and to have extra logging code.

rhauch:
Thanks for the review, @gharris1727. I've incorporated or responded to all feedback so far.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -115,6 +118,14 @@ public IndexableRecord convertRecord(
     if (record.value() == null) {
       switch (behaviorOnNullValues) {
         case IGNORE:
+          if (log.isTraceEnabled()) {
+            log.trace(
+                "Ignoring record with null value at topic '{}', partition {}, offset {}",
+                record.topic(),
```
gharris1727: I think these getters are relatively cheap, and the real cost for this log message (toString and writing to the logger) is already guarded internally.
I added this guard to all the trace calls in my testing, and didn't notice a significant change in throughput, so I think this may be some premature optimization.

Disabling the cyclomatic complexity check and decreasing the overall code readability is probably not worth the microsecond saved here.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -125,15 +132,17 @@ private Runnable farmerTask() {
     return new Runnable() {
       @Override
       public void run() {
-        log.debug("Starting farmer task");
-        try {
-          while (!stopRequested) {
-            submitBatchWhenReady();
+        try (LogContext context = logContext.create("Farmer1")) {
```
gharris1727: What's the significance of the `1` here, if it's just a constant number it seems like noise.
```suggestion
        try (LogContext context = logContext.create("Farmer")) {
```

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -324,6 +353,34 @@ public void flush(long timeoutMs) {
     } finally {
       flushRequested = false;
     }
+    if (log.isDebugEnabled()) {
```
gharris1727: okay this guard is definitely not necessary.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
gharris1727: The logic has already been refactored in this PR, you've changed the `return` flow. I agree that it's the same control flow overall, and I don't think it should be put back the way it was. I just don't think that it's a valid argument as to why you wouldn't refactor the log messages.

You can leave the else log message, since that's valuable. I would even make it a WARN level.

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchSinkTaskIT.java
```diff
@@ -27,12 +27,32 @@
...
+import org.slf4j.MDC;
 
 @Category(IntegrationTest.class)
 public class ElasticsearchSinkTaskIT extends ElasticsearchIntegrationTestBase {
 
+  private ElasticsearchSinkTask task = new ElasticsearchSinkTask();
+
+  @Before
+  public void beforeEach() {
+    //MDC.put("connector.context", "[MyConnector|task1] ");
```
gharris1727: Should this be active for the test? The remove is uncommented, and I'm assuming that's a no-op if the parameter is not configured.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
rhauch: I left this as `Farmer1` because it aligns the log messages cleanly with `. I'd prefer to keep it unless you *really* prefer not to.

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchSinkTaskIT.java
rhauch: Will fix.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
gharris1727: haha okay.
Maybe a space instead?
```suggestion
        try (LogContext context = logContext.create("Farmer ")) {
```

gharris1727: Oh okay, I didn't realize thread had a number after it as well.
This is fine as-is I suppose.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
rhauch: Well, I thought it was because `time.seconds()` is not free, and actually has a surprisingly non-trivial overhead.

gharris1727: This is only triggered during a flush, which is a very infrequent operation.
I've found that the only log messages and operations that affect performance are in the tight loop being executed per-record.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
rhauch: Fixed.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
rhauch: ack

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
rhauch: The WARN log messages were there before; I just clarified the message enough that they now need to span multiple lines. And it's a warning because the behavior is a function of the version, so if we can't get an actual version then the behavior may not be correct. It's not an error because we don't know for sure.

I've tried restructuring this section of code, and I just don't see an easy way to do it that keeps the code readable and straightforward.

rhauch:
@gharris1727, I would appreciate another review. Thanks!

rhauch:
I ran a system test run with the changes in this PR: https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/3165/, and the test runs passed.

rhauch:
Ran another system test run: https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/3166/

PR #373, opened 2020-01-10
CC-7644: Add support for http(s) proxy

levzem:
follow up PR on #277

thank you @sslavian812

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -15,6 +15,7 @@
 
 package io.confluent.connect.elasticsearch;
 
+import io.confluent.connect.elasticsearch.jest.JestElasticsearchClient;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Importance;
@@ -195,7 +196,8 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   // Proxy group
   public static final String PROXY_HOST_CONFIG = "proxy.host";
   public static final String PROXY_HOST_DISPLAY = "Proxy Host";
-  public static final String PROXY_HOST_DOC = "The address of the proxy host to connect through.";
+  public static final String PROXY_HOST_DOC = "The address of the proxy host to connect through. "
+      + "Supports the basic authentication scheme only.";
   public static final String PROXY_HOST_DEFAULT = "";
 
   public static final String PROXY_PORT_CONFIG = "proxy.port";
@@ -445,7 +447,6 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         "Create indices at startup"
     );
   }
-
 
   public boolean secured() {
     SecurityProtocol securityProtocol = securityProtocol();
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -37,6 +37,10 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
       + "http://eshost2:9200`` or ``https://eshost3:9200``. HTTPS is used for all connections "
       + "if any of the URLs starts with ``https:``. A URL without a protocol is treated as "
       + "``http``.";
+  public static final String CONNECTION_PROXY_CONFIG = "connection.proxy";
+  private static final String CONNECTION_PROXY_DOC =
+      "A URL of Elasticsearch HTTP(S) proxy e.g. ``https://proxy1:443``,"
+          + " ``http://proxy2:9876``, etc. Format: (http|https)://<host>:<port>";
   public static final String CONNECTION_USERNAME_CONFIG = "connection.username";
   private static final String CONNECTION_USERNAME_DOC =
       "The username used to authenticate with Elasticsearch. "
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -193,6 +199,52 @@ public static HttpClientConfig getClientConfig(ElasticsearchSinkConnectorConfig
     return builder.build();
   }
 
+  private static void configureProxy(ElasticsearchSinkConnectorConfig config,
+      HttpClientConfig.Builder builder) {
+
+    if (config.isBasicProxyConfigured()) {
+      HttpHost proxy = new HttpHost(
+          config.getString(ElasticsearchSinkConnectorConfig.PROXY_HOST_CONFIG),
+          config.getInt(ElasticsearchSinkConnectorConfig.PROXY_PORT_CONFIG)
```
aakashnshah: is this the general style of the ES connector? or can you make getter methods?

src/test/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClientTest.java
```diff
@@ -144,6 +147,33 @@ public void compressedConnectsSecurely() {
     assertTrue(httpClientConfig.isRequestCompressionEnabled());
   }
 
+  @Test
+  public void connectsWithProxy() throws NoSuchFieldException, IllegalAccessException {
```
aakashnshah: is there any test/any way to test proxy with auth?

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
levzem: general style - altho maybe time to slowly start moving the style over

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -369,6 +442,42 @@ private SecurityProtocol securityProtocol() {
...
+    return !getString(PROXY_HOST_CONFIG).isEmpty();
+  }
+
+  public boolean isProxyWithAuthenticationConfigured() {
+    return isBasicProxyConfigured()
+        && !getString(PROXY_USERNAME_CONFIG).isEmpty()
+        && getPassword(PROXY_PASSWORD_CONFIG) != null;
+  }
+
+  private void validateProxyConfigs() {
```
cyrusv: Do you think we can add a validator that the proxy can be connected to?   This will validate that the user is putting in a real host, their firewall/network is correct, etc.    My concern is that if someone does something like misspell their hostname or put the wrong username/password, they will see some weird Java stacktrace, but it will be much better to see a validation error.

This would kind of be a step higher than we've done with validators in the past, but I think it will vastly improve UX and reduce confusion by users, make them dig through fewer stacktraces where it's unclear if it is user error or a true bug.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfigTest.java
```diff
@@ -94,4 +100,44 @@ public void testSecured() {
     props.put("connection.url", "https://host:9999");
     assertFalse(new ElasticsearchSinkConnectorConfig(props).secured());
   }
+
+  @Test
+  public void shouldAcceptValidBasicProxy() {
```
cyrusv: Nice unit tests

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -186,10 +189,32 @@
           + "Values can be `PLAINTEXT` or `SSL`. If `PLAINTEXT` is passed, "
           + "all configs prefixed by " + CONNECTION_SSL_CONFIG_PREFIX + " will be ignored.";
 
+  // Proxy group
+  public static final String PROXY_HOST_CONFIG = "proxy.host";
+  public static final String PROXY_HOST_DISPLAY = "Proxy Host";
+  public static final String PROXY_HOST_DOC = "The address of the proxy host to connect through.";
```
cyrusv: Can we be more explicit about which proxies are and aren't supported?  I remember we had an issue where it was unclear if we supported NTLM proxy.

levzem: addressed

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
levzem: basically the way I see this going is we try to setup a connection and ping some website i.e `confluent.io` through their proxy. This is a pretty heavy weight validation. If you have a better suggestion, I am open to it.

I think this is definitely an awesome validation for cloud specifically and we should have very heavy duty validation for as much as possible in all of the cloud connectors. Is it worth it right for on prem? Should we implement this validation as we onboard to cloud? Should we develop some sort of framework or approach to this heavy duty validation? 

I am open to implementing it now but I think then we should develop a standard for it and do it right

wicknicks: why would you have to ping `confluent.io`? can't we add logic to the validate() method in the connector to make a connection to the ES server via the proxy and check if the server exists and try to do some test reads/writes?

levzem: I can create a `new JestElasticsearchClient()` in the config which will try to connect to ES to get the version and will fail if it can't connect and that will test the entire connection configs... it seems a bit heavyweight but how does that sound?

wicknicks: and the same method is used in the tasks as well? if so, then sgtm.

levzem: turns out creating a `new JestElasticsearchClient()` will not work because it creates its own internal `Config` which results in an infinite loop of creating each other. I am opting to not do full validation that connection will be successful as that will be quite involved and am going to just file a ticket to get this done. 

The tasks already create their own client, which will fail if it can't connect.

Not sure by what you are asking, but the above method is called every time a `config` is constructed, so it will be validated immediately.

levzem:
thanks for your contribution @sslavian812

PR #384, opened 2020-02-06
Support retry_on_conflict parameter Fixes #374

### LINKED ISSUE #374: Ignoring version conflicts for items causes loss of data
We noticed this issue after our Elasticsearch data got corrupted. We checked the topics and the data was not matching.

When reading from multiple topics and using the UPSERT write strategy, there is a high chance that update conflicts will happen on Elasticsearch. The current logic ignores any update conflicts reported by Elasticsearch, leading to data loss.
```
WARN Ignoring version conflicts for items: [Key{index_v3/_doc/US725754}] (io.confluent.connect.elasticsearch.jest.JestElasticsearchClient)
```

I can think of three possible solutions to this:

1. Retry on the server side by using [retry_on_conflict](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#docs-update-api-query-params).
This is an easy and efficient approach. It might be a good idea to add a configuration parameter to enable the user to set how many update retries they want to allow, but even a hard set number of retries (as bellow) is a better default than missing updates and ending up with corrupt data. 
To add the retry we can change this code:
```
// JestElasticsearchClient.java
private Update toUpdateRequest(IndexableRecord record) {
    String payload = "{\"doc\":" + record.payload
        + ", \"doc_as_upsert\":true}";
    return new Update.Builder(payload)
        .index(record.key.index)
        .type(record.key.type)
        .id(record.key.id)
        .setParameter("retry_on_conflict", 5)  // here: including the server retry query parameter
        .build();
  }
```

2. Retry on the client side using the existing retry logic. 
In this case it might be a good idea to filter the batch items to only include the conflicting documents.
To do that we have to modify the existing logic in `JestElasticsearchClient.executeBulk(BulkRequest)`, because it currently ignores update conflicts.

3. Assign the same partition numbers (from different topics) to the same connector Task.
This way no concurrency issues would happen, UNLESS there are other applications writing to the same index. I am not sure, but this does not look like a trivial implementation.

adrianotr:
Fixes https://github.com/confluentinc/kafka-connect-elasticsearch/issues/374

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -345,6 +345,16 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         ++order,
         Width.SHORT,
         "Create indices at startup"
+    ).define(
+        RETRY_ON_CONFLICT_CONFIG,
+        Type.INT,
+        0,
+        Importance.LOW,
+        RETRY_ON_CONFLICT_DOC,
+        group,
+        ++order,
+        Width.SHORT,
+        "Retry on conflict"
     );
   }
 
@@ -471,17 +481,7 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         group,
         ++order,
         Width.SHORT,
-        "Write method"
-    ).define(
-        RETRY_ON_CONFLICT_CONFIG,
-        Type.INT,
-        0,
-        Importance.LOW,
-        RETRY_ON_CONFLICT_DOC,
-        group,
-        ++order,
-        Width.SHORT,
-        "Retry on conflict");
+        "Write method");
   }
 
   public static final ConfigDef CONFIG = baseConfigDef();
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -166,6 +166,11 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
           + "increasing the " + FLUSH_TIMEOUT_MS_CONFIG + ", " + READ_TIMEOUT_MS_CONFIG
           + ", and decrease " + BATCH_SIZE_CONFIG + " configuration properties.";
 
+  public static final String RETRY_ON_CONFLICT_CONFIG = "retry.on.conflict";
+  private static final String RETRY_ON_CONFLICT_DOC = "Specify how many times the operation "
+      + "should be retried by Elasticsearch, when conflicts occur, while using the write method "
+      + WriteMethod.UPSERT.toString() + ". The value is the number of retries.";
+
   public static final String CONNECTION_SSL_CONFIG_PREFIX = "elastic.https.";
 
   public static final String AUTO_CREATE_INDICES_AT_START_CONFIG = "auto.create.indices.at.start";
@@ -466,7 +471,17 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
         group,
         ++order,
         Width.SHORT,
-        "Write method");
+        "Write method"
+    ).define(
+        RETRY_ON_CONFLICT_CONFIG,
+        Type.INT,
+        0,
+        Importance.LOW,
+        RETRY_ON_CONFLICT_DOC,
+        group,
+        ++order,
+        Width.SHORT,
+        "Retry on conflict");
   }
 
   public static final ConfigDef CONFIG = baseConfigDef();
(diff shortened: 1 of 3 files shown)


adrianotr:
[clabot:check]

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -466,7 +471,17 @@ private static void addConversionConfigs(ConfigDef configDef) {
...
+    ).define(
+        RETRY_ON_CONFLICT_CONFIG,
+        Type.INT,
+        0,
+        Importance.LOW,
+        RETRY_ON_CONFLICT_DOC,
+        group,
+        ++order,
+        Width.SHORT,
+        "Retry on conflict");
```
levzem: I think this belongs in the connector configs group

src/test/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClientTest.java
```diff
@@ -33,12 +33,7 @@
 import io.searchbox.client.config.ElasticsearchVersion;
 import io.searchbox.client.config.HttpClientConfig;
 import io.searchbox.cluster.NodesInfo;
-import io.searchbox.core.BulkResult;
-import io.searchbox.core.Delete;
-import io.searchbox.core.Index;
-import io.searchbox.core.Search;
-import io.searchbox.core.SearchResult;
-import io.searchbox.core.Update;
+import io.searchbox.core.*;
```
levzem: our convention is not to use *

src/test/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClientTest.java
```diff
@@ -375,6 +370,33 @@ public void toBulkableAction(){
     assertEquals("{\"doc\":" + idx.payload + ", \"doc_as_upsert\":true}", ba.getData(null));
   }
 
+  @Test
+  public void retryOnConflict() throws Exception {
```
levzem: is this actually replicating a conflict?

adrianotr: No, it's checking for the presence of the `retry_on_conflict` parameter in the request. The retry is done on the server-side by Elasticsearch.

src/test/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClientTest.java
adrianotr: Sorry, my IDE did that automatically. I'll fix that.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
adrianotr: Makes sense. I'll change it. Thanks

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -166,6 +166,11 @@
           + "increasing the " + FLUSH_TIMEOUT_MS_CONFIG + ", " + READ_TIMEOUT_MS_CONFIG
           + ", and decrease " + BATCH_SIZE_CONFIG + " configuration properties.";
 
+  public static final String RETRY_ON_CONFLICT_CONFIG = "retry.on.conflict";
+  private static final String RETRY_ON_CONFLICT_DOC = "Specify how many times the operation "
+      + "should be retried by Elasticsearch, when conflicts occur, while using the write method "
```
levzem:
```suggestion
      + "should be retried by Elasticsearch, when write conflicts occur, while using the write method "
```

levzem:
@adrianotr you have a checkstyle error
```[2020-02-11T22:13:19.642Z] [ERROR] /home/jenkins/workspace/fka-connect-elasticsearch_PR-384/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java:223:3: Method length is 154 lines (max allowed is 150). [MethodLength]```

you can add a checkstyle supression in `checkstyle/supressions.xml`

abezardpro:
Still Opened ? Il could be a useful improvment

abezardpro:
> Still Opened ? It could be a useful improvment

@levzem is there a way to help get this pull request through quickly ?

levzem:
@abezardpro

 if you can rebase this branch on `master` one more time, fix the checkstyle error, and address the nit I mentioned, I will happily LGTM and merge. Running short on time myself. 

I would appreciate the help :) Let me know if you are able to.

abezardpro:
> @abezardpro
> 
> if you can rebase this branch on `master` one more time, fix the checkstyle error, and address the nit I mentioned, I will happily LGTM and merge. Running short on time myself.
> 
> I would appreciate the help :) Let me know if you are able to.

Sure i am !

levzem:
@abezardpro thanks much appreciated!

if you can do it in this PR that would be great. But if you can't push to @adrianotr 's branch, you can open a new PR and tag me on it.

ernestas-vinted:
This change would solve our issue as well

LeonardoBonacci:
Any update on this issue?

dosvath:
Taking over in https://github.com/confluentinc/kafka-connect-elasticsearch/pull/452 as this branch had no push permissions.

Issue #325, opened 2019-07-24
Make "behavior.on.malformed.documents" also cover "action_request_validation_exception"


milofelipe:
Background: I have "key.ignore=false" and have old messages in my topic that don't have keys or have empty keys. Connector tasks fail because Elasticsearch is returning a validation error:

{"root_cause":[{"type":"action_request_validation_exception","reason":"Validation Failed: 1: an id must be provided if version type or value are set;"}],"type":"action_request_validation_exception","reason":"Validation Failed: 1: an id must be provided if version type or value are set;"}

Would be great if there is an option to ignore errors like this so tasks don't stop running. Looking at the code, "behavior.on.malformed.documents" seems to be only concerned with "mapper_parsing_exception" and "illegal_argument_exception" errors. Requesting to add "action_request_validation_exception" to the list or implement a new configuration option for this.

levzem:
@milofelipe this looks a relatively simple fix. Would you like to issue a PR?

milofelipe:
> @milofelipe this looks a relatively simple fix. Would you like to issue a PR?

Yes.

levzem:
closing because merged, thanks @milofelipe!

PR #388, opened 2020-02-15
Make "behavior.on.malformed.documents" also cover "action_request_validation_exception" (#325)

milofelipe:
Background: I have "key.ignore=false" and have old messages in my topic that don't have keys or have empty keys. Connector tasks fail because Elasticsearch is returning a validation error:

{"root_cause":[{"type":"action_request_validation_exception","reason":"Validation Failed: 1: an id must be provided if version type or value are set;"}],"type":"action_request_validation_exception","reason":"Validation Failed: 1: an id must be provided if version type or value are set;"}

Ignore errors like this so tasks don't stop running. Addig "action_request_validation_exception" to the list covered by "behavior.on.malformed.documents".


milofelipe:
[clabot:check]

levzem:
@milofelipe this looks good to me, let's just update `ElasticsearchSinkConnectorConfig.BEHAVIOR_ON_MALFORMED_DOCS_DOC` to document this change.

PR #391, opened 2020-02-18
Use Elasticsearch auto-generated ids for performance

gigsda:
this option is for someone who need better indexing performance but not need 'exactly once'
- https://www.elastic.co/guide/en/elasticsearch/reference/master/tune-for-indexing-speed.html#_use_auto_generated_ids


gigsda:
[clabot:check=gigsda]

PR #398, opened 2020-03-16
CC-8284: Log errors during index exist check at info level

ncliang:


ncliang:
Good catch @levzem . I meant to target `5.2.x`, not `master` :-p

PR #404, opened 2020-04-24
CC-8869: Fix kafka sslfactory changes causing ES connector to break

aakashnshah:
The `SslFactory` class in Apache Kafka was recently changed and no longer includes `sslEngineBuilder()`. The `SslEngineBuilder` class has also been removed from Apache Kafka and a new class, `SslEngineFactory` is being used instead. There wasn't a clean way to leverage the new API to provide an `SSLContext` to the ElasticSearch connector, so instead I copied the code from AK into the connector.

Signed-off-by: Aakash Shah <ashah@confluent.io>

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
@@ -61,6 +61,7 @@ import org.apache.http.nio.conn.ssl.SSLIOSessionStrategy;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.config.types.Password;
 import org.apache.kafka.common.network.Mode;
+import org.apache.kafka.common.security.ssl.DefaultSslEngineFactory;
 import org.apache.kafka.common.security.ssl.SslFactory;
 import org.apache.kafka.common.utils.SystemTime;
 import org.apache.kafka.common.utils.Time;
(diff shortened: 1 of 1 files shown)


cyrusv:
@ewencp, the concern with updating like this is because we've had this regression happen before -- it's generally just unwise for us to depend on an API that is not actually a public API. Moving forward our connectors won't ship with CP, so we're also going to need our connectors to be compatible across versions of AK. It would be ideal if these classes were a public API that was maintained so we could import it with more confidence, but this is not the first time we've been burned on it.

ewencp:
We don't have a great track record with copy/pasting and then maintaining, but if we think we need to do that, this probably isn't the right place for it. The usage isn't unique to this repo, so if we're going to make the tradeoff of maintaining a copy, it should be managed centrally. I can see at least basically the same usage in ksql, kafka-connect-rabbitmq, and at least newwave also uses SslFactory in a non-test context. (There's also kafka-broker-plugins but that's only for older branches so shouldn't be affected unless we want to backport for compatibility and maintenance reasons.)

Also, should make sure someone from core consults on a choice to copy code in like this. While there are definitely compatibility risks, different parts of the code, even if all internal, have different risk profiles wrt changes, compatibility, etc.

aakashnshah:
@ewencp I agree with maintaining a central copy of the code if we decide to copy it, and the potential risks that come with it. This was not ideal, and only done as what I presumed to be the only minimal way. Thanks for the example, I've used that instead as the subsitution.

@cyrusv I agree with this strategy in the long term, but since it is a blocker for packing, system testing, etc. it is quite high priority and I think for now we should make this simple fix on master and then loop back for a permanent fix when it is not a blocker. WDYT?

C0urante:
+1 for merging this right now to stop the bleeding, and looking into a general-purpose library that could be shared among our connectors (if not among all of our services) at a later date.

ijuma:
Can we merge this simple PR to unblock everything else and do a follow up to handle all concerns?

aakashnshah:
> Can we merge this simple PR to unblock everything else and do a follow up to handle all concerns?

Yes, we are currently running a system test to verify everything works fine with this fix, and then we will promptly merge it.

C0urante:
Green build here: https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/3772

Merging now, we will revisit this strategy before publishing the 6.0 release so that this connector doesn't break on pre-6.0 versions of CP.

PR #410, opened 2020-05-26
CC-9314: Allow disabling hostname verification

ncliang:
When `elastic.https.ssl.endpoint.identification.algorithm` is empty, instead
of default hostname verifier, use an all-trusting verifier.

Updated SecurityIT with testcase to verify

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -25,7 +25,6 @@ import java.util.Map;
 
 import static io.confluent.connect.elasticsearch.DataConverter.BehaviorOnNullValues;
 import static io.confluent.connect.elasticsearch.bulk.BulkProcessor.BehaviorOnMalformedDoc;
-import static org.apache.http.util.TextUtils.isBlank;
 import static org.apache.kafka.common.config.SslConfigs.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG;
 import static org.apache.kafka.common.config.SslConfigs.addClientSslSupport;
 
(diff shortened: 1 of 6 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -25,6 +25,8 @@
 
 import static io.confluent.connect.elasticsearch.DataConverter.BehaviorOnNullValues;
 import static io.confluent.connect.elasticsearch.bulk.BulkProcessor.BehaviorOnMalformedDoc;
+import static org.apache.http.util.TextUtils.isBlank;
```
wicknicks, line 28: nit: `import org.apache.commons.lang.StringUtils` instead?

ncliang: I don't think we currently pull in apache commons and I didn't think it is worth it to add the dependency just for this method.

ncliang:
Thanks for review @wicknicks and @levzem . I think you are right that w.r.t. backwards compatibility, there is a risk that someone may have accidentally set this config to empty string and it would have been doing nothing before but now it will disable hostname verification for them. I think this would likely be fine though given that the default for this config is "https" and we have documentation that says setting `ssl.endpoint.identification.algorithm` to empty string will disable hostname verification [1]. Users that have explicitly set this to empty string were likely trying to disable hostname verification and failing. They may have since worked around the issue by adding the hostname to their key/trust store but forgot to remove the setting from their config. In which case, this change would be doing them no harm.

The ConfigDef comes from AK `SslConfigs` so updating the doc there is none-trivial. I can update our website documentation [2] to add something talking about how empty string will disable hostname verification though.

[1]: https://docs.confluent.io/current/kafka/authentication_ssl.html
[2]: https://docs.confluent.io/current/connect/kafka-connect-elasticsearch/configuration_options.html

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
wicknicks: that's fair, but I'm not sure if this is intended to be a public API though, for the http library. given the complexity of the code, I'd be more inclined to simply make a copy of the function here.

wicknicks:
@ncliang that's fine to do as well but maybe not in a patch release, since it sounds like the intended behavior is to verify hosts (and seems like it was doing fine). so maybe change the base of the branch to master, if that's the path we want to take?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
ncliang: I remove the use of this method and replace with simple `isEmpty()` check. The difference is that null and strings containing spaces will not trigger the feature, which I think is fine since our documentation explicitly says empty string.

PR #411, opened 2020-05-27
CC-9261: call get() on futures so that we properly handle exceptions …

ncliang:
…thrown in executor threads

https://stackoverflow.com/a/39091808/12414477

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
@@ -167,7 +167,10 @@ public class BulkProcessor<R, B> {
       wait(Math.max(0, lingerMs - elapsedMs));
     }
     // at this point, either stopRequested or canSubmit
-    return stopRequested ? Futures.immediateFuture(null) : submitBatch();
+    return stopRequested
+            ? Futures.immediateFuture(
+                    BulkResponse.failure(false, "request not submitted during shutdown"))
+            : submitBatch();
   }
 
   private synchronized Future<BulkResponse> submitBatch() {
--- a/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
@@ -18,6 +18,8 @@ package io.confluent.connect.elasticsearch.bulk;
 
 import io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig;
 import io.confluent.connect.elasticsearch.RetryUtil;
+
+import com.google.common.util.concurrent.Futures;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.connect.errors.ConnectException;
@@ -30,6 +32,7 @@ import java.util.Deque;
 import java.util.List;
 import java.util.Locale;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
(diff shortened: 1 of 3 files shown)


ncliang:
@levzem @wicknicks the `UncaughtExceptionHandler` on the bulk processor threads were not being called because we use `submit` to submit the bulk requests instead of `execute`. I still like the submit and Future interface since that makes testing easier. The approach here is to keep track of these futures and call `get` on them after they are complete so that any thrown exceptions will get to the farmer thread wrapped in an `ExecutionException`. At this point, the farmer thread can rethrow it and trigger its `UncaughtExceptionHandler`.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -122,20 +125,35 @@ public Thread newThread(Runnable r) {
...
+          // after we submit, look at any previous futures that were completed and call get() on
+          // them so that exceptions are propagated.
+          List<Future<BulkResponse>> unfinishedFutures = new ArrayList<>();
+          for (Future<BulkResponse> f : futures) {
+            if (f.isDone()) {
+              log.debug("Calling get on completed future");
+              f.get();
+            } else {
+              unfinishedFutures.add(f);
+            }
```
wicknicks, line 147: should we just (submit all and) wait for the first one with a timeout? otherwise this goes into a tight loop, right?

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -149,7 +167,7 @@ public void run() {
       wait(Math.max(0, lingerMs - elapsedMs));
```
wicknicks, line 167: here, should we wait if the number of pending futures<> is larger than some limit?

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
ncliang: yeah, the logic here is a bit tricky. It takes into account both cases `unsentRecords>batchSize` and `unsentRecords<batchSize`. For the former case, we enter into a tight loop creating Futures, each sending request of size batchSize. For the later, we wait for `lingerMs` to see how much of the batch we can fill before sending it off.

I'd kind of prefer not to change this and focus on fixing the bug in exception propagation.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
levzem: that should be limited else where by max pending requests IIRC

ncliang: looks like `maxInFlightRequests` is used to limit the number of pending requests. It does this by creating the threadpool with that number of threads. Further submits would just be queued up and we would accumulate the futures in the futures queue. I am not sure I want to enforce a hard limit on the size of the futures queue since this could break backwards compatibility when handling huge spikes of traffic that we could handle before. WDYT?

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
wicknicks: just curious, why did the exception handler not catch these errors already? looks like we are already applying it to the `farmer` as well as every thread in the `executor`, so any error in any of these threads should be automatically propagated to that handler, which in turn should fail the task.

ncliang: The uncaught exception handler wouldn't be invoked here because with `submit` the exception is considered part of the future output and wrapped in `ExcecutionException` on `get`. This is different than if you use the `execute` method on `ExecutorService`. With `execute`, the exception would be caught and handled by the unhandled exception handler as expected. This change will have the `farmer` thread handle and rethrow the `ExecutionException`s from BulkProcessor threads. Since the `farmer` thread is started directly, the rethrown exception would get handled by the uncaught exception handler as expected.

ncliang: Better explanation here - https://stackoverflow.com/a/29050729/12414477

We did not catch this earlier because our tests are calling `get` on the returned future and blocking on the response while our production code was not.

wicknicks: sounds good. thanks, @ncliang!

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -122,20 +125,35 @@ public Thread newThread(Runnable r) {
...
+        while (!stopRequested) {
+          // submitBatchWhenReady waits for lingerMs so we won't spin here unnecessarily
+          futures.add(submitBatchWhenReady());
+
+          // after we submit, look at any previous futures that were completed and call get() on
+          // them so that exceptions are propagated.
+          List<Future<BulkResponse>> unfinishedFutures = new ArrayList<>();
+          for (Future<BulkResponse> f : futures) {
+            if (f.isDone()) {
+              log.debug("Calling get on completed future");
```
wicknicks, line 143: Since `get()` is guaranteed to return quickly, maybe we can add print some info from the response object here?

```suggestion
              log.debug("Bulk request completed with status {}", f.get());
```

`BulkResponse` would need a `toString()` method.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -122,20 +125,35 @@ public Thread newThread(Runnable r) {
...
+          // submitBatchWhenReady waits for lingerMs so we won't spin here unnecessarily
+          futures.add(submitBatchWhenReady());
+
+          // after we submit, look at any previous futures that were completed and call get() on
+          // them so that exceptions are propagated.
+          List<Future<BulkResponse>> unfinishedFutures = new ArrayList<>();
+          for (Future<BulkResponse> f : futures) {
+            if (f.isDone()) {
+              log.debug("Bulk request completed with status {}", f.get());
+              f.get();
```
wicknicks, line 144: you can remove this second `get()`.

ncliang: I'm unsure if the debug statements would get hit if log level is not at least DEBUG, so I left both in. Yeah, that's a bit confusing. Let me write that a little better.

wicknicks: `f.get()` will get evaluated always. the only thing that we will avoid with the debug level is the string formatting/concatenation.

PR #412, opened 2020-05-28
CC-9347: Inspect interrupted state before next attempt

ncliang:
During executor shutdown, it attempts to terminate existing tasks by sending interrupts to the threads. The `Time.sleep` that we use to sleep between retry attempts wraps around AK `Utils.sleep` that handles the `InterruptedException` by setting `Thread.currentThread().interrupt()`. We just need to inspect the interrupted state of current thread before making our next retry attempt.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
@@ -418,7 +418,7 @@ public class BulkProcessor<R, B> {
             time.sleep(sleepTimeMs);
             if (Thread.interrupted()) {
               log.error(
-                      "Shutdown interrupted batch {} of {} records after attempt {}/{}",
+                      "Retrying batch {} of {} records interrupted after attempt {}/{}",
                       batchId, batch.size(), attempts, maxAttempts, e);
               throw e;
             }
--- a/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
@@ -416,6 +416,12 @@ public class BulkProcessor<R, B> {
                       + "will attempt retry after {} ms. Failure reason: {}",
                       batchId, batch.size(), attempts, maxAttempts, sleepTimeMs, e.getMessage());
             time.sleep(sleepTimeMs);
+            if (Thread.currentThread().isInterrupted()) {
+              log.error(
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -395,6 +395,12 @@ private BulkResponse execute() throws Exception {
                       + "will attempt retry after {} ms. Failure reason: {}",
                       batchId, batch.size(), attempts, maxAttempts, sleepTimeMs, e.getMessage());
             time.sleep(sleepTimeMs);
+            if (Thread.currentThread().isInterrupted()) {
+              log.error(
```
wicknicks, line 399: clear the interrupted flag with `Thread.interrupted()` here?

kkonstantine: why is this is a error log message?

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -395,6 +395,12 @@ private BulkResponse execute() throws Exception {
                       + "will attempt retry after {} ms. Failure reason: {}",
                       batchId, batch.size(), attempts, maxAttempts, sleepTimeMs, e.getMessage());
             time.sleep(sleepTimeMs);
+            if (Thread.interrupted()) {
+              log.error(
+                      "Shutdown interrupted batch {} of {} records after attempt {}/{}",
```
kkonstantine, line 400: Also, adding the cause here seems too specific. Probably better to say something to signify that the thread that was waiting to retry a batch has been interrupted.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
kkonstantine: hmm never mind.

ncliang: I think error level is justified here because we failed to deliver the batches due to interrupt. Since we log error bellow if we fail to deliver after exhausting retries, it seems appropriate to log at error level here too.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
ncliang: I update the message to be a bit more generic and just talk about the retry being interrupted. PTAL?

PR #413, opened 2020-05-29
CC-9373: Skip only exceptions during conversion when `drop.invalid.message` is true

ncliang:


ncliang:
@levzem yes, that is the intent of the `drop.invalid.message` config - log and skip messages that are malformed. The bug that this PR is fixing is where exceptions thrown in `bulkprocessor.add` were also mistakenly skipped. I added a testcase to verify this by shutting down the bulk processor before attempting to process any (valid) records. Without this fix, the operation would succeed despite the bulkprocess.add() throwing due to it being stopped.

PR #418, opened 2020-06-14
Add query-parameter support for index/bulk APIs

javabrett:
Adds processing for configuration properties of the form `elasticsearch.index.param.<name>=<value>`.  The name/value should be a valid query-string parameter name and value for the Index/Bulk APIs, to be applied to API calls.

See:
- https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#docs-index-api-query-params
- https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#docs-bulk-api-query-params

That the original motivation for this PR is to allow configuration of pipeline-processing when writing records, e.g. `?pipeline=mypipelineinelastic`.  This change itself is not sufficient to allow that, since in the Jest client used by the connector, query-param names are filtered by an enumeration.  So to achieve testing of Elasticsearch pipelines, this Jest PR and a snapshot build are required: https://github.com/searchbox-io/Jest/pull/679 .

items worth review:
- Config property name/prefix
- Should separate config be required for Index and Bulk APIs, and maybe even for Delete (tombstone when configured) - if these have different subsets of allowed query-string parameters, or if different behaviour is desired for each operation.

# Testing

There's a `cp-demo` custom branch/build with an updated readme for testing: https://github.com/confluentinc/cp-demo/tree/javabrett/FF-1556-elasticsearch-test .  This includes a custom build which incorporates a Jest patch to demonstrate parameter support including `?pipeline` for date-based index-name processing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,7 +43,7 @@
         <lucene.version>5.5.2</lucene.version>
         <hamcrest.version>1.3</hamcrest.version>
         <mockito.version>2.28.2</mockito.version>
-        <jest.version>6.3.1</jest.version>
+        <jest.version>6.3.2-SNAPSHOT</jest.version>
         <test.containers.version>1.11.4</test.containers.version>
         <kafka.connect.maven.plugin.version>0.11.1</kafka.connect.maven.plugin.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -43,7 +43,7 @@
         <lucene.version>5.5.2</lucene.version>
         <hamcrest.version>1.3</hamcrest.version>
         <mockito.version>2.28.2</mockito.version>
-        <jest.version>6.3.1</jest.version>
+        <jest.version>6.3.2-SNAPSHOT</jest.version>
```
javabrett, line 46: Testing used snapshot built from https://github.com/searchbox-io/Jest/pull/679 , so that `?pipeline=...` could be used as a test, and this is also the desired functionality.

PR #421, opened 2020-06-22
CC-7635: implement DLQ reporter for malformed doc records

levzem:
## Problem
the connector currently can fail/ignore records that are malformed on the ES side. Reporting them using the new AK 2.6 reporter feature would be beneficial.

## Solution
Add the AK 2.6 reporter DLQ feature

- [ ] yes
- [x] no

## Test Strategy
modify unit tests to check whether the reporter is called

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
this is a new feature therefore pushing to `master`

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -91,7 +91,7 @@ public class ElasticsearchSinkTask extends SinkTask {
           config.getInt(ElasticsearchSinkConnectorConfig.MAX_RETRIES_CONFIG);
       boolean dropInvalidMessage =
           config.getBoolean(ElasticsearchSinkConnectorConfig.DROP_INVALID_MESSAGE_CONFIG);
-      boolean createIndicesAtStartTime =
+      final boolean createIndicesAtStartTime =
           config.getBoolean(ElasticsearchSinkConnectorConfig.AUTO_CREATE_INDICES_AT_START_CONFIG);
 
       DataConverter.BehaviorOnNullValues behaviorOnNullValues =
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -140,10 +140,15 @@ public class ElasticsearchSinkTask extends SinkTask {
           .setBehaviorOnMalformedDoc(behaviorOnMalformedDoc);
 
       try {
+        if (context.errantRecordReporter() == null) {
+          log.info("Errant record reporter not configured.");
+        }
+
         // may be null if DLQ not enabled
         builder.setErrantRecordReporter(context.errantRecordReporter());
       } catch (NoClassDefFoundError e) {
         // Will occur in Connect runtimes earlier than 2.6
+        log.warn("AK versions prior to 2.6 do not support the errant record reporter");
       }
 
       this.createIndicesAtStartTime = createIndicesAtStartTime;
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -139,6 +139,13 @@ public class ElasticsearchSinkTask extends SinkTask {
           .setBehaviorOnNullValues(behaviorOnNullValues)
           .setBehaviorOnMalformedDoc(behaviorOnMalformedDoc);
 
+      try {
+        // may be null if DLQ not enabled
+        builder.setErrantRecordReporter(context.errantRecordReporter());
+      } catch (NoClassDefFoundError e) {
+        // Will occur in Connect runtimes earlier than 2.6
+      }
+
       this.createIndicesAtStartTime = createIndicesAtStartTime;
 
       writer = builder.build();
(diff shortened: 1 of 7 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchWriter.java
```diff
@@ -53,6 +54,7 @@
 
   private final Set<String> existingMappings;
   private final BehaviorOnMalformedDoc behaviorOnMalformedDoc;
+  private final ErrantRecordReporter reporter;
```
kkonstantine, line 57: this member variable (like the one above) is unused in this class. It's used in the `BulkProcessor` instead.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -449,10 +459,21 @@ private BulkResponse execute() throws Exception {
                   System.currentTimeMillis() - startTime
               );
             }
+            batch.forEach(r -> recordMap.remove(r));
```
kkonstantine, line 462: how about? 
```suggestion
            recordMap.keySet().removeAll(batch);
```

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -449,10 +459,21 @@ private BulkResponse execute() throws Exception {
...
+            for (R record : batch) {
+              SinkRecord original = recordMap.get(record);
+              if (original != null && reporter != null) {
+                reporter.report(
+                    original,
+                    new ConnectException("Bulk request failed: " + bulkRsp.getErrorInfo())
+                );
+              }
+            }
+            batch.forEach(r -> recordMap.remove(r));
```
kkonstantine, line 476: same suggestion as above

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -449,10 +459,21 @@ private BulkResponse execute() throws Exception {
...
               );
             }
+            batch.forEach(r -> recordMap.remove(r));
             return bulkRsp;
           } else if (responseContainsMalformedDocError(bulkRsp)) {
             retriable = bulkRsp.isRetriable();
             handleMalformedDoc(bulkRsp);
+            for (R record : batch) {
+              SinkRecord original = recordMap.get(record);
+              if (original != null && reporter != null) {
```
kkonstantine, line 469: I read this and I'm thinking: should we make more things conditional to the `reporter` being not `null`?
At least this whole `for` loop does not need to run unless `reporter != null`, right?

src/test/java/io/confluent/connect/elasticsearch/bulk/BulkProcessorTest.java
```diff
@@ -413,6 +428,8 @@ public void ignoreOrWarnOnMalformedDoc() throws InterruptedException {
         fail(e.getMessage());
       }
     }
+
+    verify(reporter, times(4)).report(any(), any());
```
kkonstantine, line 432: would it make sense to capture and assert the exact record argument?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -139,6 +139,13 @@ public void start(Map<String, String> props, ElasticsearchClient client) {
           .setBehaviorOnNullValues(behaviorOnNullValues)
           .setBehaviorOnMalformedDoc(behaviorOnMalformedDoc);
 
+      try {
+        // may be null if DLQ not enabled
+        builder.setErrantRecordReporter(context.errantRecordReporter());
+      } catch (NoClassDefFoundError e) {
+        // Will occur in Connect runtimes earlier than 2.6
```
wicknicks, line 146: should we have a (warn?) message saying that errant record reporting will be disabled for this task?

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -453,6 +462,18 @@ private BulkResponse execute() throws Exception {
           } else if (responseContainsMalformedDocError(bulkRsp)) {
             retriable = bulkRsp.isRetriable();
             handleMalformedDoc(bulkRsp);
+            if (reporter != null) {
+              for (R record : batch) {
+                SinkRecord original = recordMap.get(record);
+                if (original != null) {
+                  reporter.report(
+                      original,
+                      new ConnectException("Bulk request failed: " + bulkRsp.getErrorInfo())
```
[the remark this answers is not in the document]
levzem: its protected - the only exception that sets the stack trace to not writable is `AbortException`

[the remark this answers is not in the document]
levzem: I found a way to supress the stack trace but that means creating a new exception and overriding the stack trace

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -594,4 +617,16 @@ public String toString() {
       return name().toLowerCase(Locale.ROOT);
     }
   }
+
+  public static class NoStackTraceConnectException extends RuntimeException {
```
wicknicks, line 621: maybe we can call this class `ReportingException`?. the exception doesn't have anything do with `ConnectException`.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -594,4 +617,16 @@ public String toString() {
...
+
+    public NoStackTraceConnectException(String message) {
+      super(message);
+    }
+
+    @Override
+    public synchronized Throwable fillInStackTrace() {
+      return this;
```
wicknicks, line 629: maybe add a javadoc explaining what we are doing here?

PR #428, opened 2020-07-20
Check for unexpected datatypes in response from server

cyrusv:
## Problem
Sometimes the response from the ES server does not have the "nodes" element we expect to parse, and we get an NPE when we should be returning default version. 

## Solution
I added some more look-before-you-leap logic to log a warning and return default.

Alternative proposal: Wrap the whole function in blanket try-catch and return default version. I considered this to just put this to rest once and for all, but decided this would be faster and less controversial.

## Release Plan
Will target 5.1.x which is the earliest one that has logic kind of like this.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
@@ -178,6 +178,11 @@ public class JestElasticsearchClient implements ElasticsearchClient {
       LOG.warn("Couldn't get Elasticsearch version, result is null");
       return defaultVersion;
     }
+    if (!result.has("nodes")) {
+      LOG.error("Couldn't get Elasticsearch version (result is has no nodes); "
+              + "assuming {}, result: {}", defaultVersion, result);
+      return defaultVersion;
+    }
 
     checkForError(result);
 
--- a/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
@@ -179,8 +179,7 @@ public class JestElasticsearchClient implements ElasticsearchClient {
       return defaultVersion;
     }
     if (!result.has("nodes")) {
-      LOG.error("Couldn't get Elasticsearch version (result is has no nodes); "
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -178,6 +178,11 @@ private Version getServerVersion() throws IOException {
       LOG.warn("Couldn't get Elasticsearch version, result is null");
       return defaultVersion;
     }
+    if (!result.has("nodes")) {
+      LOG.error("Couldn't get Elasticsearch version (result is has no nodes); "
+              + "assuming {}, result: {}", defaultVersion, result);
```
levzem, line 183:
```suggestion
      LOG.warn("Couldn't get Elasticsearch version from result {} (result has no nodes). Assuming {}.", result, defaultVersion);
```
nit on log level and message

mageshn:
Few questions

1. What do we do with the version that is returned? Is it critical info for the connector to function?
2. Is there a mechanism to turn off this feature in ES or do we know if there are any specific version after which this is available?

cyrusv:
@mageshn 
1. There are some incompatibilities between versions of ES (these are major version changes), which we check specifically. It was initially added in `64bb4402f80c9eb4` to support changes in query data schema mappings.  Elasticsearch has very aggressive EOL schedule, so by this October they only support 7.0+, so in principle we can just get rid of this function and only support 7.0+ https://www.elastic.co/support/eol
2. "turn off this feature in ES", not sure which feature you are referring to?

cyrusv:
Basically you can grep the code for `ES_` to see where this is used -- it is Schema Mappings and Query Builders.  If we just want to support 6+ we could probably remove these version checks, and certainly with 7+, if they don't introduce new incompatibilities with v8.

mageshn:
@cyrusv For # 2 I was specifically interested in understanding under what circumstances the response will be `null`. If the response is null, can we simply put a message stating that the response from the <URI> was null and is required for the connector to function?

cyrusv:
@mageshn I'm not sure when it is null, but I'm not sure it is relevant in the context of this function -- My understanding is that the function should fall back to default version on failure, so a response for `getVersion()` is never required. If there is some connection requirement that is not filled, it would be there responsibility of another method to determine that. Does that make sense?

PR #430, opened 2020-07-25
fix reporter to catch NoSuchMethodFoundError as well

levzem:

## Problem
the reporter will fail with `NoSuchMethodFoundError` on old Connect workers. 

## Solution
add `NoSuchMethodFoundError` to the try catch
https://cwiki.apache.org/confluence/display/KAFKA/KIP-610%3A+Error+Reporting+in+Sink+Connectors

- [ ] yes
- [x] no

## Test Strategy
unfortunately I do not see a way that this can be tested. If anyone has suggestions, I will gladly implement. 

edit: I have run a manual test using CP 5.5.0 which runs on AK < 2.6 and the connector runs

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
`master` where this feature was first implemented


levzem:
I ran a manual test and updated the PR description to cover that

PR #435, opened 2020-08-18
CC-9455: address SslFactor backwards/forwards compatibility

levzem:

## Problem
the connector used non-public AK APIs to handle SSL configuration. AK 2.6 removed the non-public APIs and introduced a public one. The consequence is that the connector is not forwards compatible with newer AK versions.

## Solution
Use reflection to determine which AK version is on the Connect worker and use the appropriate API for that version

- [x] yes
- [ ] no

##### If yes, where?
RabbitMQ connector 

## Test Strategy
manually verified forward and backward compatibility by configuring SSL for CP 5.2.5-SNAPSHOT (AK 2.2), CP 5.5.0 (AK 2.5) and CP 6.0.0-SNAPSHOT (AK 2.6) to hit all 3 apis and logical branches.

verified that integration test still passes

##### Testing done:
- [ ] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
backport to `5.2.x` where SSL feature was first introduced

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
@@ -194,27 +194,38 @@ public class JestElasticsearchClient implements ElasticsearchClient {
     SslFactory kafkaSslFactory = new SslFactory(Mode.CLIENT, null, false);
     kafkaSslFactory.configure(config.sslConfigs());
 
-    Object ss;
-    try {
-      // try pre AK 2.6 version first
-      ss = SslFactory.class.getDeclaredMethod("sslEngineBuilder").invoke(kafkaSslFactory);
-      log.trace("Using AK 2.6+ SslFactory.");
-    } catch (Exception e) {
-      // must be running AK 2.6+
-      log.trace("AK 2.6+ SslFactory not found. Trying AK 2.6<.");
-      try {
-        ss = SslFactory.class.getDeclaredMethod("sslEngineFactory").invoke(kafkaSslFactory);
-        log.trace("Using AK 2.6< SslFactory.");
-      } catch (Exception ex) {
-          throw new ConnectException("Failed to find methods for SslFactory.", ex);
-      }
-    }
-
     SSLContext sslContext;
     try {
-      sslContext = (SSLContext) ss.getClass().getDeclaredMethod("sslContext").invoke(ss);
+      // try AK <= 2.2 first
+      sslContext =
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -193,7 +193,40 @@ private static void configureSslContext(HttpClientConfig.Builder builder,
...
     SslFactory kafkaSslFactory = new SslFactory(Mode.CLIENT, null, false);
     kafkaSslFactory.configure(config.sslConfigs());
-    SSLContext sslContext = kafkaSslFactory.sslContext();
+
+    SSLContext sslContext;
+    try {
+      // try AK <= 2.2 first
+      sslContext =
+          (SSLContext) SslFactory.class.getDeclaredMethod("sslContext").invoke(kafkaSslFactory);
+      log.trace("Using AK 2.2< SslFactory.");
```
wicknicks, line 202: since this method is called once during initialization, I think all of these can be debug statements.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -193,7 +193,40 @@ private static void configureSslContext(HttpClientConfig.Builder builder,
...
+    } catch (Exception e) {
+      // must be running AK 2.3+
+      log.trace("Could not find SslFactory::sslContext. Trying AK 2.3+ APIs.");
+
+      Object ss;
```
wicknicks, line 207: maybe (for better readability): 

```suggestion
      Object sslEngine;
```

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -193,7 +193,40 @@ private static void configureSslContext(HttpClientConfig.Builder builder,
...
+      try {
+        // try AK <= 2.6 second
+        ss = SslFactory.class.getDeclaredMethod("sslEngineBuilder").invoke(kafkaSslFactory);
+        log.trace("Using AK 2.6< SslFactory.");
+
+      } catch (Exception ex) {
+        // must be running AK 2.6+
+        log.trace("AK 2.6< SslFactory not found. Trying AK 2.6+.");
```
wicknicks, line 215: we should word such errors so that users not familiar with the internals can understand what is going on.

```suggestion
        log.trace("SslFactory not found. Switching to SslEngine libs from AK 2.6 and beyond.");
```

might want to take another pass at the other trace statements.

PR #442, opened 2020-09-25
HCD-448: Adding support for size-based batching of sink records.

sagar-ab-2702:
The feature request for adding support of size-based batching of sink records in ElasticSearch sink connector.

The batching of records on the basis of size is now configurable using ``max.batch.size.bytes`` configuration.
With this implementation number of records in batches will depend upon the configured records ``batch.size`` and ``max.batch.size.bytes``.
Jira: https://confluentinc.atlassian.net/browse/HCD-448


levzem:
this will be easier to implement after #468 so holding review until then

levzem:
implemented in #512

Issue #444, opened 2020-09-28
Out of memory detected with kafka-connect-elasticsearch 10.0.0


pkgonan:
Hi, memory leak is detected at kafka-connect-elasticsearch version 10.0.0.

Before 18:33 - Deployment version : kafka-connect-elasticsearch 10.0.0
After 18:33 - Deployment version : kafka-connect-elasticsearch 5.5.1

![Old Gen](https://user-images.githubusercontent.com/15858772/94417891-a1343f00-01bb-11eb-9c7f-feb90b4ee055.png)

[Before kafka-connect-elasticsearch 10.0.0 - Dockerfile]
```
FROM confluentinc/cp-kafka-connect-base:6.0.0

ENV CONNECT_PLUGIN_PATH="/usr/share/java,/usr/share/confluent-hub-components" \
    CUSTOM_SMT_PATH="/usr/share/confluent-hub-components/custom-smt"


ARG DEBEZIUM_VERSION=1.2.5
ARG CONNECT_TRANSFORM_VERSION=1.3.2
ARG CONNECT_ELASTICSEARCH_VERSION=10.0.0


# Download Using confluent-hub
RUN confluent-hub install --no-prompt confluentinc/connect-transforms:$CONNECT_TRANSFORM_VERSION \
 && confluent-hub install --no-prompt confluentinc/kafka-connect-elasticsearch:$CONNECT_ELASTICSEARCH_VERSION


# Download Custom SMT
RUN mkdir $CUSTOM_SMT_PATH && cd $CUSTOM_SMT_PATH && \
    curl -sO https://repo1.maven.org/maven2/io/debezium/debezium-core/$DEBEZIUM_VERSION.Final/debezium-core-$DEBEZIUM_VERSION.Final.jar
```

[After kafka-connect-elasticsearch 5.5.1 - Dockerfile]
```
FROM confluentinc/cp-kafka-connect-base:6.0.0

ENV CONNECT_PLUGIN_PATH="/usr/share/java,/usr/share/confluent-hub-components" \
    CUSTOM_SMT_PATH="/usr/share/confluent-hub-components/custom-smt"


ARG DEBEZIUM_VERSION=1.2.5
ARG CONNECT_TRANSFORM_VERSION=1.3.2
ARG CONNECT_ELASTICSEARCH_VERSION=5.5.1


# Download Using confluent-hub
[...]

PR #446, opened 2020-09-28
CC-11959: BulkProcessor does not empty internal record map correctly

levzem:

## Problem
the connector has to keep an internal map (`recordMap`) to keep track of the original `SinkRecords` so that it can report any errors using the DLQ reporter. the connector has been experiencing a memory leak first noticed in #444. On further analysis it was found that the connector did not empty the map on successful writes (only on [failed writes](https://github.com/confluentinc/kafka-connect-elasticsearch/blob/master/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java#L478)).

## Solution
empty the map correctly in all cases

- [ ] yes
- [x] no

## Test Strategy
modified all existing unit tests to verify that the map is indeed empty

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport to `10.0.x` where the DLQ reporter was first introduced

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
@@ -77,7 +77,7 @@ public class BulkProcessor<R, B> {
   // shared state, synchronized on (this), may be part of wait() conditions so need notifyAll() on
   // changes
   private final Deque<R> unsentRecords;
-  protected final ConcurrentHashMap<R, SinkRecord> recordMap; // visible for tests
+  protected final ConcurrentHashMap<R, SinkRecord> recordsToReportOnError; // visible for tests
   private int inFlightRecords = 0;
   private final LogContext logContext = new LogContext();
 
@@ -104,7 +104,9 @@ public class BulkProcessor<R, B> {
     this.reporter = reporter;
 
     unsentRecords = new ArrayDeque<>(maxBufferedRecords);
-    recordMap = new ConcurrentHashMap<>(maxBufferedRecords);
+    recordsToReportOnError = reporter != null
+        ? new ConcurrentHashMap<>(maxBufferedRecords)
+        : new ConcurrentHashMap<>();
 
     final ThreadFactory threadFactory = makeThreadFactory();
     farmer = threadFactory.newThread(farmerTask());
@@ -344,7 +346,10 @@ public class BulkProcessor<R, B> {
     }
 
     unsentRecords.addLast(record);
-    recordMap.put(record, original);
+    if (reporter != null) {
+      // avoid unnecessary operations if not using the reporter
+      recordsToReportOnError.put(record, original);
+    }
     notifyAll();
   }
 
@@ -438,7 +443,7 @@ public class BulkProcessor<R, B> {
             batch.size(),
             e
         );
-        recordMap.keySet().removeAll(batch);
+        recordsToReportOnError.keySet().removeAll(batch);
         throw e;
       }
       final int maxAttempts = maxRetries + 1;
@@ -459,14 +464,14 @@ public class BulkProcessor<R, B> {
                   System.currentTimeMillis() - startTime
               );
             }
-            recordMap.keySet().removeAll(batch);
+            recordsToReportOnError.keySet().removeAll(batch);
             return bulkRsp;
           } else if (responseContainsMalformedDocError(bulkRsp)) {
             retriable = bulkRsp.isRetriable();
             handleMalformedDoc(bulkRsp);
             if (reporter != null) {
               for (R record : batch) {
-                SinkRecord original = recordMap.get(record);
+                SinkRecord original = recordsToReportOnError.get(record);
                 if (original != null) {
                   reporter.report(
                       original,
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -77,7 +77,7 @@
   // shared state, synchronized on (this), may be part of wait() conditions so need notifyAll() on
   // changes
   private final Deque<R> unsentRecords;
-  private final ConcurrentHashMap<R, SinkRecord> recordMap;
+  protected final ConcurrentHashMap<R, SinkRecord> recordMap; // visible for tests
```
kkonstantine, line 80: can we change the name of this variable to something more intuitive?
Also, let's not bake the type in the variable name. 
`recordsToReportOnError` ?

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -623,6 +627,7 @@ public String toString() {
    * (mapper_parser_exception, illegal_argument_exception, and action_request_validation_exception)
    * resulting from bad records using the AK 2.6 reporter DLQ interface.
    */
+  @SuppressWarnings("serial")
```
kkonstantine, line 630: For my information, why did we choose to suppress the stacktrace here, instead of wrap the exception into a new exception, or even report the exception itself?

kkonstantine (commented):
Thanks for the fix @levzem 
Looks good overall, I have a few minor comments that hopefully will make the coder easier to follow.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
levzem: https://github.com/confluentinc/kafka-connect-elasticsearch/pull/421#discussion_r444408464

kkonstantine:
One more thing. Is it correct to say that the task will add and remove records from this hashmap, even if the error reporter is not configured (which if I remember correct, is possible even on AK versions >= 2.6)

levzem:
> One more thing. Is it correct to say that the task will add and remove records from this hashmap, even if the error reporter is not configured (which if I remember correct, is possible even on AK versions >= 2.6)

that is correct. I was thinking of gating the adding of records by whether the reporter is configured or not as a small optimization

kkonstantine:
> > One more thing. Is it correct to say that the task will add and remove records from this hashmap, even if the error reporter is not configured (which if I remember correct, is possible even on AK versions >= 2.6)
> 
> that is correct. I was thinking of gating the adding of records by whether the reporter is configured or not as a small optimization

yeah, let's do that. Optimization and can also allow for workarounds when a bug like this one comes up.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -104,7 +104,9 @@ public BulkProcessor(
     this.reporter = reporter;
 
     unsentRecords = new ArrayDeque<>(maxBufferedRecords);
-    recordMap = new ConcurrentHashMap<>(maxBufferedRecords);
+    recordsToReportOnError = reporter != null
+        ? new ConcurrentHashMap<>(maxBufferedRecords)
+        : new ConcurrentHashMap<>();
```
kkonstantine, line 109:
```suggestion
        : null
```

We can use `Collections.emptyMap()` if we declare the interface above to me `Map` instead of `ConcurrentMap`. But since we are already gating on `null` all the calls to this map, maybe we can go with `null` here to avoid creating a new map with every object when the error reporter is not used.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -344,7 +346,10 @@ public synchronized void add(R record, SinkRecord original, long timeoutMs) {
     }
 
     unsentRecords.addLast(record);
-    recordMap.put(record, original);
+    if (reporter != null) {
+      // avoid unnecessary operations if not using the reporter
+      recordsToReportOnError.put(record, original);
```
kkonstantine, line 351: Gating is required in every call. I suggest fleshing it out in a helper method. 
Example: 

```
  private void addRecordsToReport(R record, SinkRecord original) {
    if (reporter != null) {
      recordsToReportOnError.put(record, original);
    }
  }
```

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -438,6 +443,7 @@ private BulkResponse execute() throws Exception {
             batch.size(),
             e
         );
+        recordsToReportOnError.keySet().removeAll(batch);
```
kkonstantine, line 446: To keep the symmetry with the null check in `put` I'd suggest adding another helper method in these calls.
Example: 

```
 private void removeReportedRecords(List<R> batch) {
    if (reporter != null) {
      recordsToReportOnError.keySet().removeAll(batch);
    }
  }
```

kkonstantine (commented):
Thanks @levzem !
Looks good overall. 
I think we can take it one step further and avoid to create a new map with every object if we don't need to.

kkonstantine (approved):
LGTM 
Thanks for the quick fix @levzem !

PR #447, opened 2020-09-29
CC-11969: do not report successful records in failed batches

levzem:
## Problem
the reporter currently reports all records in a batch as failed if even 1 record failed due to the connector design. This is incorrect and produces a lot of unnecessary logs as well.

## Solution
refactor the reporter to only report failed records in failed batches

- [ ] yes
- [x] no

## Test Strategy
write unit test to make sure only failed records are written

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport to `10.0.x` when reporter was first introduced

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
@@ -19,6 +19,8 @@ import io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig;
 import io.confluent.connect.elasticsearch.LogContext;
 import io.confluent.connect.elasticsearch.RetryUtil;
 
+import io.searchbox.core.BulkResult.BulkResultItem;
+import java.util.Collections;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import org.apache.kafka.common.config.ConfigDef;
@@ -184,7 +186,12 @@ public class BulkProcessor<R, B> {
     // at this point, either stopRequested or canSubmit
     return stopRequested
             ? CompletableFuture.completedFuture(
-                    BulkResponse.failure(false, "request not submitted during shutdown"))
+                    BulkResponse.failure(
+                        false,
+                        "request not submitted during shutdown",
+                        Collections.emptyMap()
+                    )
+            )
             : submitBatch();
   }
 
@@ -484,12 +491,11 @@ public class BulkProcessor<R, B> {
             if (reporter != null) {
               for (R record : batch) {
                 SinkRecord original = recordsToReportOnError.get(record);
-                if (original != null) {
+                BulkResultItem result = bulkRsp.failedRecords.get(record);
+                String error = result != null ? result.error : null;
+                if (original != null && error != null) {
                   reporter.report(
-                      original,
-                      new ReportingException(
-                          "Bulk request failed: " + bulkRsp.getErrorInfo()
-                      )
+                      original, new ReportingException("Bulk request failed: " + error)
                   );
                 }
               }
--- a/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
@@ -493,7 +493,7 @@ public class BulkProcessor<R, B> {
                 SinkRecord original = recordsToReportOnError.get(record);
                 BulkResultItem result = bulkRsp.failedRecords.get(record);
                 String error = result != null ? result.error : null;
-                if (original != null && error != null) {
+                if (error != null && original != null) {
                   reporter.report(
                       original, new ReportingException("Bulk request failed: " + error)
                   );
--- a/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkResponse.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/bulk/BulkResponse.java
@@ -15,26 +15,43 @@
 
 package io.confluent.connect.elasticsearch.bulk;
 
+import io.confluent.connect.elasticsearch.IndexableRecord;
+import io.searchbox.core.BulkResult.BulkResultItem;
+import java.util.Collections;
+import java.util.Map;
+
 public class BulkResponse {
 
-  private static final BulkResponse SUCCESS_RESPONSE = new BulkResponse(true, false, "");
+  private static final BulkResponse SUCCESS_RESPONSE =
+      new BulkResponse(true, false, "", Collections.emptyMap());
 
   public final boolean succeeded;
   public final boolean retriable;
   public final String errorInfo;
+  public final Map<IndexableRecord, BulkResultItem> failedRecords;
 
-  private BulkResponse(boolean succeeded, boolean retriable, String errorInfo) {
+  private BulkResponse(
+      boolean succeeded,
+      boolean retriable,
+      String errorInfo,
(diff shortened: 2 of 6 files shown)


gharris1727 (approved):
LGTM!

kkonstantine (commented):
Overall makes sense. 
I'd suggest merging https://github.com/confluentinc/kafka-connect-elasticsearch/pull/446 first to help with the review here.

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -484,12 +491,11 @@ private BulkResponse execute() throws Exception {
             if (reporter != null) {
               for (R record : batch) {
                 SinkRecord original = recordsToReportOnError.get(record);
-                if (original != null) {
+                BulkResultItem result = bulkRsp.failedRecords.get(record);
```
kkonstantine, line 494: can we add a javadoc in `BulkResponse` constructor that says that this can't be `null`(never `null`) but can be empty?

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -484,12 +491,11 @@ private BulkResponse execute() throws Exception {
...
+                String error = result != null ? result.error : null;
+                if (original != null && error != null) {
```
kkonstantine, line 496: assuming that in the records reporter you have added all the batch and only a few records have errored out, let's first check error
```suggestion
                if (error != null && original != null) {
```

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkResponse.java
```diff
@@ -15,26 +15,43 @@
...
+    this.failedRecords = failedRecords;
   }
 
   public static BulkResponse success() {
     return SUCCESS_RESPONSE;
   }
 
-  public static BulkResponse failure(boolean retriable, String errorInfo) {
-    return new BulkResponse(false, retriable, errorInfo);
+  public static BulkResponse failure(
```
kkonstantine, line 49: please add javadoc saying including all the args, but specifically saying that `failedRecords` is never `null`

src/main/java/io/confluent/connect/elasticsearch/bulk/BulkResponse.java
```diff
@@ -15,26 +15,43 @@
...
+  private static final BulkResponse SUCCESS_RESPONSE =
+      new BulkResponse(true, false, "", Collections.emptyMap());
 
   public final boolean succeeded;
   public final boolean retriable;
   public final String errorInfo;
+  public final Map<IndexableRecord, BulkResultItem> failedRecords;
 
-  private BulkResponse(boolean succeeded, boolean retriable, String errorInfo) {
+  private BulkResponse(
```
kkonstantine, line 33: worth adding the javadoc here too, even if it's private

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -650,7 +652,26 @@ public BulkResponse executeBulk(BulkRequest bulk) throws IOException {
 
     final String errorInfo = errors.isEmpty() ? result.getErrorMessage() : errors.toString();
 
-    return BulkResponse.failure(retriable, errorInfo);
+    Map<IndexableRecord, BulkResultItem> failedResponses = new HashMap<>();
+    List<BulkResultItem> items = result.getItems();
+    for (int i = 0; i < items.size() && i < ((JestBulkRequest) bulk).records().size() ; i++) {
```
kkonstantine, line 657: now that `bulk` gets used in several places, maybe consider a local var at the top and casting once. 
Eg.: `JestBulkRequest jestBulk = (JestBulkRequest) bulk;` (or `jestBulkRequest` I guess)

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -650,7 +652,26 @@ public BulkResponse executeBulk(BulkRequest bulk) throws IOException {
 
     final String errorInfo = errors.isEmpty() ? result.getErrorMessage() : errors.toString();
 
-    return BulkResponse.failure(retriable, errorInfo);
+    Map<IndexableRecord, BulkResultItem> failedResponses = new HashMap<>();
+    List<BulkResultItem> items = result.getItems();
```
kkonstantine, line 656: I'd also suggest declaring and using below (example names): 
```
List<IndexableRecord> bulkRecords = jestBulk.records();
int bulkRecordsNum = bulkRecords.size();
int itemsNum = items.size();
```
probably JIT-able optimizations but since not bad to have and maybe slightly easier to follow.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -650,7 +652,26 @@ public BulkResponse executeBulk(BulkRequest bulk) throws IOException {
...
+    for (int i = 0; i < items.size() && i < ((JestBulkRequest) bulk).records().size() ; i++) {
+      BulkResultItem item = items.get(i);
+      IndexableRecord record = ((JestBulkRequest) bulk).records().get(i);
+      if (item.error != null && item.id.equals(record.key.id)) {
```
kkonstantine, line 660: I see in the constructor of `BulkResultItem` that `id` can be set to `null`. 
Are we risking NPE here?
Seems `record.key.id` can't be `null`. `Objects.equals` is an option that works with `null`

src/test/java/io/confluent/connect/elasticsearch/bulk/BulkProcessorTest.java
```diff
@@ -62,15 +75,15 @@ private Expectation(List<Integer> request, BulkResponse response) {
     }
   }
 
-  private static final class Client implements BulkClient<Integer, List<Integer>> {
+  private static final class Client implements BulkClient<IndexableRecord, List<Integer>> {
```
kkonstantine, line 78: so that change was missed before here?

kkonstantine (commented):
Great to see this fixed too. 

Left a few comments, but we seem close

kkonstantine:
Also, I added a few comments around potential NPEs, but I also see that two tests fail on NPEs already. Not sure it's for the same reason.

src/test/java/io/confluent/connect/elasticsearch/bulk/BulkProcessorTest.java
levzem: i think the original author just didn't want to set up `IndexableRecords` 🤷

sonupillai:
I have a general question given we have seen an incident because of this. Where do we log the errant records which are meant to be in the DLQ? IMO they should not be part of error logs and should be in info logs as long as they are already classified as know errant records which are being written to the DLQ topic. Error logs should only capture real errors.

levzem:
> I have a general question given we have seen an incident because of this. Where do we log the errant records which are meant to be in the DLQ? IMO they should not be part of error logs and should be in info logs as long as they are already classified as know errant records which are being written to the DLQ topic. Error logs should only capture real errors.

the logs were being generated inside the reporter itself, but because we were reporting the whole batch of records it was generating a lot of logs, and now this change only reports the relevant records in a batch.

kkonstantine (approved):
LGTM

PR #449, opened 2020-10-07
CC-12129: catch resource_already_exists_exception when creating indices

levzem:

## Problem
index exists can return a 404 Not Found when it exists

## Solution
check that the reason index creation failed was because it already exists and only then check if the index exists

- [ ] yes
- [x] no

## Test Strategy
add unit test to mock a failed index creation

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport where this was first introduced

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
@@ -85,6 +85,8 @@ public class JestElasticsearchClient implements ElasticsearchClient {
       = "version_conflict_engine_exception";
   protected static final String ALL_FIELD_PARAM
       = "_all";
+  protected static final String RESOURCE_ALREADY_EXISTS_EXCEPTION
+      = "resource_already_exists_exception";
 
   private static final Logger LOG = LoggerFactory.getLogger(JestElasticsearchClient.class);
 
@@ -401,8 +403,11 @@ public class JestElasticsearchClient implements ElasticsearchClient {
       JestResult result = client.execute(createIndex);
       log.debug("Received response for request to create index '{}'", index);
       if (!result.isSucceeded()) {
+        boolean exists = result.getErrorMessage().contains(RESOURCE_ALREADY_EXISTS_EXCEPTION)
+            || indexExists(index);
+
         // Check if index was created by another client
-        if (!indexExists(index)) {
+        if (!exists) {
           String msg =
               result.getErrorMessage() != null ? ": " + result.getErrorMessage() : "";
           throw new ConnectException("Could not create index '" + index + "'" + msg);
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -402,8 +404,11 @@ private void createIndex(String index, CreateIndex createIndex) throws ConnectEx
...
       log.debug("Received response for request to create index '{}'", index);
       if (!result.isSucceeded()) {
+        boolean exists = result.getErrorMessage().contains(RESOURCE_ALREADY_EXISTS_EXCEPTION)
+            || indexExists(index);
+
         // Check if index was created by another client
-        if (!indexExists(index)) {
+        if (!exists) {
           String msg =
               result.getErrorMessage() != null ? ": " + result.getErrorMessage() : "";
```
Prathibha-m, line 413: Does this logic check if the index was created by another client (like the comment says)? It seems to just check if it existed, not if it was by a different client

Prathibha-m: Q: Wouldn't it make sense to check `indexExists(index)` before executing the action to create index rather than the other way around? Then after creation of index, if it still failed, we can add debug logs to understand the real root cause

levzem: https://github.com/confluentinc/kafka-connect-elasticsearch/pull/449/files#diff-784dd8ae0d081fd9ae1b2ecc191392a2L371

levzem: the logic flow is a bit confusing but we do check beforehand

src/test/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClientTest.java
```diff
@@ -367,7 +392,7 @@ public void closes() throws IOException {
   public void toBulkableAction(){
     JestElasticsearchClient client = new JestElasticsearchClient(jestClient);
     IndexableRecord del = new IndexableRecord(new Key("idx", "tp", "xxx"), null, 1L);
-    BulkableAction ba = client.toBulkableAction(del);
+    BulkableAction<DocumentResult> ba = client.toBulkableAction(del);
```
Prathibha-m, line 395: nit: Better variable name

Prathibha-m (commented):
Were we able to reproduce the issue through an integration test?

Prathibha-m:
Q:Could this have happened because all shards on the resource were not ready yet? https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#index-wait-for-active-shards  I see a parameter `wait_for_active_shards` on the index API and wondering if it's in any way relevant

ilanjiR (approved):
LGTM, lets make sure IT are passing

PR #452, opened 2020-10-15
Support retry_on_conflict parameter Fixes #374

### LINKED ISSUE #374: Ignoring version conflicts for items causes loss of data

dosvath:
## Problem
#374 

## Solution
Add a parameter `retry.on.conflict` and include it in the request to ES. 

- [ ] yes
- [X] no

##### Testing done:
- [X] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
New config to master.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
@@ -76,6 +76,7 @@ import java.util.stream.Collectors;
 import javax.net.ssl.SSLContext;
 
 import static org.apache.kafka.connect.runtime.ConnectorConfig.TASKS_MAX_CONFIG;
+import static org.apache.kafka.connect.runtime.ConnectorConfig.TASKS_MAX_DEFAULT;
 
 public class JestElasticsearchClient implements ElasticsearchClient {
   private static final Logger log = LoggerFactory.getLogger(JestElasticsearchClient.class);
@@ -157,7 +158,8 @@ public class JestElasticsearchClient implements ElasticsearchClient {
       this.retryBackoffMs = config.retryBackoffMs();
       this.maxRetries = config.maxRetries();
       this.timeout = config.readTimeoutMs();
-      this.retryOnConflict = Integer.parseInt(props.get(TASKS_MAX_CONFIG));
+      String tasksMax = props.get(TASKS_MAX_CONFIG);
+      this.retryOnConflict = (tasksMax == null) ? TASKS_MAX_DEFAULT : Integer.parseInt(tasksMax);
 
     } catch (IOException e) {
       throw new ConnectException(
--- a/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
@@ -75,9 +75,6 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import javax.net.ssl.SSLContext;
 
-import static org.apache.kafka.connect.runtime.ConnectorConfig.TASKS_MAX_CONFIG;
-import static org.apache.kafka.connect.runtime.ConnectorConfig.TASKS_MAX_DEFAULT;
-
 public class JestElasticsearchClient implements ElasticsearchClient {
   private static final Logger log = LoggerFactory.getLogger(JestElasticsearchClient.class);
 
@@ -157,9 +154,8 @@ public class JestElasticsearchClient implements ElasticsearchClient {
       this.writeMethod = config.writeMethod();
       this.retryBackoffMs = config.retryBackoffMs();
       this.maxRetries = config.maxRetries();
-      this.timeout = config.readTimeoutMs();
-      String tasksMax = props.get(TASKS_MAX_CONFIG);
-      this.retryOnConflict = (tasksMax == null) ? TASKS_MAX_DEFAULT : Integer.parseInt(tasksMax);
+      this.timeout = config.readTimeoutMs();;
+      this.retryOnConflict = config.maxInFlightRequests();
 
     } catch (IOException e) {
       throw new ConnectException(
--- a/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
@@ -154,7 +154,7 @@ public class JestElasticsearchClient implements ElasticsearchClient {
       this.writeMethod = config.writeMethod();
       this.retryBackoffMs = config.retryBackoffMs();
       this.maxRetries = config.maxRetries();
-      this.timeout = config.readTimeoutMs();;
+      this.timeout = config.readTimeoutMs();
       this.retryOnConflict = config.maxInFlightRequests();
 
     } catch (IOException e) {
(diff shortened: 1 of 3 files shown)


dosvath:
Taking over https://github.com/confluentinc/kafka-connect-elasticsearch/pull/384 in this PR as other branch had no push permissions.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -257,6 +257,13 @@
       + ", and decrease " + BATCH_SIZE_CONFIG + " configuration properties.";
   private static final String WRITE_METHOD_DISPLAY = "Write Method";
 
+  public static final String RETRY_ON_CONFLICT_CONFIG = "retry.on.conflict";
+  private static final String RETRY_ON_CONFLICT_DISPLAY = "Retry on conflict";
+  private static final String RETRY_ON_CONFLICT_DOC = "Specify how many times the operation should"
+      + " be retried by Elasticsearch, when write conflicts occur, while using the write method "
+      + WriteMethod.UPSERT.toString() + ". The value is the number of retries.";
+  private static final int RETRY_ON_CONFLICT_DEFAULT = 0;
```
Prathibha-m, line 265: Was there any reason why this default was chosen? 

I found this q&a https://discuss.elastic.co/t/whats-appropriate-value-at-retry-on-conflict/54312 which says 

> If you need parallel indexing of similar documents, what are the worst case outcomes. 

So I'm wondering if this means an appropriate default would be proportional to the number of tasks that are set

Prathibha-m: Since each task is a thread, I think parallel indexing would be of the order of numOfTasks (maybe there is more detail as to how parallel indexing happens within the connector and how it is split) 

While we can add a new config to make this configurable going forward, we might be able to backport a fix to set retry_on_conflict(numOfTasks) without the use of an extra config. WDYT? 

Can we run a manual test/IT with parallel update requests to see if that is a possible solution?

dosvath: That approach would work, for `numOfTasks` the property we have available is `tasks.max`, is that the prop you had in mind?

Prathibha-m: Yes, that's right 👍

dosvath: @levzem what do you think about the above approach? Do you think users would need their own config of this value based on the [raised issue](https://github.com/confluentinc/kafka-connect-elasticsearch/issues/374)? Should we go for the backported default value in this PR and issue a new config later if requested?

levzem: ideally we shouldn't even be producing write conflicts and not have to introduce this config. otherwise, I would like to implement this without the config at all as I don't think this adds value to the customer and adds unnecessary complexity. the Jest client rework will remove the need for this config, so let's not add it in the first place

dosvath:
Updated PR to use `tasks.max` for retries.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -154,6 +158,9 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
       this.retryBackoffMs = config.retryBackoffMs();
       this.maxRetries = config.maxRetries();
       this.timeout = config.readTimeoutMs();
+      String tasksMax = props.get(TASKS_MAX_CONFIG);
+      this.retryOnConflict = (tasksMax == null) ? TASKS_MAX_DEFAULT : Integer.parseInt(tasksMax);
```
levzem, line 162: if we are trying to retry the same amount of times that there are concurrent requests, the safer number to pick here is max concurrent requests instead plus it is much cleaner

dosvath: Do you mean the `MAX_IN_FLIGHT_REQUESTS_CONFIG` ?

dosvath:
Update: Manually verified the connector writes correctly with this change.

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -154,6 +155,8 @@ protected JestElasticsearchClient(Map<String, String> props, JestClientFactory f
       this.retryBackoffMs = config.retryBackoffMs();
       this.maxRetries = config.maxRetries();
       this.timeout = config.readTimeoutMs();
+      this.retryOnConflict = config.maxInFlightRequests();
+
```
kkonstantine, line 159: extra blank

kkonstantine (approved):
LGTM

src/main/java/io/confluent/connect/elasticsearch/jest/JestElasticsearchClient.java
```diff
@@ -102,6 +102,7 @@
   private int maxRetries;
   private long retryBackoffMs;
   private final Time time = new SystemTime();
+  private int retryOnConflict;
```
kkonstantine, line 105: can be `final` I suppose

dosvath: final would require us to initialize it in the other 2-3 constructors :/

kkonstantine: Adding the statements should not be an issue. The variable is initialized implicitly as `false` but it's not final. Member fields that are immutable outside the constructor should be `final`.

PR #454, opened 2020-10-26
CC-12447: add comprehensive ITs to ES

levzem:
## Problem
Elasticsearch is missing some basic ITs

## Solution
refactor ITs and add missing ITs

- [ ] yes
- [x] no

## Test Strategy
new ITs testing primitives, deletes, null values, and restarting the connector

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
`5.2.x` where all ITs were introduced


src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
```diff
@@ -0,0 +1,128 @@
...
+
+    // wait for tasks to spin up
+    waitForConnectorToStart(CONNECTOR_NAME, TASKS_MAX);
+
+    for (int i  = 0; i < 10; i++) {
+      connect.kafka().produce(TOPIC, String.valueOf(i),  String.valueOf(i));
+    }
+
+    waitForRecords(0);
+  }
```
Prathibha-m, line 127: Can we add a test for `write.method: UPSERT` as well? I believe we could've caught retry conflict errors (we saw some time ago) if we had a test for it

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchContainer.java
```diff
@@ -191,7 +192,7 @@ protected void configure() {
               .forStatusCodeMatching(status -> status == HTTP_OK || status == HTTP_UNAUTHORIZED)
               .withStartupTimeout(Duration.ofMinutes(2))
       );
-      image = new RemoteDockerImage(imageName);
+      image = new RemoteDockerImage(DockerImageName.parse(imageName));
```
Prathibha-m, line 195: Is moving from a TestContainer approach to using https://confluentinc.atlassian.net/wiki/spaces/CONNECT/pages/1549697235/Configure+Connectors+Repositories+to+run+integration-tests+on+Jenkins not in the scope of this PR?

It would involve some work to rewrite these tests to hit an actual ES instance, and I'm wondering why add more testing with the Container approach when there is a plan to migrate

Prathibha-m (commented):
Really like the organization of IT's 💯 
Had a few questions:

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchContainer.java
levzem: these ITs are being run in Jenkins right now, on top of that, TestContainers just brings up a docker image of Elasticsearch so these tests are actually running against actual ES instances and even lets us configure the ES instance to however we like programatically (i.e general ssl certs and test those) which we wouldnt be able to do with an "actual" ES instance easily

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
levzem: that config isnt added in this branch, but I will add an IT for upsert in a subsequent PR

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchContainer.java
Prathibha-m: Ah my bad. I thought TestContainers being run on Jenkins had some n/w problems

levzem: yeah testcontainers is actually a really nifty thing :)

Prathibha-m (approved):
LGTM pending green build

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchContainer.java
Prathibha-m: So are we running only SecurityIT per this comment? https://github.com/confluentinc/kafka-connect-elasticsearch/blob/66fcdf2594297ebafb5e84190d2d70e61db6f9a5/pom.xml#L334

levzem: hmm good catch, thats why my test is failing, I will have to disable it

PR #455, opened 2020-10-27
CC-12530: do not throw exception on flush after task close

levzem:

## Problem
the worker calls `SinkTask::close` followed by `SinkTask::flush`, but `SinkTask::close` closes the `BulkProcessor` which causes the consequent flush to throw an exception.

## Solution
only flush if the `BulkProcessor is still open

- [ ] yes
- [x] no

## Test Strategy
write a test that flushes after close

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport as far back as possible

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -35,7 +35,7 @@ import java.util.concurrent.TimeUnit;
 public class ElasticsearchSinkTask extends SinkTask {
 
   private static final Logger log = LoggerFactory.getLogger(ElasticsearchSinkTask.class);
-  private ElasticsearchWriter writer;
+  private volatile ElasticsearchWriter writer;
   private ElasticsearchClient client;
   private Boolean createIndicesAtStartTime;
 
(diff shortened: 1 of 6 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -124,8 +124,12 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
 
   @Override
   public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
-    log.trace("Flushing data to Elasticsearch with the following offsets: {}", offsets);
-    writer.flush();
+    if (writer != null) {
+      log.trace("Flushing data to Elasticsearch with the following offsets: {}", offsets);
+      writer.flush();
+    } else {
+      log.trace("Could not flush data to Elasticsearch because ESWriter already closed.");
```
kkonstantine, line 131: should this be debug?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -138,6 +142,7 @@ public void stop() throws ConnectException {
     log.info("Stopping ElasticsearchSinkTask.");
     if (writer != null) {
       writer.stop();
+      writer = null;
```
kkonstantine, line 145: if `flush` and `stop` are executed in two different threads (which I think they could) this won't necessarily be visible without the variable being `volatile`

kkonstantine (commented):
Thanks. Couple of comments.

kkonstantine (approved):
LGTM

PR #457, opened 2020-10-29
CC-12452: add config combination validators 

levzem:

## Problem
the connector does not validate config combinations at all
config combinations for ssl, proxy, and connectivity will come in later PRs

## Solution
add config combo validations

- [x] yes
- [ ] no

##### If yes, where?
other branches for configs added there i.e proxy and ssl

## Test Strategy
new unit tests that test both valid and invalid configs

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport to `5.0.x` and then add more validators in follow up PRs

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,7 +44,7 @@
         <hamcrest.version>1.3</hamcrest.version>
         <mockito.version>2.13.0</mockito.version>
         <jest.version>6.3.1</jest.version>
-        <test.containers.version>1.11.4</test.containers.version>
+        <test.containers.version>1.15.0-rc2</test.containers.version>
         <kafka.connect.maven.plugin.version>0.11.1</kafka.connect.maven.plugin.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
     </properties>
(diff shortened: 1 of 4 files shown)


Prathibha-m:
To clarify, we aren't adding all the validators(in place of existing error maps) to check the validity of credentials, URI as part of this PR, are we? Is that to happen as future work?

levzem:
> To clarify, we aren't adding all the validators to check the validity of credentials, URI as part of this PR, are we? Is that to happen as future work?

connectivity validators will be future PRs (as it depends on future work, we have tickets tracking this), but this lays down the foundation so that we can easily add them

pom.xml
```diff
@@ -44,7 +44,7 @@
         <hamcrest.version>1.3</hamcrest.version>
         <mockito.version>2.13.0</mockito.version>
         <jest.version>6.3.1</jest.version>
-        <test.containers.version>1.11.4</test.containers.version>
+        <test.containers.version>1.15.0-rc2</test.containers.version>
```
kkonstantine, line 47: is there a specific feature we are looking for in the 1.15 and we are getting the release candidate? Or the previous one is good enough?

levzem: the latest mac OS update broke test containers - `1.15.0-rc2` is the only version that contains this fix and was released by testcontainers for this reason

levzem: since Jenkins doesn't break, we don't necessarily need to bump it up but it would be nice to have working ITs for all OSes

kkonstantine: That's what I was wondering about. Sure, let's upgrade to the latest then

kkonstantine (approved):
LGTM!

PR #459, opened 2020-10-30
CC-12621: add config combo validator for ssl configs

levzem:

## Problem
the connector does not validate different combinations of SSL configs

## Solution
add validator to validate that SSL configs are only specified when elasticsearch.security.protocol is set to `SSL`

- [x] yes
- [ ] no

##### If yes, where?
proxy and connection configs need to be validated as well

## Test Strategy
add unit tests for both valid and invalid combinations for ssl configs

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport to `5.2.x` where the ssl configs were introduced


src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -72,7 +76,7 @@ private void validateCredentials() {
     boolean onlyOneSet = config.username() != null ^ config.password() != null;
```
Prathibha-m, line 76: We can use `StringUtils.isBlank` to ensure it's not empty whitespaces and null

Prathibha-m (approved):
One minor comment
Otherwise, LGTM!

PR #461, opened 2020-11-02
MINOR: fix tests

levzem:

## Problem
for some reason, specifying a parameter breaks Jenkins tests on `5.5.x` and `10.0.x` (green locally)

## Solution
remove the parameter for now

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport to `5.5.x` where the tests break


ilanjiR (approved):
LGTM, can we link a follow-up PR or jira ticket so we can track this somewhere? Seems like something worth investigating when we have more time later

levzem:
> LGTM, can we link a follow-up PR or jira ticket so we can track this somewhere? Seems like something worth investigating when we have more time later

this method and parameter is only used in tests and future work is going to replace this anyways, so I will hold off on that

PR #465, opened 2020-11-09
adding a null check when validating list of urls

BlueEldur:
## Problem
throwing NPE when validator receives a null list of ES endpoint url's

## Solution
add a null check and test to see if configexpcetion is thrown

- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -849,8 +849,13 @@ public WriteMethod writeMethod() {
   private static class UrlListValidator implements Validator {
 
     @Override
+    @SuppressWarnings("unchecked")
     public void ensureValid(String name, Object value) {
-      @SuppressWarnings("unchecked")
+      if (value == null) {
+        throw new ConfigException(
+            name, value, "The provided url list is null."
+        );
```
levzem, line 857: might fit on one line
```suggestion
        throw new ConfigException(name, value, "Url list must be provided and non-null.");
```

levzem (approved):
rebase to `5.0.x` but otherwise LGTM

PR #468, opened 2020-11-10
CC-12445: Replace the JestClient with the ElasticsearchRestClient

levzem:

## Problem
the jest client is outdated and needs to be replaced by the supported and maintained ES high level rest client

## Solution
replace the client

major points
- `BulkProcessor` is no longer required as ES provides its own
- all Jest classes are no longer necessary
- `ElasticsearchWriter` was absorbed into `SinkTask`
- retries are supported for all operations
- some configs are deprecated/not used anymore WIP, full list TBD)
    - `type.name` as it is deprecated as of ES 7.x 

- [ ] yes
- [x] no

## Test Strategy
an additional manual upgrade test was performed. write records to ES cloud with a `v10.0.2`, then upgrade to the new client and try writing to the same ES cloud index 

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
`master` because this is a major change


pom.xml
```diff
@@ -68,9 +67,22 @@
...
+            <groupId>org.elasticsearch</groupId>
+            <artifactId>elasticsearch</artifactId>
+            <version>${es.version}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.8.6</version>
+            <scope>${gson.version</scope>
```
kkonstantine, line 85: interesting that this doesn't fail the build. At least gives a big warning block. 

I'm suspecting this should have `test` scope now, since all the refs to gson have been removed from the production code.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Confluent Inc.
+ * Copyright 2020 Confluent Inc.
```
kkonstantine, line 2: we don't usually update those on existing files

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+  private static final String VERSION_CONFLICT_EXCEPTION = "version_conflict_engine_exception";
+  private static final Set<String> MALFORMED_DOC_ERRORS = new HashSet<>(
+      Arrays.asList(
+          "mapper_parsing_exception",
+          "illegal_argument_exception",
+          "action_request_validation_exception"
+      )
+  );
+
+  private AtomicReference<ConnectException> error;
```
kkonstantine, line 82: it'd be nice to make `final` all the class member fields that can be marked as such

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+  private ConcurrentHashMap<String, SinkRecord> recordMap;
+  private ElasticsearchSinkConnectorConfig config;
+  private ErrantRecordReporter reporter;
+  private RestHighLevelClient client;
+
+  public ElasticsearchClient(
+      ElasticsearchSinkConnectorConfig config,
+      ErrantRecordReporter reporter
+  ) {
+    int pos = 0;
```
kkonstantine, line 93: if this is too much boiler plate for you, you can also use `List#toArray` below (the typed version)

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+      Arrays.asList(
+          "mapper_parsing_exception",
+          "illegal_argument_exception",
+          "action_request_validation_exception"
+      )
+  );
+
+  private AtomicReference<ConnectException> error;
+  private BulkProcessor bulkProcessor;
+  private ConcurrentHashMap<String, SinkRecord> recordMap;
```
kkonstantine, line 84: Interfaces are preferable in general
```suggestion
  private ConcurrentMap<String, SinkRecord> recordMap;
```

kkonstantine: Also, usually best not to encode the type to the name if possible. This can also be `records` probably. 
Or `somethingToSomethingElse` is a common naming pattern if you want to be more specific. Otherwise, `records` is as informative as `recordMap`

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+   */
+  public void index(SinkRecord record, DocWriteRequest<?> request) {
+    if (isFailed()) {
+      close();
+      throw error.get();
+    }
+
+    if (recordMap != null) {
+      // make sure that only unique records are being sent in a batch
+      while (recordMap.contains(request.id())) {
```
kkonstantine, line 228: Seems you should be looking for the key, right?. `contains` looks for the value. It's also a legacy method. Using `ConcurrentMap` will help you avoid this issue.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+        .setConnectionRequestTimeout(config.readTimeoutMs());
+  }
+
+  /**
+   * Gets the mapping for an index.
+   *
+   * @param index the index to fetch the mapping for
+   * @return the MappingMetaData for the index
+   */
+  private MappingMetaData getMapping(String index) {
```
kkonstantine, line 443: we usually don't use `get/set` in connector-land when naming methods. 
nit: `mapping` since we are doing a general refactoring here and keeping the old name might not worth it.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+    GetMappingsResponse response = callWithRetries(
+        "get mapping for index " + index,
+        () -> client.indices().getMapping(request, RequestOptions.DEFAULT)
+    );
+    return response.mappings().get(index);
+  }
 
   /**
-   * Executes a search.
+   * Processes a response from a BulkAction. Successful responses are ignored. Failed responses are
```
kkonstantine, line 453: adding `@link`s in javadoc is nice to have when referring to actual classes

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+   */
+  @SuppressWarnings("serial")
+  public static class ReportingException extends RuntimeException {
+
+    public ReportingException(String message) {
+      super(message);
+    }
+
+    /**
+     * This method is overriden to swallow the stack trace.
```
kkonstantine, line 573:
```suggestion
     * This method is overridden to swallow the stack trace.
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+     *
+     * @return Throwable
+     */
+    @Override
+    public synchronized Throwable fillInStackTrace() {
```
kkonstantine, line 578: why is this `synchronized` ?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -21,8 +21,8 @@
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+import java.util.concurrent.TimeUnit;
```
kkonstantine, line 25: probably wrong reordering by accident.

kkonstantine (commented):
Thanks @levzem 
This is going to be a tremendous upgrade! 

Took a first quick pass over half the files and left a few comments. Will have to return soon for a more careful read.

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -152,51 +144,71 @@ public IndexableRecord convertRecord(
...
+    final String id = config.shouldIgnoreKey(record.topic())
+        ? String.format("%s+%d+%d", record.topic(), record.kafkaPartition(), record.kafkaOffset())
+        : convertKey(record.keySchema(), record.key());
+
+    if (record.value() != null) {
+      if (config.writeMethod() == WriteMethod.UPSERT) {
+        return new UpdateRequest(index, id)
+            .doc(payload, XContentType.JSON)
+            .upsert(payload, XContentType.JSON)
+            .retryOnConflict(Math.max(config.maxInFlightRequests(), 5));
```
avocader, line 182: Why are we capping the number of retries to 5?

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -152,51 +144,71 @@ public IndexableRecord convertRecord(
...
+      }
+
+      return addExternalVersioning(
```
avocader, line 185: Maybe add a comment or use a `switch` to handle the `WriteMethod` enum so that it's more obvious that we are handling `INSERT`?

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -152,51 +144,71 @@ public IndexableRecord convertRecord(
...
     }
 
-    final String payload = getPayload(record, ignoreSchema);
-    final Long version = ignoreKey ? null : record.kafkaOffset();
-    return new IndexableRecord(new Key(index, type, id), payload, version);
+    return addExternalVersioning(new DeleteRequest(index).id(id), record);
   }
 
-  private String getPayload(SinkRecord record, boolean ignoreSchema) {
+  private DocWriteRequest<?> addExternalVersioning(DocWriteRequest<?> request, SinkRecord record) {
```
avocader, line 194: Looks like we are adding external versioning only for the case when we are not ignoring the key, so maybe name the method as `maybeAddExternalVersioning`?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
-  void createMapping(String index, String type, Schema schema) throws IOException;
+  private void configureAuthentication(HttpAsyncClientBuilder builder) {
+
+    CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
+    if (config.isAuthenticatedConnection()) {
+      config.connectionUrls().forEach(
+          url ->
+              credentialsProvider.setCredentials(
+                  new AuthScope(new HttpHost(url)),
+                  new UsernamePasswordCredentials(config.username(), config.password().value())
```
avocader, line 327: Can password be null?

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -0,0 +1,451 @@
...
+
+  @After
+  public void cleanup() throws IOException {
+    if (helperClient.indexExists(INDEX)){
+      helperClient.deleteIndex(INDEX);
+    }
+  }
+
+  @Test
+  public void testClose() {
```
avocader, line 102: Is it possible to test a negative case when `bulkProcessor` doesn't finish within a minute?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+    }
+
+    CreateIndexRequest request = new CreateIndexRequest(index);
+    callWithRetries(
+        "create index " + index,
+        () -> {
+          try {
+            client.indices().create(request, RequestOptions.DEFAULT);
+          } catch (ElasticsearchStatusException | IOException e) {
+            if (!e.getMessage().contains(RESOURCE_ALREADY_EXISTS_EXCEPTION)) {
```
avocader, line 171: Is it still possible to get this exception if we already checked whether the index exists earlier?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+  /**
+   * Creates a mapping for the given index and schema.
+   *
+   * @param index the index to create the mapping for
+   * @param schema the schema to map
+   */
+  public void createMapping(String index, Schema schema) {
+    PutMappingRequest request = new PutMappingRequest(index).source(Mapping.buildMapping(schema));
+    callWithRetries(
+        "create mapping",
```
avocader, line 190: Add index and schema to the description as in `createIndex`?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+  }
+
+  /**
+   * Checks whether the index already has a mapping or not.
+   * @param index the index to check
+   * @return true if a mapping exists, false if it does not
+   */
+  public boolean hasMapping(String index) {
+    MappingMetaData mapping = getMapping(index);
+    return mapping != null && !mapping.sourceAsMap().isEmpty();
```
avocader, line 209: Can `mapping.sourceAsMap()` be `null` here?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+      close();
+      throw error.get();
+    }
+
+    if (recordMap != null) {
+      // make sure that only unique records are being sent in a batch
+      while (recordMap.contains(request.id())) {
+        flush();
+        try {
+          Thread.sleep(TimeUnit.SECONDS.toMillis(1));
```
avocader, line 231: Isn't 1 millisecond too short?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+    };
+  }
 
   /**
-   * Creates indices.
+   * Calls the specified function with retries and backoffs until the retries are exhausted or the
+   * function succeeds.
    *
-   * @param indices the set of index names to create, not null
+   * @param description description of the attempted action n present tense
```
avocader, line 296:
```suggestion
   * @param description description of the attempted action present tense
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+   * Configures the client to use SSL if configured.
    *
-   * @param index the index
-   * @param type the type
-   * @throws IOException if the client cannot execute the request
+   * @param builder the HttpAsyncClientBuilder
    */
-  JsonObject getMapping(String index, String type) throws IOException;
+  private void configureSslContext(HttpAsyncClientBuilder builder) {
+    SslFactory kafkaSslFactory = new SslFactory(Mode.CLIENT, null, false);
```
avocader, line 354: We are using this SSL context to talk to ES, not to Kafka, right? The variable name is confusing.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+
+      try {
+        sslContext =
+            (SSLContext) sslEngine.getClass().getDeclaredMethod("sslContext").invoke(sslEngine);
+      } catch (Exception ex) {
+        throw new ConnectException("Could not create SSLContext.", ex);
+      }
+    }
+
+    HostnameVerifier hostnameVerifier = config.shouldDisableHostnameVerification()
```
avocader, line 396: Is this needed for self-signed certificates? Maybe add a comment why are we doing this

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+      if (response.getFailureMessage().contains(VERSION_CONFLICT_EXCEPTION)) {
+        log.warn(
+            "Ignoring version conflict for operation {} on document '{}' version {} in index '{}'.",
+            response.getOpType(),
+            response.getId(),
+            response.getVersion(),
+            response.getIndex()
+        );
+
+        reportBadRecord(response);
```
avocader, line 478: The method's Javadoc and the `warn` message says we are ignoring version conflict,s but it seems like we are still reporting this record. Is that intentional?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -15,152 +15,191 @@
...
-        log.warn("AK versions prior to 2.6 do not support the errant record reporter");
+    log.info("Started ElasticsearchSinkTask.");
+  }
+
+  @Override
+  public void put(Collection<SinkRecord> records) throws ConnectException {
+    log.debug("Putting {} records to Elasticsearch.", records.size());
+    for (SinkRecord record : records) {
+      if (shouldSkipRecord(record)) {
+        log.trace("Ignoring sink {} with null value.", recordString(record));
```
avocader, line 83:
```suggestion
        log.trace("Ignoring sink record {} with null value.", recordString(record));
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -15,152 +15,191 @@
...
+    if (index.startsWith("-") || index.startsWith("_")) {
+      index = index.substring(1);
     }
+
+    if (index.equals(".") || index.equals("..")) {
+      index = index.replace(".", "dot");
+      log.warn("Elasticsearch cannot have indices named {}. Index will be named {}.", topic, index);
+    }
+
+    log.trace("Topic '{}' was translated to index '{}'.", topic, index);
```
avocader, line 163: Should it be `warn`?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -15,152 +15,191 @@
...
+
+  private void tryWriteRecord(SinkRecord sinkRecord) {
+    DocWriteRequest<?> record = null;
+    try {
+      record = converter.convertRecord(sinkRecord, convertTopicToIndexName(sinkRecord.topic()));
+    } catch (DataException convertException) {
+      reportBadRecord(sinkRecord, convertException);
+
+      if (config.dropInvalidMessage()) {
+        log.error("Can't convert {}.", recordString(sinkRecord), convertException);
```
avocader, line 185: Missing a placeholder?

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
```diff
@@ -27,94 +23,206 @@
...
-      String type,
-      Schema schema
-  )
-      throws IOException {
-    client.createMapping(index, type, schema);
+  public static XContentBuilder buildMapping(Schema schema) {
+    try {
+      XContentBuilder builder = XContentFactory.jsonBuilder();
+      builder.startObject();
+      {
```
avocader, line 66: Why do we need these curly brackets?

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
```diff
@@ -27,94 +23,206 @@
...
+  }
+
+  private static XContentBuilder inferPrimitive(
+      XContentBuilder builder,
+      String type,
+      Object defaultValue
+  ) throws IOException {
+
+    if (type == null) {
+      throw new DataException("Invalid primitive type " + type + ".");
```
avocader, line 161: Use string formatting instead?

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -0,0 +1,451 @@
...
+import org.elasticsearch.ElasticsearchStatusException;
+import org.elasticsearch.common.xcontent.XContentBuilder;
+import org.elasticsearch.common.xcontent.XContentFactory;
+import org.elasticsearch.search.SearchHit;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.testcontainers.shaded.okio.Sink;
```
avocader, line 60: Unused import.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -0,0 +1,451 @@
...
+
+    client.createIndex(INDEX);
+    assertTrue(helperClient.indexExists(INDEX));
+
+    client.createIndex(INDEX);
+    assertTrue(helperClient.indexExists(INDEX));
+  }
+
+  @Test
+  public void testIndexExists() throws IOException {
```
avocader, line 129: What is the difference between this one and `testIndexAlreadyExists`?

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -0,0 +1,451 @@
...
+    client.createIndex(INDEX);
+    assertTrue(helperClient.indexExists(INDEX));
+  }
+
+  @Test
+  public void testIndexAlreadyExists() throws IOException {
+    ElasticsearchClient client = new ElasticsearchClient(config, null);
+    assertFalse(helperClient.indexExists(INDEX));
+
+    client.createIndex(INDEX);
```
avocader, line 121: Maybe change `createIndex` to return `true` if the index was created?

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -0,0 +1,451 @@
...
+    builder.flush();
+    ByteArrayOutputStream stream = (ByteArrayOutputStream) builder.getOutputStream();
+    System.out.println(stream.toString());
+
+    helperClient.getMapping(INDEX).sourceAsMap();
+    XContentBuilder mapping = XContentFactory.jsonBuilder();
+    mapping.map(helperClient.getMapping(INDEX).sourceAsMap());
+    mapping.flush();
+    stream = (ByteArrayOutputStream) mapping.getOutputStream();
+    System.out.println(stream.toString());
```
avocader, line 164: nit: delete this?

src/test/java/io/confluent/connect/elasticsearch/RetryUtilTest.java
```diff
@@ -45,6 +61,40 @@ public void computeNonRandomRetryTimes() {
...
+    MockTime mockClock = new MockTime();
+
+    assertTrue(RetryUtil.callWithRetries("test", () -> testFunction(4), 3, 100, mockClock));
+    verify(mockClock, times(3)).sleep(anyLong());
+  }
+
+  private boolean testFunction(int timesToThrow) throws IOException {
+    if (timesThrown < timesToThrow) {
+      timesThrown++;
+      throw new IOException("oh no i iz borke, plz retry");
```
avocader, line 92: Looks like this function is drunk :)

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorBaseIT.java
```diff
@@ -113,47 +109,45 @@ protected void runSimpleTest(Map<String, String> props) throws Exception {
...
     );
   }
+
+  protected void writeRecords(int numRecords) {
+    for (int i  = 0; i < numRecords; i++) {
+      connect.kafka().produce(TOPIC, String.valueOf(i), String.format("{\"doc_num\":%d}", i));
+    }
+  }
+
+  protected void writeRecordsFromIndex(int start, int numRecords) {
```
avocader, line 148:
```suggestion
  protected void writeRecordsStartFromIndex(int start, int numRecords) {
```

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -152,51 +144,71 @@ public IndexableRecord convertRecord(
...
-           + "+" + String.valueOf((int) record.kafkaPartition())
-           + "+" + String.valueOf(record.kafkaOffset());
-    } else {
-      id = convertKey(record.keySchema(), record.key());
+    final String payload = getPayload(record);
+    final String id = config.shouldIgnoreKey(record.topic())
+        ? String.format("%s+%d+%d", record.topic(), record.kafkaPartition(), record.kafkaOffset())
+        : convertKey(record.keySchema(), record.key());
+
+    if (record.value() != null) {
```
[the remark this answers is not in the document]
levzem: yes, if behavior.on.null.values=DELETE

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
levzem: good catch, should've been `min`

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
levzem: record is generated as part of `recordString()`

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
levzem: not necessarily, it could be a straight translation from the topic. I can add `warn` for every if branch instead if you like

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
levzem: logger can take an exception as the last argument without a placeholder

src/main/java/io/confluent/connect/elasticsearch/Mapping.java
levzem: this is the convention that ES documentation used, so I kept for
1. consistency
2. readability (helps visualize nested objects)

src/test/java/io/confluent/connect/elasticsearch/RetryUtilTest.java
levzem: 🍻

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
levzem: we have seen it before in escalations and this maintains the previous behavior of the `JestClient` so I kept it just to be safe

unfortunately we have not been able to understand why this happens

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
levzem: in my tests, if an index does not have a mapping, it returns an empty map. the ES documentation does not specify unfortunately

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
levzem: it is 1 seconds or 1000 milliseconds

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+  }
 
-  enum Version {
-    ES_V1, ES_V2, ES_V5, ES_V6, ES_V7
+  /**
+   * Maps a record to the document id.
+   * @param id the document id
+   * @param record the record
+   */
+  private void addToRecordMap(String id, SinkRecord record) {
```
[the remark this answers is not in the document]
levzem: leftover from `BulkProcessor` implementation, while yes it is only called once, given it is wrapped in a `null` check it is probably better to abstract it away in case of future use

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,568 @@
...
+      public void afterBulk(long executionId, BulkRequest request, BulkResponse response) {
+        for (BulkItemResponse bulkItemResponse : response) {
+          handleResponse(bulkItemResponse);
+          removeFromRecordMap(bulkItemResponse.getId());
+        }
+      }
+
+      @Override
+      public void afterBulk(long executionId, BulkRequest request, Throwable failure) {
+        error.compareAndSet(null, new ConnectException("Bulk request failed.", failure));
```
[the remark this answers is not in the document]
levzem: this will fail the task so it wont matter if we empty the map or not, but I added it regardless

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
levzem: nope :), https://github.com/confluentinc/kafka-connect-elasticsearch/blob/master/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java#L657-L659

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
levzem: yes to ES, but this class is provided by `Kafka` hence the naming. this is ported from the `JestClient`, but can change to `sslFactory` if you feel it is necessary

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
levzem: this section is ported from `JestClient` 

here is the PR that added this functionality https://github.com/confluentinc/kafka-connect-elasticsearch/pull/410

doesn't say the reason we need this, not sure myself

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
levzem: reporting the record is the same as ignoring it (reporting a bad record does not fail anything), but I figured the user should be know which records had version conflicts in case they want to remediate manually somehow

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
levzem: the method it is overriding is `synchronized`

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
levzem: `testIndexAlreadyExists` tests that the client does not try to create an index that does not already exist

I will update the name

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
levzem: oops

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
levzem: not quite sure what you mean, could you provide an example?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
avocader: check for `mapping.sourceAsMap() != null` to be on a safe side?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,667 @@
...
+    // limit the internal buffer
+    long maxWaitTime = System.currentTimeMillis() + config.flushTimeoutMs();
+    while (numRecords.get() >= config.maxBufferedRecords()) {
+      try {
+        Thread.sleep(TimeUnit.SECONDS.toMillis(1));
+      } catch (InterruptedException e) {
+        throw new ConnectException(e);
+      }
+      if (System.currentTimeMillis() > maxWaitTime) {
+        throw new ConnectException("Could not make space in the internal buffer fast enough.");
```
avocader, line 275: Should it be actionable? E.g. should users update `max.buffered.records` parameter?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,667 @@
...
+   * @return true if a mapping exists, false if it does not
+   */
+  public boolean hasMapping(String index) {
+    MappingMetaData mapping = mapping(index);
+    return mapping != null && !mapping.sourceAsMap().isEmpty();
+  }
+
+  /**
+   * Buffers a record to index. Will ensure that there are no concurrent requests for the same
+   * document id.
```
avocader, line 242: IIUC, the latter statement is valid for a case when DLQ is enabled or `key.ignore=false`. Maybe update the Javadoc and mention, why we are not ensuring the absence of concurrent requests for some cases.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,667 @@
...
+        flush();
+        try {
+          Thread.sleep(TimeUnit.SECONDS.toMillis(1));
+        } catch (InterruptedException e)  {
+          throw new ConnectException(e);
+        }
+      }
+    }
+
+    // limit the internal buffer
```
avocader, line 266: nit: for better readability, maybe mention that we are waiting until the internal buffer's size drops to a minimal configured size(set by `max.buffered.records`)

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
avocader: IMO it's less confusing if you call it `sslFactory`.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
avocader: I think it's important information if we are changing the index name to be different from the expected name(topic name).

src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
```diff
@@ -86,5 +94,73 @@ public static long computeRetryWaitTimeInMillis(int retryAttempts,
...
 
+  /**
+   * Call the supplied function up to the {@param maxAttempts maximum number of attempts}.
+   *
+   * <p>The description of the function should be a succinct, human-readable present tense phrase
+   * that summarizes the function, such as "read tables" or "connect to database" or
+   * "make remote request". This description will be used within exception and log messages.
+   *
+   * @param description      present tense description of the action, used to create the error
+   *                         message; may not be null
```
avocader, line 105: Should we add an assertion that it's not null?

src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
```diff
@@ -86,5 +94,73 @@ public static long computeRetryWaitTimeInMillis(int retryAttempts,
...
+      ++attempt;
+      try {
+        log.trace(
+            "Try {} (attempt {} of {})",
+            description,
+            attempt,
+            maxAttempts
+        );
+        return function.call();
+      } catch (IOException e) {
```
avocader, line 155: If this is a generic util method, do we have to catch `Exception` instead?

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
```diff
@@ -46,6 +40,8 @@
 
   private static Logger log = LoggerFactory.getLogger(ElasticsearchConnectorIT.class);
 
+  // TODO: test compatibility
```
avocader, line 43: Was it addressed already?

levzem: this will be a separate ticket

src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
levzem: no because the client can throw runtime exceptions which should not be retried only http client exceptions should be retried

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -40,25 +46,22 @@
...
+  protected static final String MAP_VALUE = "value";
 
   static {
     JSON_CONVERTER = new JsonConverter();
     JSON_CONVERTER.configure(Collections.singletonMap("schemas.enable", "false"), false);
   }
 
-  private final boolean useCompactMapEntries;
-  private final BehaviorOnNullValues behaviorOnNullValues;
+  private ElasticsearchSinkConnectorConfig config;
```
avocader, line 64: nit: make it `final`?

src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -152,58 +143,84 @@ public IndexableRecord convertRecord(
...
+                  BehaviorOnNullValues.IGNORE
+              )
+          );
         default:
-          throw new RuntimeException(String.format(
-              "Unknown value for %s enum: %s",
-              BehaviorOnNullValues.class.getSimpleName(),
-              behaviorOnNullValues
-          ));
+          throw new ConnectException(
```
avocader, line 161: nit: do we even need this section? Since `behaviorOnNullValues()` returns enum type `BehaviorOnNullValues`?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,677 @@
...
+      Object sslEngine;
+      try {
+        // try AK <= 2.6 second
+        sslEngine = SslFactory.class.getDeclaredMethod("sslEngineBuilder").invoke(sslFactory);
+        log.debug("Using AK 2.2-2.5 SslFactory methods.");
+
+      } catch (Exception ex) {
+        // must be running AK 2.6+
+        log.debug(
+            "Could not find Ak 2.3-2.5 methods for SslFactory."
```
avocader, line 434: "Could not find AK 2.3-2.5 methods for SslFactory."
```suggestion
            "Could not find Ak 2.3-2.5 methods for SslFactory."
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,677 @@
...
+   * @param builder the HttpAsyncClientBuilder
    */
-  default void refresh() throws IOException {
-    throw new UnsupportedOperationException("refresh is not implemented yet by this client");
+  private void configureSslContext(HttpAsyncClientBuilder builder) {
+    SslFactory sslFactory = new SslFactory(Mode.CLIENT, null, false);
+    sslFactory.configure(config.sslConfigs());
+
+    SSLContext sslContext;
+    try {
```
avocader, line 417: Have you thought about breaking this method down, something like:
```

try {
      // try AK <= 2.2 first
      sslContext = tryLessOrEqual22(sslFactory);
    } catch (Exception e) {
      sslContext = tryGreater22(sslFactory);
}
```
and then
```
private void tryGreater22(SslFactory sslFactory) {
Object sslEngine;
      try {
        // try AK <= 2.6 second
        sslEngine = tryLessOrEqual26(sslFactory)
      } catch (Exception ex) {
        // must be running AK 2.6+
        sslEngine = tryGreater26(sslFactory)
      }
}
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -15,135 +15,199 @@
...
-  public void stop() throws ConnectException {
-    log.info("Stopping ElasticsearchSinkTask");
-    if (writer != null) {
-      writer.stop();
-      writer = null;
+  public String version() {
+    return Version.getVersion();
+  }
+
+  private void checkMapping(SinkRecord record) {
```
avocader, line 119: This method name is a bit misleading IMO, maybe consider renaming it to `maybeCreateMapping`?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -15,135 +15,199 @@
...
+    String index = convertTopicToIndexName(record.topic());
+    if (!config.shouldIgnoreSchema(record.topic()) && !existingMappings.contains(index)) {
+      if (!client.hasMapping(index)) {
+        client.createMapping(index, record.valueSchema());
+      }
+      log.debug("Locally caching mapping for index '{}'", index);
```
avocader, line 125: nit: not a native speaker here, but maybe `Caching mapping for index '{}' locally`?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -15,135 +15,199 @@
...
+   * Returns the converted index name from a given topic name. Elasticsearch accepts:
+   * <ul>
+   *   <li>all lowercase</li>
+   *   <li>less than 256 bytes</li>
+   *   <li>does not start with - or _</li>
+   *   <li>is not . or ..</li>
+   * </ul>
+   * (<a href="https://github.com/elastic/elasticsearch/issues/29420">ref</a>_.)
+   */
+  private String convertTopicToIndexName(String topic) {
```
avocader, line 140: I recommend adding a `warn` message for every case we can't have 1-1 topic to index name mapping.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfigTest.java
```diff
@@ -140,4 +140,9 @@ public void shouldNotAllowInvalidSecurityProtocol() {
     props.put(SECURITY_PROTOCOL_CONFIG, "unsecure");
     new ElasticsearchSinkConnectorConfig(props);
   }
+
+  public static Map<String, String> addNecessaryProps(Map<String, String> props) {
+      props.put(ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG, "http://localhost:8080");
```
avocader, line 145: nit: check that `props!=null`?

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTaskTest.java
```diff
@@ -15,129 +15,242 @@
...
 import java.util.Map;
-import java.util.Collection;
-import java.util.ArrayList;
 import java.util.Collections;
 
+public class ElasticsearchSinkTaskTest {
 
-@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
-public class ElasticsearchSinkTaskTest extends ElasticsearchSinkTestBase {
+  protected static final String TOPIC = "topic";
```
avocader, line 49: Very creative name :)

src/test/java/io/confluent/connect/elasticsearch/RetryUtilTest.java
```diff
@@ -45,6 +61,40 @@ public void computeNonRandomRetryTimes() {
...
+  public void testCallWithRetriesNoRetries() throws Exception {
+    MockTime mockClock = new MockTime();
+    long expectedTime = mockClock.milliseconds();
+
+    assertTrue(RetryUtil.callWithRetries("test", () -> testFunction(0), 3, 100, mockClock));
+    assertEquals(expectedTime, mockClock.milliseconds());
+  }
+
+  @Test
+  public void testCallWithRetriesSomeRetries() throws Exception {
```
avocader, line 74: nit: add a unit-test that ensures wait time is always lesser than `MAX_RETRY_TIME_MS`?

src/test/java/io/confluent/connect/elasticsearch/helper/ElasticsearchHelperClient.java
```diff
@@ -0,0 +1,87 @@
...
+    GetMappingsResponse response = client.indices().getMapping(request, RequestOptions.DEFAULT);
+    return response.mappings().get(index);
+  }
+
+  public boolean indexExists(String index) throws IOException {
+    GetIndexRequest request = new GetIndexRequest(index);
+    return client.indices().exists(request, RequestOptions.DEFAULT);
+  }
+
+  public void refresh(String index) throws IOException {
```
avocader, line 70: Not used anywhere, do we need it?

avocader (approved):
I left a few non-critical suggestions, LGTM otherwise, thanks @levzem !

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,677 @@
...
+    } catch (Exception e) {
+      // must be running AK 2.3+
+      log.debug("Could not find AK 2.2 SslFactory methods. Trying AK 2.3+ methods for SslFactory.");
+
+      Object sslEngine;
+      try {
+        // try AK <= 2.6 second
+        sslEngine = SslFactory.class.getDeclaredMethod("sslEngineBuilder").invoke(sslFactory);
+        log.debug("Using AK 2.2-2.5 SslFactory methods.");
+
```
kkonstantine, line 430:
```suggestion
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
kkonstantine, line 434: I wonder if AK is obvious that means Apache Kafka to everyone. 

```suggestion
            "Could not find AK 2.3-2.5 methods for SslFactory."
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
kkonstantine: You don't mutate anything. Fine to keep the contract though.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -15,135 +15,199 @@
...
+    String index = topic.toLowerCase();
+    if (index.length() > 255) {
```
kkonstantine, line 142: Happy coincidence that Kafka topic names are more limited and accept only single byte characters. But if multi-byte were accepted this wouldn't necessarily be correct. 

Should we link to the docs instead of a github issue?
E.g.: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#indices-create-api-path-params

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -33,6 +33,7 @@
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.SECURITY_PROTOCOL_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.SSL_CONFIG_PREFIX;
 
+import io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.SecurityProtocol;
```
kkonstantine, line 36: nit: usually static imports will follow regular imports. 
Not sure we need more stylistic changes though. Just noticed.

kkonstantine (commented):
Another round of comments. Mostly done. I have a few more files to check. Mostly tests.

src/test/java/io/confluent/connect/elasticsearch/RetryUtilTest.java
levzem: already tested by https://github.com/confluentinc/kafka-connect-elasticsearch/blob/master/src/test/java/io/confluent/connect/elasticsearch/RetryUtilTest.java#L57
and
https://github.com/confluentinc/kafka-connect-elasticsearch/blob/master/src/test/java/io/confluent/connect/elasticsearch/RetryUtilTest.java#L23-L46

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,677 @@
...
+   * @throws ConnectException if all of the records fail to flush before the timeout.
    */
-  void createIndices(Set<String> indices);
+  public void close() {
+    try {
+      if (!bulkProcessor.awaitClose(config.flushTimeoutMs(), TimeUnit.MILLISECONDS)) {
+        throw new ConnectException("Failed to process all outstanding requests in time.");
+      }
+    } catch (InterruptedException e) {
+      log.error("Bulk processor close was interrupted.", e);
```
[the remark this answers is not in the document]
levzem: updated the logic, let me know what you think. should catch the leak now

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -15,102 +15,677 @@
...
-   * @param batch the list of records
-   * @return the bulk request
+   * @param record the record to index
+   * @param request the associated request to send
+   * @throws ConnectException if one of the requests failed
    */
-  BulkRequest createBulkRequest(List<IndexableRecord> batch);
+  public void index(SinkRecord record, DocWriteRequest<?> request) {
+    if (isFailed()) {
+      close();
```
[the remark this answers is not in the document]
levzem: put the close in a try catch

levzem:
@kkonstantine, addressed all review comments, ready for another pass!

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -0,0 +1,526 @@
...
+
+    writeRecord(sinkRecord(1), client);
+    client.flush();
+
+    waitUntilRecordsInES(2);
+    assertEquals(2, helperClient.getDocCount(INDEX));
+    client.close();
+  }
+
+  @Test(expected = ConnectException.class)
```
kkonstantine, line 369: It'd be nice to start using assertThrows and available hamcrest matchers for maps etc. But we can consider on a follow up.

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorBaseIT.java
```diff
@@ -53,7 +53,7 @@
```
kkonstantine, line 53: TYPE is unused

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
```diff
@@ -141,21 +125,16 @@ public void testUpsert() throws Exception {
...
-    JsonObject result = client.search("", TOPIC, null);
-    JsonArray rawHits = result.getAsJsonObject("hits").getAsJsonArray("hits");
-
-    for (int i = 0; i < rawHits.size(); ++i) {
-      JsonObject hitData = rawHits.get(i).getAsJsonObject();
-      JsonObject source = hitData.get("_source").getAsJsonObject();
-      assertTrue(source.has("doc_num"));
-      if (Integer.valueOf(hitData.get("_id").getAsString()) == lastRecord) {
+    for (SearchHit hit : helperClient.search(TOPIC)) {
+      if (Integer.valueOf(hit.getId()) == lastRecord) {
```
kkonstantine, line 134:
```suggestion
      if (Integer.parseInt(hit.getId()) == lastRecord) {
```

kkonstantine (approved):
Very nice work.
Left a few more minor comments. 
Also please make sure to run `mvn javadoc:javadoc` and fix the javadoc issues because they'll break during the release. 

But this LGTM! 👏

PR #470, opened 2020-11-11
CC-12597: remove topic.index.map

levzem:

## Problem
`topic.index.map` is deprecated and is set to be removed for the upcoming jest client replacement work #468 

## Solution
remove the config

- [x] yes
- [ ] no

##### If yes, where?
update the docs to mark this config as removed

## Test Strategy
N/A - verify existing tests still pass

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
`master` to a new major version because this breaks backward compatibility

### DIFF
--- a/src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
+++ b/src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
@@ -17,6 +17,7 @@ package io.confluent.connect.elasticsearch;
 
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.confluent.connect.elasticsearch.bulk.BulkProcessor.BehaviorOnMalformedDoc;
 import org.apache.kafka.connect.data.Decimal;
 import org.apache.kafka.connect.data.Schema;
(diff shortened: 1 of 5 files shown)


src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
```diff
@@ -379,7 +361,7 @@ public void testIneffectiveDelete() throws Exception {
 
     ElasticsearchWriter writer = initWriter(client, BehaviorOnNullValues.DELETE);
     writeDataAndRefresh(writer, records);
-    verifySearchResults(new ArrayList<SinkRecord>());
+    verifySearchResults(new ArrayList<>());
```
ilanjiR, line 364: Any particular reason we changed this?

ilanjiR (approved):
LGTM with 1 comment

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchWriterTest.java
levzem: just cleaning up unnecessary types since that method accepts `Collection<SinkRecord>`

fidemin:
I wonder why this feature was removed. Is there any way to map kafka topic to elasticsearch index or alias?

PR #472, opened 2020-11-16
CC-12825: remove auto.create.index.at.start config

levzem:

## Problem
`auto.create.index.at.start` is an obsolete config and the connector should permanently function as if `auto.create.index.at.start=false`

## Solution
remove the config

- [ ] yes
- [x] no

## Test Strategy
remove old test and modify existing test to make sure index is still being created properly

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
`master` and bump to `11.0.x` because breaking backward compatibility

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -159,13 +159,6 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   private static final String READ_TIMEOUT_MS_DISPLAY = "Read Timeout";
   private static final int READ_TIMEOUT_MS_DEFAULT = (int) TimeUnit.SECONDS.toMillis(3);
 
-  public static final String CREATE_INDICES_AT_START_CONFIG = "auto.create.indices.at.start";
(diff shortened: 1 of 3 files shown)


ilanjiR (approved):
LGTM

1300371:
This ticket annonces that `auto.create.index.at.start=false` is now the default behaviour but the current version `14.0.0` tries to create index when starting connector.  

```
/**
   * Creates an index or data stream. Will not recreate the index or data stream if
   * it already exists. Will create a data stream instead of an index if the data stream
   * configurations are set.
   *
   * @param name the name of the index or data stream to create
   * @return true if the index or data stream was created, false if it already exists
   */
  public boolean createIndexOrDataStream(String name) {
    if (indexExists(name)) {
      return false;
    }
    return config.isDataStream() ? createDataStream(name) : createIndex(name);
  }
```

Did I missed some other changes or documentation about this ? 
I don't want the connector creates index at statup. This is something we can change with the current version ?  
See https://docs.confluent.io/kafka-connectors/elasticsearch/current/changelog.html#version-11-0-0  
https://github.com/confluentinc/kafka-connect-elasticsearch/blob/a4671d70a1c9d58bc9368cd6664e8ad103f0328c/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java#L251

PR #475, opened 2020-11-30
CC-8546: add Kerberos support

levzem:
## Problem
the connector does not support kerberos authentication

## Solution
implement kerberos authentication, built on top of #468 

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests - new ITs testing kerberos auth
- [ ] System tests
- [x] Manual tests - set up a KDC docker instance and verified manually the connect can authenticate with kerberos

## Release Plan

 
new feature hence `master`, building on top of #468

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ClientConfigCallbackHandler.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ClientConfigCallbackHandler.java
@@ -291,9 +291,10 @@ public class ClientConfigCallbackHandler implements HttpClientConfigCallback {
   }
 
   /**
-   * Creates and returns a login context for the given kerberos user principle.
+   * Logs in and returns a login context for the given kerberos user principle.
+   *
    * @return the login context
-   * @throws PrivilegedActionException
+   * @throws PrivilegedActionException if the login failed
    */
   private LoginContext loginContext() throws PrivilegedActionException {
     Configuration conf = new Configuration() {
--- a/src/main/java/io/confluent/connect/elasticsearch/ClientConfigCallbackHandler.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ClientConfigCallbackHandler.java
@@ -17,9 +17,6 @@ package io.confluent.connect.elasticsearch;
 
 import com.sun.security.auth.module.Krb5LoginModule;
 import java.security.AccessController;
-import java.security.KeyManagementException;
-import java.security.KeyStoreException;
-import java.security.NoSuchAlgorithmException;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.util.Collections;
@@ -45,7 +42,6 @@ import org.apache.http.config.Registry;
 import org.apache.http.config.RegistryBuilder;
 import org.apache.http.conn.ssl.NoopHostnameVerifier;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
-import org.apache.http.conn.ssl.TrustAllStrategy;
 import org.apache.http.impl.auth.SPNegoSchemeFactory;
 import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
@@ -57,7 +53,6 @@ import org.apache.http.nio.conn.SchemeIOSessionStrategy;
 import org.apache.http.nio.conn.ssl.SSLIOSessionStrategy;
 import org.apache.http.nio.reactor.ConnectingIOReactor;
 import org.apache.http.nio.reactor.IOReactorException;
-import org.apache.http.ssl.SSLContextBuilder;
 import org.apache.kafka.common.network.Mode;
 import org.apache.kafka.common.security.ssl.SslFactory;
 import org.apache.kafka.connect.errors.ConnectException;
@@ -207,11 +202,10 @@ public class ClientConfigCallbackHandler implements HttpClientConfigCallback {
     builder.setDefaultAuthSchemeRegistry(authSchemeRegistry);
 
     try {
-      LoginContext loginContext = login();
+      LoginContext loginContext = loginContext();
       GSSCredential credential = Subject.doAs(
           loginContext.getSubject(),
-          (PrivilegedExceptionAction<GSSCredential>) ()
-              -> gssManager.createCredential(
+          (PrivilegedExceptionAction<GSSCredential>) () -> gssManager.createCredential(
               null,
               GSSCredential.DEFAULT_LIFETIME,
               SPNEGO_OID,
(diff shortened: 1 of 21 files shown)


sonupillai:
@levzem Could you please add details on the files which needs to be reviewed for only Kerb support. Also could you please add a note around how you tested this manually using the authentic kerb server? I believe IT's use mini kdc

levzem:
@sonupillai all the files that need to be reviewed are under one commit https://github.com/confluentinc/kafka-connect-elasticsearch/pull/475/commits/4cc12b4a118b89e6f064dc5926d55f2e9719a4b5

I spoke with Stan and he said that MiniKDC is sufficient. I can run a manual test but I don't see what it gives us.

src/test/java/io/confluent/connect/elasticsearch/helper/ElasticsearchContainer.java
```diff
@@ -158,67 +160,108 @@ public boolean isSslEnabled() {
...
+  }
+
   @Override
   protected void configure() {
     super.configure();
-    Future<String> image;
+
+    waitingFor(
+        Wait
+            .forLogMessage(".*(Security is enabled|license .* valid).*", 1)
```
kkonstantine, line 206: asdf
```suggestion
        Wait.forLogMessage(".*(Security is enabled|license .* valid).*", 1)
```

kkonstantine (commented):
Took a first pass, but we'll have to wait for a rebase to take another look. 
Seems good overall.

tikimims:
@kkonstantine @levzem Just heads up, there is a Jenkins build failure with this one.

src/main/java/io/confluent/connect/elasticsearch/ClientConfigCallbackHandler.java
```diff
@@ -210,4 +308,61 @@ private SSLContext sslContext() {
...
+    configs.put("principal", config.kerberosUserPrincipal());
+    configs.put("storeKey", "false");
+    configs.put("doNotPrompt", "true");
+    configs.put("isInitiator", "true");
+    return configs;
+  }
+
+  private static Oid spnegoOid() {
+    try {
+      return new Oid("1.3.6.1.5.5.2");
```
dosvath, line 363: What does this ID represent? We could use a constant with a descriptive name for this value.

src/main/java/io/confluent/connect/elasticsearch/ClientConfigCallbackHandler.java
```diff
@@ -210,4 +308,61 @@ private SSLContext sslContext() {
       }
     }
   }
+
+  private LoginContext login() throws PrivilegedActionException {
```
dosvath, line 312: Comment to describe this gets the context to log into ES via Kerberos/ or more descriptive method name will be helpful.

src/main/java/io/confluent/connect/elasticsearch/ClientConfigCallbackHandler.java
```diff
@@ -151,6 +191,64 @@ private PoolingNHttpClientConnectionManager configureConnectionManager() {
...
+      builder.setDefaultCredentialsProvider(credentialsProvider);
+    } catch (PrivilegedActionException e) {
+      throw new ConnectException(e);
+    }
+
+    if (!config.isSslEnabled()) {
+      try {
+        SSLContextBuilder sslContextBuilder = new SSLContextBuilder();
+        sslContextBuilder.loadTrustMaterial(null, new TrustAllStrategy());
+        builder.setSSLContext(sslContextBuilder.build());
```
dosvath, line 240: the default SSL Context is to have this on in the builder? Wondering why we need to create a null builder when SSL is disabled

dosvath (commented):
Looks close, I have one conceptual question on the ES client configs. In many cases Kerberos clients require to specify the KDC  IP address/domain, realm and ports. I wasn't able to clearly see if/where we are setting that for the ES client. Do you know why those may not be needed if that's the case?

src/main/java/io/confluent/connect/elasticsearch/ClientConfigCallbackHandler.java
```diff
@@ -210,4 +308,61 @@ private SSLContext sslContext() {
...
+    configs.put("useTicketCache", "false");
+    configs.put("useKeyTab", "true");
+    configs.put("keyTab", config.keytabPath());
+    //Krb5 in GSS API needs to be refreshed so it does not throw the error
+    //Specified version of key is not available
+    configs.put("refreshKrb5Config", "true");
+    configs.put("principal", config.kerberosUserPrincipal());
+    configs.put("storeKey", "false");
+    configs.put("doNotPrompt", "true");
+    configs.put("isInitiator", "true");
```
SajanaW, line 357: So [here's my jaas config for Cassandra ](https://github.com/confluentinc/kafka-connect-cassandra/blob/master/src/main/java/io/confluent/connect/cassandra/CassandraSessionFactoryImpl.java#L49-L56).  
I set the following because I think the connector needs to refresh the TGT.  WDYT?
`   loginConfig.put("useTicketCache", "true");
      loginConfig.put("renewTGT", "true");`

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -276,16 +277,30 @@
...
-          + "Values can be `PLAINTEXT` or `SSL`. If `PLAINTEXT` is passed, "
-          + "all configs prefixed by " + SSL_CONFIG_PREFIX + " will be ignored.";
+      "The security protocol to use when connecting to Elasticsearch. Values can be `PLAINTEXT` or"
+          + " `SSL`. If `PLAINTEXT` is passed, all configs prefixed by " + SSL_CONFIG_PREFIX
+          + " will be ignored.";
   private static final String SECURITY_PROTOCOL_DISPLAY = "Security protocol";
   private static final String SECURITY_PROTOCOL_DEFAULT = SecurityProtocol.PLAINTEXT.name();
 
+  // Kerberos configs
+  public static final String USER_PRINCIPAL_CONFIG = "kerberos.user.principal";
```
dosvath, line 287:
```suggestion
  public static final String KERBEROS_PRINCIPAL_CONFIG = "kerberos.user.principal";
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -276,16 +277,30 @@
...
+  private static final String USER_PRINCIPAL_DISPLAY = "Kerberos User Principal";
+  private static final String USER_PRINCIPAL_DOC = "The Kerberos user principal the connector may"
+      + " use to authenticate with Kerberos.";
+  private static final String USER_PRINCIPAL_DEFAULT = null;
+
+  public static final String KEYTAB_FILE_PATH_CONFIG = "kerberos.keytab.path";
```
dosvath, line 293:
```suggestion
  public static final String KERBEROS_KEYTAB_PATH_CONFIG = "kerberos.keytab.path";
```

src/main/java/io/confluent/connect/elasticsearch/ClientConfigCallbackHandler.java
levzem: it identifies the SPNEGO auth mechanis,

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -749,6 +796,14 @@ public boolean ignoreSchema() {
     return new HashSet<>(getList(IGNORE_SCHEMA_TOPICS_CONFIG));
   }
 
+  public String kerberosUserPrincipal() {
+    return getString(USER_PRINCIPAL_CONFIG);
+  }
+
+  public String keytabPath() {
```
dosvath, line 803:
```suggestion
  public String kerberosKeytabPath() {
```

dosvath (commented):
Final question, how does the ES client know which KDC URL to connect to? Does it look in the `/etc/krb5.conf` file?

levzem:
> Final question, how does the ES client know which KDC URL to connect to? Does it look in the `/etc/krb5.conf` file?

@dosvath yes, it looks at the user principal realm and the `/etc/krb5.conf` file

dosvath (approved):
Code LGTM, but we may consider at least one passing manual test.

PR #488, opened 2021-01-28
MINOR: Fix semver test to correctly accept qualifiers in the end

kkonstantine:
## Problem
We need to be more lax on the qualifiers we accept in versions

## Solution
Fix regex

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorTest.java
+++ b/src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorTest.java
@@ -77,6 +77,7 @@ public class ElasticsearchSinkConnectorTest {
     assertNotNull(connector.version());
     assertFalse(connector.version().equals("0.0.0.0"));
     assertFalse(connector.version().equals("unknown"));
-    assertTrue(connector.version().matches("^(\\d+\\.)?(\\d+\\.)?(\\*|\\d+)(-\\w+)?$"));
+    // Match semver with potentially a qualifier in the end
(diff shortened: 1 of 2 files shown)


Prathibha-m (approved):
LGTM, thanks @kkonstantine !

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorTest.java
```diff
@@ -77,6 +77,7 @@ public void testVersion() {
     assertNotNull(connector.version());
     assertFalse(connector.version().equals("0.0.0.0"));
     assertFalse(connector.version().equals("unknown"));
-    assertTrue(connector.version().matches("^(\\d+\\.)?(\\d+\\.)?(\\*|\\d+)(-\\w+)?$"));
+    // Match semver with potentially a qualifier in the end
+    assertTrue(connector.version().matches("^(\\d+\\.){2}?(\\*|\\d+)(-.*)?$"));
```
Prathibha-m, line 81: not a deal breaker, but we could also restrict the qualifier to only be `rc*` I suppose?

kkonstantine: I'd like to, but then there's the case insensitive check (RC and rc) and others. All that takes us a bit away from semver overall ... 
So decide to be slightly more liberal so we don't have to go back and change this minor test too often. (the previous check assumed `-` and a word. Now we check for `-`).  We can return to that in a follow up if we want to be more strict.

Issue #487, opened 2021-01-27
slow calls to elastic, and wondering about need for the -- clock.sleep(TimeUnit.SECONDS.toMillis(1)) 


nbuesing:
I am having some serious performance issues with the latest elastic sink but it is on some new code I am working on, so I am not sure if it is related to the latest version or not, but based on the code change I did to verify, I believe it is.

In the ElasticsearchClient there are two spots where clock.sleep of one second is performed.  Looks like while there are attempts to increase buffer size, there is a hardcoding 1 second sleep.  

Locally I changed the TimeUnit from Seconds to Milliseconds and the code is being performant.  Oddly, I do not always see this behavior, but I think it could be related to the amount of updates vs inserts (in addition to batch sizes). 
  

```
    if (docIdToRecord != null) {
      // make sure that only unique records are being sent in a batch
      // every request that is flushed and succeeds triggers a callback that removes it from the map
      while (docIdToRecord.containsKey(request.id())) {
        flush();
        clock.sleep(TimeUnit.MILLISECONDS.toMillis(1));
      }
    }

    // wait for internal buffer to be less than max.buffered.records configuration
    long maxWaitTime = clock.milliseconds() + config.flushTimeoutMs();
    while (numRecords.get() >= config.maxBufferedRecords()) {
      clock.sleep(TimeUnit.MILLISECONDS.toMillis(1));
      if (clock.milliseconds() > maxWaitTime) {
        throw new ConnectException(
            String.format(
                "Could not make space in the internal buffer fast enough. Consider increasing %s"
                    + " or %s.",
                FLUSH_TIMEOUT_MS_CONFIG,
                MAX_BUFFERED_RECORDS_CONFIG
            )
        );
      }
    }
```
CC-12445: Replace the JestClient with the ElasticsearchRestClient
https://github.com/confluentinc/kafka-connect-elasticsearch/pull/468

levzem:
@nbuesing I agree, this seems to be a concern for data sets with frequent updates to documents

I have published a PR #490 that address this issue and will follow up with a bug fix release shortly

nbuesing:
in comparing to 10.0.2 which I am told would be compatible with connect cluster 6.0.x, there is no issue.  Also it looks as though the flushing to avoid duplicates in a bulk isn't done "1 at a time", 

[...]

PR #490, opened 2021-02-01
CCMSG-796: fix poor performance for lots of update requests

levzem:

## Problem
data sets that have frequent updates will result in slow performance because of long wait times

## Solution
reduce the wait times to 10ms to fix performance bottlenecking

remove the unique constraint by relying on `BulkItemResponse::getItemId` which returns the order of the request in the bulk request, which we can use to map the request to the `SinkRecord` it came from

- [ ] yes
- [x] no

## Test Strategy
modified a test to upsert multiple records with the same key
existing reporter tests cover the rest of the test cases

PERF TEST RESULTS:

configs
```
{
  "task.max": 1,
  "linger.ms": 1000,
  "batch.size": 2000,
  "max.buffered.records": 20000,
  "max.in.flight.requests": 5,
}
```

v10.0.2 with the same key for every single record
Screen Shot 2021-02-08 at 4 24 40 PM

a steady 6.7K records/s

this PR:
Screen Shot 2021-02-08 at 4 02 20 PM

the first rate is a steady 32K records/s (the dip is from a rebalance) with unique keys
the second rate is a steady 6.7K records/s with identical records which is what v10.0.2 also had, therefore this PR addresses the performance regression

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests - performance testing

## Release Plan

 
backporting to `11.0.x` where this was introduced

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -64,6 +64,7 @@ public class ElasticsearchClient {
 
   private static final Logger log = LoggerFactory.getLogger(ElasticsearchClient.class);
 
+  private static final long WAIT_TIME = TimeUnit.MILLISECONDS.toMillis(10);
   private static final String RESOURCE_ALREADY_EXISTS_EXCEPTION =
       "resource_already_exists_exception";
   private static final String VERSION_CONFLICT_EXCEPTION = "version_conflict_engine_exception";
@@ -255,14 +256,14 @@ public class ElasticsearchClient {
       // every request that is flushed and succeeds triggers a callback that removes it from the map
       while (docIdToRecord.containsKey(request.id())) {
         flush();
-        clock.sleep(TimeUnit.SECONDS.toMillis(1));
+        clock.sleep(WAIT_TIME);
       }
     }
 
     // wait for internal buffer to be less than max.buffered.records configuration
     long maxWaitTime = clock.milliseconds() + config.flushTimeoutMs();
     while (numRecords.get() >= config.maxBufferedRecords()) {
-      clock.sleep(TimeUnit.SECONDS.toMillis(1));
+      clock.sleep(WAIT_TIME);
       if (clock.milliseconds() > maxWaitTime) {
         throw new ConnectException(
             String.format(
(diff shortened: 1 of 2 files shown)


dosvath (approved):
LGTM, thanks @levzem

dosvath (changes requested):
One final question, what is the purpose of the sleep time here, would 1MS work as well? [follow up comment on issue](https://github.com/confluentinc/kafka-connect-elasticsearch/issues/487#issuecomment-771189932)

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -250,19 +253,10 @@ public void index(SinkRecord record, DocWriteRequest<?> request) {
...
-        flush();
-        clock.sleep(TimeUnit.SECONDS.toMillis(1));
-      }
-    }
-
     // wait for internal buffer to be less than max.buffered.records configuration
     long maxWaitTime = clock.milliseconds() + config.flushTimeoutMs();
     while (numRecords.get() >= config.maxBufferedRecords()) {
-      clock.sleep(TimeUnit.SECONDS.toMillis(1));
+      clock.sleep(WAIT_TIME);
```
Prathibha-m, line 259: Is there a reason why this wait time was chosen? Would this value change if there were x times more requests?
I suppose perf tests will give more backing to what the value would be 

Could be overkill but this made me think of setting a [rate limiter](https://guava.dev/releases/22.0/api/docs/index.html?com/google/common/util/concurrent/RateLimiter.html) on `addToRequestToRecordMap`. Suppose we hit the rate limit, we increase the rate limit as well as the wait time accordingly to better performance

dosvath (approved):

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
levzem: we actually don't want to rate limit, we want to run as fast as possible and only wait if we dont have enough buffer space

Prathibha-m: I see 👍🏽

Prathibha-m (approved):

PR #498, opened 2021-02-19
CCMSG-201: add connection configs validator

levzem:

## Problem
we are missing a validator whether the ES service is available

## Solution
add a validator by pinging ES

- [ ] yes
- [x] no

## Test Strategy
add unit tests testing the validator, existing ITs also run the added code

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport to `11.0.x` where this validaator should be added

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/Validator.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/Validator.java
@@ -17,17 +17,25 @@
 package io.confluent.connect.elasticsearch;
 
 import io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.SecurityProtocol;
+import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import org.apache.http.HttpHost;
 import org.apache.kafka.common.config.Config;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.config.ConfigValue;
 import org.apache.kafka.common.config.SslConfigs;
+import org.elasticsearch.client.RequestOptions;
+import org.elasticsearch.client.RestClient;
+import org.elasticsearch.client.RestHighLevelClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.BATCH_SIZE_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.CONNECTION_PASSWORD_CONFIG;
+import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.CONNECTION_URL_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.CONNECTION_USERNAME_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.FLUSH_TIMEOUT_MS_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.IGNORE_KEY_CONFIG;
@@ -40,6 +48,7 @@ import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfi
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.MAX_IN_FLIGHT_REQUESTS_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.PROXY_HOST_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.PROXY_PASSWORD_CONFIG;
+import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.PROXY_PORT_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.PROXY_USERNAME_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.SECURITY_PROTOCOL_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.SSL_CONFIG_PREFIX;
@@ -47,9 +56,12 @@ import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfi
 
 public class Validator {
 
+  private static final Logger log = LoggerFactory.getLogger(Validator.class);
+
   private ElasticsearchSinkConnectorConfig config;
   private Map<String, ConfigValue> values;
   private List<ConfigValue> validations;
+  private RestHighLevelClient client;
 
   public Validator(Map<String, String> props) {
     try {
@@ -62,10 +74,20 @@ public class Validator {
     values = validations.stream().collect(Collectors.toMap(ConfigValue::name, Function.identity()));
   }
 
+  // Exposed for testing
+  protected Validator(Map<String, String> props, RestHighLevelClient client) {
+    this(props);
+    this.client = client;
+  }
+
   public Config validate() {
     if (config == null) {
       // individual configs are invalid, no point in validating combinations
       return new Config(validations);
+    }
+
+    if (client == null) {
+      client = createClient();
     }
 
     validateCredentials();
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -62,12 +74,22 @@ public Validator(Map<String, String> props) {
...
+  }
+
   public Config validate() {
     if (config == null) {
       // individual configs are invalid, no point in validating combinations
       return new Config(validations);
     }
 
+    if (client == null) {
+      client = createClient();
```
avocader, line 90: Should it be made `final` and guarded by a `synchronized` block?

avocader: Ok, I see a logic below, and not sure why we even have the `client` as a class field vs just have it as a local variable, since we are not reusing it between calls?

levzem: for testing. there is no clean way that I have found to mock the client, open to suggestions

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -238,7 +272,91 @@ private void validateSsl() {
...
+            CONNECTION_PASSWORD_CONFIG,
+            e.getMessage()
+        );
+        addErrorMessage(CONNECTION_USERNAME_CONFIG, errorMessage);
+        addErrorMessage(CONNECTION_PASSWORD_CONFIG, errorMessage);
+      }
+
+      if (config.isSslEnabled()) {
+        errorMessage = String.format(
+            "Could not connect to Elasticsearch. Check your SSL settings. Error message: %s",
```
avocader, line 300: nit: which `SSL` configs? Give the user a prefix maybe, which configs should he check?

levzem: this error message is added for every single config that it affects below

levzem: https://github.com/confluentinc/kafka-connect-elasticsearch/pull/498/files/34da65886dbbe4dea395a9ec9f7584eb2c825ca2#diff-01aa6b506fb570054df3884b36841d52284ab0a07ac84ce1767da88987b6a2c8R304

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -238,7 +272,91 @@ private void validateSsl() {
...
+            e.getMessage()
+        );
+
+        addErrorMessage(SECURITY_PROTOCOL_CONFIG, errorMessage);
+      }
+
+      if (config.isKerberosEnabled()) {
+        errorMessage = String.format(
+            "Could not connect to Elasticsearch. Check your Kerberos settings. Error message: %s",
```
avocader, line 309: nit: same as for SSL error, can we give more guidance to the user, which configs should she check?

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -238,7 +272,91 @@ private void validateSsl() {
...
+            e.getMessage()
+        );
+
+        addErrorMessage(KERBEROS_PRINCIPAL_CONFIG, errorMessage);
+        addErrorMessage(KERBEROS_KEYTAB_PATH_CONFIG, errorMessage);
+      }
+
+      if (config.isBasicProxyConfigured()) {
```
avocader, line 317: If we can't ping the ES, doesn't necessarily mean it's because of proxy settings. Maybe we can do some heuristics here, e.g. inspect exception, whether it has HTTP 407 - that means proxy authentication had failed.

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -238,7 +272,91 @@ private void validateSsl() {
...
+                    .map(HttpHost::create)
+                    .collect(Collectors.toList())
+                    .toArray(new HttpHost[config.connectionUrls().size()])
+            )
+            .setHttpClientConfigCallback(configCallbackHandler)
+            .setRequestConfigCallback(configCallbackHandler)
+    );
+  }
+
+  private boolean hasErrors() {
```
avocader, line 353: Don't we have a similar existing method already? If not, can be a good candidate to be generalized and moved to a common library.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorTest.java
```diff
@@ -62,16 +62,6 @@ public void shouldStartAndStop() {
     connector.stop();
   }
 
-
-  @Test
```
avocader, line 66: Why are we dropping this test? Can we do an incremental change instead?

src/test/java/io/confluent/connect/elasticsearch/ValidatorTest.java
```diff
@@ -313,12 +323,29 @@ public void testValidSsl() {
...
+  @Test
+  public void testValidConnection() {
+    validator = new Validator(props, mockClient);
+
+    Config result = validator.validate();
+    assertNoErrors(result);
+  }
+
+  @Test
+  public void testInvalidConnection() throws IOException {
```
avocader, line 341: Add another test, that asserts the validator's behavior when `client.ping()` throws an exception?

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -238,7 +272,91 @@ private void validateSsl() {
     }
   }
 
+  private void validateConnection() {
+    try {
+      if (!client.ping(RequestOptions.DEFAULT)) {
+        throw new IOException("Cannot connect to Elasticsearch.");
```
avocader, line 278: Is it worth creating a relatively expensive expectation object(mostly due to stack trace allocation) and the catch it immediately? Maybe just factor it out in a separate method, e.g?
```
try{
..
if (!client.ping(RequestOptions.DEFAULT)) {
handleFailedPing("Failed to send a ping request to Elasticsearch");
} catch (IOException e) {
handleFailedPing(e.getMessage());
}
```

src/main/java/io/confluent/connect/elasticsearch/Validator.java
avocader:
```
protected Validator(Map<String, String> props, ClientFactory clientFactory) {}
```

```
private RestHighLevelClient createClient() {
if this.clientFactory != null {
return clientFactory.create();
} else {
//use your default factory
}}
```

```
interface ClientFactory {
RestHighLevelClient create();
}
```

avocader: Then in your test, just pass a lambda function that creates a mocked client to `Validator` constructor.

src/main/java/io/confluent/connect/elasticsearch/Validator.java
levzem: same comment as above, this is added to every kerberos config. hence resolving

src/main/java/io/confluent/connect/elasticsearch/Validator.java
levzem: `ping` only returns `boolean`. We don't actually get exposed to the response codes so unfortunately we can't determine the exact root cause.

src/main/java/io/confluent/connect/elasticsearch/Validator.java
levzem: that would have to be added in AK
another option would be to add to internal libraries but that wouldn't be available to any OSS connectors

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorTest.java
levzem: again, not sure how to mock the client here and this does not test much

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -68,18 +86,26 @@ public Config validate() {
...
+      validateMaxBufferedRecords();
+      validateProxy();
+      validateSsl();
+
+      if (!hasErrors()) {
+        // no point if previous configs are invalid
+        validateConnection(client);
+      }
+    } catch (IOException e) {
+      log.warn("Closing the client failed ");
```
avocader, line 103: nit: log `e` as well?

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorTest.java
avocader: We should be able to add a test since you can pass a `clientFactory` now.

avocader (approved):
Added few minor comments, otherwise LGTM.

Issue #503, opened 2021-03-03
Add create action type for the bulk request


danprueitt:
Add the "create" action type for the bulk request as found in the Elastic Docs here:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html

The reasoning for this change is I would like to be able to Sink directly into a data stream, however, data streams only allow the create action. `only write ops with an op_type of create are allowed in data streams`.

I believe this would involve adding a config for the option and including it in  `DataConverter.java` `convertRecord` method by either adding a new case or adding a boolean setting onto the `new IndexRequest().create(<setting>)`

This would be useful to be an option in the Confluent Cloud Hosted version of the connector too as that is what I am currently using.

danprueitt:
Seems there is already a pull request with a proposed solution: #403 as well as one I started to work on #504

danprueitt:
This seems to be a duplicate  of this issue #369 and can be closed as duplicate

PR #403, opened 2020-04-21
#396 new configuration option write.method "create"

graham2071:
Add the possibility to set `write.method` configuration option to  `"create"`.

If `write.method` is set to `create`, a new document is created if the id does not already exist. If a document already exist with this id, it is NOT updated. The implementation uses `op_type=create` elasticsearch parameter.

PR #504, opened 2021-03-04
Add create action option for bulk requests

danprueitt:
## Problem
As mentioned in #503 I would like to be able to sink directly to data streams. Data streams only work with the "create" action on bulk requests. Currently, the kafka-connect sink only supports "index" and "update".

## Solution
Added an option to specify if you would like to replace existing documents if they exist or not with the same id. The determines if to use the create action or index action in elastic.


danprueitt:
Seems there is already a pull request for a similar solution: https://github.com/confluentinc/kafka-connect-elasticsearch/pull/403.

PR #506, opened 2021-03-10
CCMSG-835, CCMSG-834: Update outdated commons-codec to fix CVE

BlueEldur:
## Problem
Shown by the mvn dependency:tree command, the httpclient dependency drags in commons-codec to 1.11, which is marked by twistlock as a vulnerability

## Solution
Exclude the dependency from httpclient, and bump to the latest version (1.15).

- [x] yes
- [ ] no

##### If yes, where?
It also applies to 5.5.x, so but it should be fixed in the backport of this PR.

## Test Strategy
mvn verify and mvn integration-test to make sure nothing broke

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
targeting 5.4.x

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
             <artifactId>httpclient</artifactId>
             <version>4.5.13</version>
         </dependency>
-        <!-- force commons-codec version to address CVE CCMSG-835 -->
+        <!-- Override httpClient transitive dependency commons-codec version to address CVE CCMSG-835 -->
         <dependency>
             <groupId>commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
             <artifactId>httpclient</artifactId>
             <version>4.5.13</version>
         </dependency>
-        <!-- Override httpClient transitive dependency commons-codec version to address CVE CCMSG-835 -->
+        <!-- Force commons-codec (httpclient transitive dependency) version to address CVE CCMSG-835 -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -77,6 +77,18 @@
...
+                    <groupId>commons-codec</groupId>
+                    <artifactId>commons-codec</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <!-- force commons-codec version to address CVE CCMSG-835 -->
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
+            <version>1.15</version>
```
levzem, line 91: usually if you pin the version, you dont need to exclude

pom.xml
```diff
@@ -78,6 +78,12 @@
             <artifactId>httpclient</artifactId>
             <version>4.5.13</version>
         </dependency>
+        <!-- force commons-codec version to address CVE CCMSG-835 -->
```
kkonstantine, line 81: let's mention that this dependency is a transitive dependency of `httpclient` above. Unfortunately there's no recent enough version of `httpclient` that would allow us to use and upgrade the codec without having to explicitly list it as a dependency. Thus, your fix makes sense (as long as it doesn't break the `httpclient`) but we should consider for removal once `httpclient` gets upgraded in the future. So, a comment would help here.

kkonstantine (approved):
LGTM with a note

PR #524, opened 2021-05-13
Address CVEs by upgrading commons-codec to 1.15

ilanjiR:
## Problem
Commons-codec had unresolved CVE's

## Solution
Upgrade commons-codec to 1.15
Upgrade ES client to latest bugfix version  

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <parent>
         <groupId>io.confluent</groupId>
         <artifactId>common</artifactId>
-        <version>5.5.3</version>
+        <version>5.3.5-SNAPSHOT</version>
     </parent>
 
     <groupId>io.confluent</groupId>
@@ -39,7 +39,7 @@
     </scm>
 
     <properties>
-        <es.version>2.4.1</es.version>
+        <es.version>2.4.4</es.version>
         <lucene.version>5.5.2</lucene.version>
         <hamcrest.version>1.3</hamcrest.version>
         <mockito.version>2.28.2</mockito.version>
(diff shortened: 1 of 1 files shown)


ilanjiR:
PR is not ready yet. Just needed to check if the changes resolved the CVE or not

pom.xml
```diff
@@ -176,6 +182,25 @@
             <version>${test.containers.version}</version>
             <scope>test</scope>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.apache.zookeeper/zookeeper -->
+        <dependency>
+            <groupId>org.apache.zookeeper</groupId>
+            <artifactId>zookeeper</artifactId>
+            <version>3.4.14</version>
```
kkonstantine, line 189: we should get these fixes from common probably

ilanjiR: tried to look for an alternative, but it seems like it's just not possible to exclude a dependency that's inherited from the parent. For now, I think our only choice is to pin the version to a clean one, then update common down the road when it's fixed upstream

kkonstantine: On these release branches common and ES connector are released together. There's no point to fix here only while common is having the affected dependency. 
https://github.com/confluentinc/common/pull/335 
or 
https://github.com/confluentinc/common/pull/331
Which ever lands first.

pom.xml
```diff
@@ -7,7 +7,7 @@
     <parent>
         <groupId>io.confluent</groupId>
         <artifactId>common</artifactId>
-        <version>5.2.5-SNAPSHOT</version>
+        <version>5.3.5-SNAPSHOT</version>
```
kkonstantine, line 10: This is not a valid change here.

pom.xml
```diff
@@ -39,7 +39,7 @@
     </scm>
 
     <properties>
-        <es.version>2.4.1</es.version>
+        <es.version>2.4.4</es.version>
```
kkonstantine, line 42: I see this affects only tests. 👍

pom.xml
```diff
@@ -78,6 +78,11 @@
             <artifactId>httpclient</artifactId>
             <version>4.5.13</version>
         </dependency>
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
+            <version>1.15</version>
```
kkonstantine, line 84: Let's get this version number in the properties section.

kkonstantine (approved):
One last comment but otherwise LGTM

PR #525, opened 2021-05-13
CCMSG-1028: Validate Elasticsearch version to be compatible with data stream.

amuamushu:
## Problem
Data streams is available on Elasticsearch versions 7.9+. If the user wants to write to data stream but have an invalid Elasticsearch version, we should prevent them from creating the connector.

## Solution
Add validation to the Validator class that checks the Elasticsearch version.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

Will be release as part of the Elasticsearch connector data stream feature.


src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -565,6 +566,102 @@ public void testNoVersionConflict() throws Exception {
...
+    container.close();
+    container = ElasticsearchContainer.fromSystemProperties();
+    container.start();
+  }
+
+  @Test
+  public void testCompatibleESVersionDataStreamNotSet() throws Exception {
+    container.close();
+    String ESVersion = "7.9.3";
+    container = new ElasticsearchContainer(DEFAULT_DOCKER_IMAGE_NAME + ":" + ESVersion);
```
levzem, line 603: let's add a constructor that just takes the version number and constructs the docker image name inside

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -102,6 +105,7 @@ public Config validate() {
       validateMaxBufferedRecords();
       validateProxy();
       validateSsl();
+      validateVersion(client);
```
levzem, line 108: i would validate the version after validating the connection because if we can't establish a connection theres no point in checking the version

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -306,6 +310,28 @@ private void validateSsl() {
...
+  private void validateVersion(RestHighLevelClient client) {
+    MainResponse response;
+    try {
+      response = client.info(RequestOptions.DEFAULT);
+    } catch (IOException e) {
+      // Same error messages as from validating the connection.
+      return;
+    }
+    String versionNumber = response.getVersion().getNumber();
+    if (config.isDataStream() && versionNumber.compareTo(DATA_STREAM_COMPATIBLE_ES_VERSION) < 0) {
```
levzem, line 322: lets only run this validation if it is a data stream so
```
if (!config.isDataStream()) {
   return;
}
```

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -565,6 +566,102 @@ public void testNoVersionConflict() throws Exception {
...
+  @Test
+  public void testIncompatibleESVersionDataStreamNotSet() throws Exception {
+    container.close();
+    container = new ElasticsearchContainer(DEFAULT_DOCKER_IMAGE_NAME + ":" + "7.8.1");
+    container.start();
+
+    String address = container.getConnectionUrl().replace(container.getContainerIpAddress(), container.hostMachineIpAddress());
+    props.put(CONNECTION_URL_CONFIG, address);
+    props.put(CONNECTION_USERNAME_CONFIG, "elastic");
+    props.put(CONNECTION_PASSWORD_CONFIG, ELASTIC_PASSWORD);
```
levzem, line 578: shouldn't be necessary

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -565,6 +566,102 @@ public void testNoVersionConflict() throws Exception {
     client2.close();
   }
 
+  @Test
+  public void testIncompatibleESVersionDataStreamNotSet() throws Exception {
+    container.close();
+    container = new ElasticsearchContainer(DEFAULT_DOCKER_IMAGE_NAME + ":" + "7.8.1");
+    container.start();
+
+    String address = container.getConnectionUrl().replace(container.getContainerIpAddress(), container.hostMachineIpAddress());
```
levzem, line 575:
```suggestion
    String address = container.getConnectionUrl();
```

should be sufficient, we only do that for the SSL tests

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -565,6 +566,102 @@ public void testNoVersionConflict() throws Exception {
...
+    props.put(CONNECTION_URL_CONFIG, address);
+    props.put(CONNECTION_USERNAME_CONFIG, "elastic");
+    props.put(CONNECTION_PASSWORD_CONFIG, ELASTIC_PASSWORD);
+    config = new ElasticsearchSinkConnectorConfig(props);
+    converter = new DataConverter(config);
+
+    ElasticsearchClient client = new ElasticsearchClient(config, null);
+    helperClient = new ElasticsearchHelperClient(address, config);
```
levzem, line 583: we can make this a separate `setupFromContainer` method that does this since its repetitive and is done in `setup()` as well

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -565,6 +566,102 @@ public void testNoVersionConflict() throws Exception {
     client2.close();
   }
 
+  @Test
+  public void testIncompatibleESVersionDataStreamNotSet() throws Exception {
```
levzem, line 570: these tests never call the validator, they should be done in `ValidatorTest` or in ITs

amuamushu: I have a set of tests like them in `ValidatorTest` that checks for the error message. The tests in this class were meant to check that records could still be written. I'll move them to the ITs.

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -306,6 +310,31 @@ private void validateSsl() {
...
+      response = client.info(RequestOptions.DEFAULT);
+    } catch (IOException e) {
+      // Same error messages as from validating the connection.
+      return;
+    }
+    String versionNumber = response.getVersion().getNumber();
+    if (versionNumber.compareTo(DATA_STREAM_COMPATIBLE_ES_VERSION) < 0) {
+      String errorMessage = String.format(
+          "Given Elasticsearch version %s is not compatible with data streams. Elasticsearch"
+              + "version must be at least %s",
```
levzem, line 328:
```suggestion
          "Elasticsearch version %s is not compatible with data streams. Elasticsearch"
              + "version must be at least %s.",
```

src/test/java/io/confluent/connect/elasticsearch/ValidatorTest.java
```diff
@@ -401,6 +407,48 @@ public void testInvalidConnectionThrows() throws IOException {
...
+    Config result = validator.validate();
+    assertNoErrors(result);
+  }
+
+  @Test
+  public void testCompatibleVersionDataStreamSet() {
+    props.put(DATA_STREAM_DATASET_CONFIG, "a_valid_dataset");
+    props.put(DATA_STREAM_TYPE_CONFIG, "logs");
+    validator = new Validator(props, () -> mockClient);
+    when(mockInfoResponse.getVersion().getNumber()).thenReturn("7.9.0");
```
levzem, line 446: can you write a for loop that tests this against a variety of versions i.e `7.9.12` `10.0.0` `7.10.0` `7.10.12` and so forth

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
```diff
@@ -171,4 +166,51 @@ public void testUpsert() throws Exception {
...
+    runSimpleTest(props);
+
+    helperClient = null;
+    container.close();
+    container = ElasticsearchContainer.fromSystemProperties();
+    container.start();
+  }
+
+  @Test
+  public void testCompatibleESVersionDataStreamSet() throws Exception {
```
levzem, line 201: happy path will test the validator so we dont need it

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
```diff
@@ -171,4 +166,51 @@ public void testUpsert() throws Exception {
...
+    runSimpleTest(props);
+
+    helperClient = null;
+    container.close();
+    container = ElasticsearchContainer.fromSystemProperties();
+    container.start();
+  }
+
+  @Test
+  public void testCompatibleESVersionDataStreamNotSet() throws Exception {
```
levzem, line 186: other tests will test this case too so we don't need it

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
```diff
@@ -171,4 +166,51 @@ public void testUpsert() throws Exception {
...
   }
+
+  @Test
+  public void testIncompatibleESVersionDataStreamNotSet() throws Exception {
+    container.close();
+    container = ElasticsearchContainer.withESVersion("7.8.1");
+    container.start();
+    setupFromContainer();
+
+    runSimpleTest(props);
```
levzem, line 177: the whole purpose of this test is to make sure the validator fails, why don't you spin up the container in ValidatorTest instead to do this test

amuamushu: Resolved in chat. Renamed the test method instead.

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -306,6 +310,56 @@ private void validateSsl() {
...
+
+  /**
+   * Compares <code>versionNumber</code> to <code>compatibleVersion</code>. Ignores
+   * the patch version number.
+   *
+   * @return  a negative integer, zero, or a positive integer if
+   *          <code>versionNumber</code> is less than, equal to, or greater
+   *          than <code>compatibleVersion</code>.
+   */
+  public int compareVersions(String versionNumber, String compatibleVersion) {
```
levzem, line 346: ope, did something break when you added more tests?

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -306,6 +310,56 @@ private void validateSsl() {
...
+    String[] versionSplit = versionNumber.split("\\.");
+    String[] compatibleSplit = compatibleVersion.split("\\.");
+
+    for (int i = 0; i < 2; i++) {
```
levzem, line 350:
```suggestion
    for (int i = 0; i < Math.min(versionSplit.length, compatibleSplit.length); i++) {
```

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -306,6 +310,56 @@ private void validateSsl() {
...
+      int comparison = Integer.compare(
+          Integer.parseInt(versionSplit[i]),
+          Integer.parseInt(compatibleSplit[i])
+      );
+      if (comparison == 0) {
+        continue;
+      }
+      return comparison;
+    }
+    return 0;
```
levzem, line 360:
```suggestion
    return versionSplit.length - compatibleSplit.length;
```

src/main/java/io/confluent/connect/elasticsearch/Validator.java
levzem: makes this more generic

src/main/java/io/confluent/connect/elasticsearch/Validator.java
levzem, line 346:
```suggestion
  private int compareVersions(String versionNumber, String compatibleVersion) {
```

src/test/java/io/confluent/connect/elasticsearch/ValidatorTest.java
```diff
@@ -401,6 +407,57 @@ public void testInvalidConnectionThrows() throws IOException {
...
+      assertNoErrors(result);
+    }
+  }
+
+  @Test
+  public void testCompatibleVersionDataStreamSet() {
+    props.put(DATA_STREAM_DATASET_CONFIG, "a_valid_dataset");
+    props.put(DATA_STREAM_TYPE_CONFIG, "logs");
+    validator = new Validator(props, () -> mockClient);
+    String[] compatibleESVersions = {"7.9.0", "7.9.3", "7.9.3-amd64", "7.10.0", "7.10.2", "7.11.0", "7.11.2", "7.12.0", "7.12.1"};
```
levzem, line 452: lets add test cases for `8.0.0`, `10.10.10` `10.1.10` `10.1.1` and `8.10.10`

levzem (commented):
overall lgtm, just a few comments left

src/main/java/io/confluent/connect/elasticsearch/Validator.java
amuamushu: Oops, forgot to change it back after I played around with it. Thanks!

amuamushu: Yes, regular string comparison sadly won't work when comparing single digit against multi-digit version numbers.

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -306,6 +310,55 @@ private void validateSsl() {
...
+   * @return  a negative integer, zero, or a positive integer if
+   *          <code>versionNumber</code> is less than, equal to, or greater
+   *          than <code>compatibleVersion</code>.
+   */
+  private int compareVersions(String versionNumber, String compatibleVersion) {
+    String[] versionSplit = versionNumber.split("\\.");
+    String[] compatibleSplit = compatibleVersion.split("\\.");
+
+    for (int i = 0; i < Math.min(versionSplit.length, compatibleSplit.length); i++) {
+      String beforeSuffix = versionSplit[i].split("-")[0];
```
levzem, line 350: both versions could have a suffix

levzem (approved):
LGTM

PR #530, opened 2021-05-22
CCMSG-1116: Prevent adding external versioning to data stream.

amuamushu:
## Problem
When writing records to regular elasticsearch indices, external versioning ensures only the latest record with the same key ends up in elasticsearch. However, external versioning is not supported for data stream as mentioned in [Elasticsearch's index API documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#docs-index_). 

Currently, external versioning is set for records with the same key regardless of whether data stream is enabled. This causes the Bulk Request to fail with the error
`org.elasticsearch.action.ActionRequestValidationException: Validation Failed: 1: create operations only support internal versioning. use index instead`

## Solution
Prevent updating the versioning type and version number if data streams is enabled. If messages with the same key exist in a topic, only the oldest message gets written to data stream.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

Releasing as part of the whole data stream feature.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -219,7 +219,7 @@ public class DataConverter {
       DocWriteRequest<?> request,
       SinkRecord record
   ) {
-    if (!config.shouldIgnoreKey(record.topic()) && !config.isDataStream()) {
+    if (!config.isDataStream() && !config.shouldIgnoreKey(record.topic())) {
       request.versionType(VersionType.EXTERNAL);
       request.version(record.kafkaOffset());
     }
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -219,7 +219,7 @@ public class DataConverter {
       DocWriteRequest<?> request,
       SinkRecord record
   ) {
-    if (!config.shouldIgnoreKey(record.topic())) {
+    if (!config.shouldIgnoreKey(record.topic()) && !config.isDataStream()) {
       request.versionType(VersionType.EXTERNAL);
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -219,7 +219,7 @@ private String maybeAddTimestamp(String payload, long timestamp) {
       DocWriteRequest<?> request,
       SinkRecord record
   ) {
-    if (!config.shouldIgnoreKey(record.topic())) {
+    if (!config.shouldIgnoreKey(record.topic()) && !config.isDataStream()) {
```
levzem, line 222: nit: let's do `isDataStream` first

src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
```diff
@@ -422,4 +424,19 @@ public void testDoNotInjectPayloadTimestampIfAlreadyExists() {
 
     assertEquals(timestamp, actualRecord.sourceAsMap().get(TIMESTAMP_FIELD));
   }
+
+  @Test
+  public void testDoNotAddExternalVersioningIfDataStream() {
```
levzem, line 429: this would throw an exception without the changes right?

levzem (approved):
LGTM

src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
amuamushu: Without the changes, the test would fail rather than throw an exception because the version type would be set to external. In this case, I don't think an exception would be thrown until the bulk request is sent.

PR #531, opened 2021-05-24
CCMSG-1115: Add a dropdown menu to setting the data stream type config.

amuamushu:
## Problem
In confluent platform, there are specific values a user can choose from for setting the `data stream type`. However, currently, this configuration is set by filling in the blank which requires users to review the config description to remember what the values are and which configs have specific values.

## Solution
Add a recommender to each of those configurations such that it recommends the specific values.

- [ ] yes
- [x] no

## Test Strategy
Added the recommenders and checked in confluent platform that those configs had the correct recommenders.

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

Releasing as part of the data stream feature.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -18,7 +18,6 @@ package io.confluent.connect.elasticsearch;
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -31,7 +30,6 @@ import org.apache.kafka.common.config.ConfigDef.Validator;
 import org.apache.kafka.common.config.ConfigDef.Width;
 
 import java.util.Map;
-import java.util.stream.Collectors;
 
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.config.types.Password;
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -159,6 +159,11 @@
             <artifactId>confluent-log4j</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>connect-utils</artifactId>
+            <version>0.3.1</version>
+        </dependency>
```
levzem, line 166: this is a propietary package which we cannot use in public repos

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -338,6 +340,17 @@
   private static final String SSL_GROUP = "Security";
   private static final String KERBEROS_GROUP = "Kerberos";
 
+  private static final Recommender BEHAVIOR_ON_MALFORMED_DOC_RECOMMENDER =
+      Recommenders.enumValues(BehaviorOnMalformedDoc.class);
```
levzem, line 344: nice for fixing it for other values too!

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -458,7 +471,8 @@ private static void addConnectorConfigs(ConfigDef configDef) {
             CONNECTOR_GROUP,
             ++order,
             Width.SHORT,
-            DATA_STREAM_TYPE_DISPLAY
+            DATA_STREAM_TYPE_DISPLAY,
+            DATA_STREAM_TYPE_RECOMMENDER
```
levzem, line 475: we don't need to put this in static variables, we can just do it inline
```suggestion
            Recommenders.enumValues(...)
```

amuamushu:
This PR is currently on hold until #532 is merged.

levzem (approved):

PR #532, opened 2021-05-24
CCMSG-1119: Add drop down menus for the existing enum configs.

amuamushu:
## Problem
The enum configs are set by filling in the blank which requires users to review the config descriptions for what the values are and which configs have specific values.

## Solution
Add a recommender to each of those configurations such that it recommends the specific values.

- [x] yes
- [ ] no

##### If yes, where?
The `data.stream.type` config could also use a recommender. (#531)

## Test Strategy
After adding the recommenders, checked in confluent platform that those configs had the correct recommenders and that the configs are settable.

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

Backporting to 11.0.x, which is the oldest version still being supported.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -857,7 +857,7 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   }
 
   private SecurityProtocol securityProtocol() {
-    return SecurityProtocol.valueOf(getString(SECURITY_PROTOCOL_CONFIG));
+    return SecurityProtocol.valueOf(getString(SECURITY_PROTOCOL_CONFIG).toUpperCase());
   }
 
   public Map<String, Object> sslConfigs() {
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -569,7 +569,8 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
             DATA_CONVERSION_GROUP,
             ++order,
             Width.SHORT,
-            BEHAVIOR_ON_NULL_VALUES_DISPLAY
+            BEHAVIOR_ON_NULL_VALUES_DISPLAY,
+            new EnumRecommender<>(BehaviorOnNullValues.class)
         ).define(
             BEHAVIOR_ON_MALFORMED_DOCS_CONFIG,
             Type.STRING,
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/elasticsearch/EnumRecommender.java
```diff
@@ -0,0 +1,67 @@
...
+
+    this.validValues = Collections.unmodifiableSet(validEnums);
+  }
+
+  @Override
+  public void ensureValid(String key, Object value) {
+    // calling toString on itself because IDE complains if the Object is passed.
+    if (value != null && !validValues.contains(value.toString())) {
+      throw new ConfigException(key, value, "Invalid enumerator");
+    }
```
levzem, line 50:
```suggestion
    String enum = value.toString().toUpperCase();
    if (value != null && !validValues.contains(enum)) {
      throw new ConfigException(key, value, "Value must be one of: " + toString());
    }
```

wanna be case insensitive and be more descriptive with the error message

src/main/java/io/confluent/connect/elasticsearch/EnumRecommender.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return validValues.toString();
+  }
+
+  @Override
+  public List<Object> validValues(String name, Map<String, Object> connectorConfigs) {
+    return Collections.unmodifiableList(Arrays.asList(validValues.toArray()));
```
levzem, line 60:
```suggestion
    return Collections.unmodifiableList(new ArrayList<>(validValues)));
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -671,7 +674,8 @@ private static void addSslConfigs(ConfigDef configDef) {
         SSL_GROUP,
         ++order,
         Width.SHORT,
-        SECURITY_PROTOCOL_DISPLAY
+        SECURITY_PROTOCOL_DISPLAY,
+        new EnumRecommender<>(SecurityProtocol.class)
```
levzem, line 678: we can also replace all of the validators with this recommender since it implements `Validator`

src/main/java/io/confluent/connect/elasticsearch/EnumRecommender.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.common.config.ConfigException;
+
+@SuppressWarnings("rawtypes")
```
levzem, line 28: is this necesesary?

levzem (commented):
great change, thanks @amuamushu

PR #533, opened 2021-05-25
CCMSG-1121: Add a boolean configuration for whether to inject the @timestamp field.

amuamushu:
## Problem
The `@timestamp` field is needed in messages to write them to data stream. Currently, the `@timestamp` field is auto-injected with the record timestamp if it is missing. However, users may not want to inject the `@timestamp` field and may instead want those records to fail.

## Solution
Add a configuration letting user decide whether the `@timestamp` field should be auto-injected. The default value is false so that users don't have to explicitly set it to false when writing to regular indices.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

Releasing as part of the data stream feature.

Not backporting or merging to 
 
### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,4 +4,5 @@ common {
   upstreamProjects = 'confluentinc/common'
   pintMerge = true
   downStreamValidate = false
+  nodeLabel = 'docker-oraclejdk8'
 }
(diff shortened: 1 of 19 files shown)


amuamushu:
Is superseded by #540, which now handles injecting all records in cases where the timestamp map is not specified, rather than just missing records.

PR #540, opened 2021-06-01
CCMSG-1135: Create a config that handles mapping to the @timestamp field.

amuamushu:
## Problem
The `@timestamp` field is needed in messages to send them to data stream. Currently, the `@timestamp` field is auto-injected with the record timestamp if it is missing. However, users may not want to inject the `@timestamp` field and may instead want those records to fail. Injecting the `@timestamp` field should only happen if records have no timestamp value at all. Otherwise, having both sets of values (the original `@timestamp` value and the injected value) would be confusing. 

Additionally, field names with special characters are not allowed in [Avro](http://avro.apache.org/docs/current/spec.html#names) and Protobuf, meaning users with Avro or Protobuf schemas can not have an `@timestamp` field. 

## Solution
Since both of these issues relate to the `@timestamp` field, we can add one connector configuration to resolve both of them. 
With this configuration, users set the field names they would like to map to the `@timestamp` field. If multiple are provided, the first match based on the order they were listed is used. If they choose to leave this config unset, all the records would have the `@timestamp` field be the record timestamp.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

Releasing as part of the data streams feature.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -341,8 +341,6 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
           + "already contain this field. This configuration can only be set if ``%s`` and ``%s`` "
           + "are set.",
       DATA_STREAM_TYPE_CONFIG,
-      DATA_STREAM_DATASET_CONFIG,
-      DATA_STREAM_TYPE_CONFIG,
       DATA_STREAM_DATASET_CONFIG
   );
   private static final String DATA_STREAM_TIMESTAMP_DISPLAY = "Data Stream Timestamp Field";
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -108,10 +108,11 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   public static final String DATA_STREAM_TIMESTAMP_CONFIG = "data.stream.timestamp";
   private static final String DATA_STREAM_TIMESTAMP_DOC = String.format(
       "The timestamp to use for the @timestamp field in messages sent to data streams. "
-          + "If multiple fields are listed, the the first field provided that also appears"
-          + " in the message would be used. If this field is not set and %s and %s are set, "
-          + "all of the messages will be auto-injected with the record timestamp. This "
-          + "field can only be set if %s and %s are set.",
+          + "If multiple fields are listed, the first field provided that also appears"
+          + " in the message would be used. If this field is not set but %s and %s are set, "
+          + "all of the messages will have the record timestamp as the @timestamp field. "
+          + "Note that the @timestamp field needs to be explicitly listed if messages "
+          + "already have this field. This field can only be set if %s and %s are set.",
       DATA_STREAM_TYPE_CONFIG,
       DATA_STREAM_DATASET_CONFIG,
       DATA_STREAM_TYPE_CONFIG,
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -78,33 +78,6 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   private static final String BULK_SIZE_BYTES_DISPLAY = "Bulk Size (bytes)";
   private static final int BULK_SIZE_BYTES_DEFAULT = 5 * 1024 * 1024;
 
-  public static final String DATA_STREAM_DATASET_CONFIG = "data.stream.dataset";
-  private static final String DATA_STREAM_DATASET_DOC =
-      "Generic name describing data ingested and its structure to be written to a data stream. "
-      + "Can be any arbitrary string that is no longer than 100 characters, is in all lowercase, "
-      + "and does not contain spaces or any of these special characters ``/\\*\"<>|,#:-``. "
-      + "Otherwise, no value indicates the connector will write to regular indices instead. "
-      + "If set, this configuration will be used alongside ``data.stream.type`` to "
-      + "construct the data stream name in the form of {``data.stream.type``"
-      + "}-{" + DATA_STREAM_DATASET_CONFIG + "}-{topic}.";
-  private static final String DATA_STREAM_DATASET_DISPLAY = "Data Stream Dataset";
-  private static final String DATA_STREAM_DATASET_DEFAULT = "";
-
-  public static final String DATA_STREAM_TYPE_CONFIG = "data.stream.type";
-  private static final String DATA_STREAM_TYPE_DOC = String.format(
-      "Generic type describing the data to be written to data stream. "
-          + "The default is %s which indicates the connector will write "
-          + "to regular indices instead. If set, this configuration will "
-          + "be used alongside %s to construct the data stream name in the form of"
-          + "{%s}-{%s}-{topic}.",
-      DataStreamType.NONE.name(),
-      DATA_STREAM_DATASET_CONFIG,
-      DATA_STREAM_TYPE_CONFIG,
-      DATA_STREAM_DATASET_CONFIG
-  );
-  private static final String DATA_STREAM_TYPE_DISPLAY = "Data Stream Type";
-  private static final DataStreamType DATA_STREAM_TYPE_DEFAULT = DataStreamType.NONE;
-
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -227,6 +227,22 @@
   private static final String COMPACT_MAP_ENTRIES_DISPLAY = "Compact Map Entries";
   private static final boolean COMPACT_MAP_ENTRIES_DEFAULT = true;
 
+  public static final String DATA_STREAM_TIMESTAMP_CONFIG = "data.stream.timestamp";
```
levzem, line 230:
```suggestion
  public static final String DATA_STREAM_TIMESTAMP_CONFIG = "data.stream.timestamp.field";
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -227,6 +227,22 @@
...
+  private static final String DATA_STREAM_TIMESTAMP_DOC = String.format(
+      "The timestamp to use for the @timestamp field in messages sent to data stream. "
```
levzem, line 232: Let's add a brief explanation that ES data streams require documents to contain this field
```suggestion
     <explanation here>  "The Kafka record field to use as the timestamp for the ``@timestamp`` field in documents sent to a data stream. "
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -227,6 +227,22 @@
...
+          + " in the message would be used. If this field is not set but ``%s`` and ``%s`` are set,"
+          + " all of the messages will have the record timestamp as the @timestamp field value. "
+          + "Note that the @timestamp field needs to be explicitly listed if messages "
+          + "already have this field. This field can only be set if ``%s`` and ``%s`` are set.",
+      DATA_STREAM_TYPE_CONFIG,
+      DATA_STREAM_DATASET_CONFIG,
+      DATA_STREAM_TYPE_CONFIG,
+      DATA_STREAM_DATASET_CONFIG
+  );
+  private static final String DATA_STREAM_TIMESTAMP_DISPLAY = "Data Stream Timestamp";
```
levzem, line 243:
```suggestion
  private static final String DATA_STREAM_TIMESTAMP_DISPLAY = "Data Stream Timestamp Field";
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -227,6 +227,22 @@
...
   private static final boolean COMPACT_MAP_ENTRIES_DEFAULT = true;
 
+  public static final String DATA_STREAM_TIMESTAMP_CONFIG = "data.stream.timestamp";
+  private static final String DATA_STREAM_TIMESTAMP_DOC = String.format(
+      "The timestamp to use for the @timestamp field in messages sent to data stream. "
+          + "If multiple fields are provided, the first field listed that also appears"
+          + " in the message would be used. If this field is not set but ``%s`` and ``%s`` are set,"
+          + " all of the messages will have the record timestamp as the @timestamp field value. "
+          + "Note that the @timestamp field needs to be explicitly listed if messages "
+          + "already have this field. This field can only be set if ``%s`` and ``%s`` are set.",
```
levzem, line 237:
```suggestion
          + "If multiple fields are provided, the first field listed that also appears"
          + " in the record will be used. If this configuration is left empty,"
          + " all of the documents will use the record timestamp as the ``@timestamp`` field value. "
          + "Note that the ``@timestamp`` field needs to be explicitly listed if records "
          + "already contain this field. This configuration can only be set if ``%s`` and ``%s`` are set.",
```

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -161,6 +162,17 @@ private void validateDataStreamConfigs() {
       );
       addErrorMessage(BEHAVIOR_ON_NULL_VALUES_CONFIG, errorMessage);
     }
+
+    if (!config.isDataStream() && !config.dataStreamTimestamp().isEmpty()) {
+      String errorMessage = String.format(
+          "Mapping a field to the @timestamp field is only necessary for data streams. "
```
levzem, line 168:
```suggestion
          "Mapping a field to the '@timestamp' field is only necessary for data streams. "
```

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -161,6 +162,17 @@ private void validateDataStreamConfigs() {
...
+              + "%s must not be set if %s and %s are set.",
```
levzem, line 169:
```suggestion
              + "%s must not be set if %s and %s are not set.",
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -848,6 +874,10 @@ public DataStreamType dataStreamType() {
     return DataStreamType.valueOf(getString(DATA_STREAM_TYPE_CONFIG).toUpperCase());
   }
 
+  public List<String> dataStreamTimestamp() {
```
levzem, line 877:
```suggestion
  public List<String> dataStreamTimestampField() {
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
levzem, line 230: at this point we have 3 data stream configs, let's create their own separate config group like we do for kerberos and ssl

src/test/java/io/confluent/connect/elasticsearch/DataConverterTest.java
```diff
@@ -438,4 +491,9 @@ public void testDoNotAddExternalVersioningIfDataStream() {
 
     assertEquals(VersionType.INTERNAL, actualRecord.versionType());
   }
+
+  private void setDataStream() {
```
levzem, line 495:
```suggestion
  private void configureDataStream() {
```

src/test/java/io/confluent/connect/elasticsearch/ValidatorTest.java
```diff
@@ -472,4 +492,9 @@ private static void assertHasErrorMessage(Config config, String property, String
   private static void assertNoErrors(Config config) {
     config.configValues().forEach(c -> assertTrue(c.errorMessages().isEmpty()));
   }
+
+  private void setDataStream() {
```
levzem, line 496:
```suggestion
  private void configureDataStream() {
```

src/main/java/io/confluent/connect/elasticsearch/Validator.java
amuamushu: Oops. Thanks for catching that!

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
```diff
@@ -328,11 +301,57 @@
...
+  private static final String DATA_STREAM_TIMESTAMP_DOC = String.format(
+      "The Kafka record field to use as the timestamp for the ``@timestamp`` field in documents "
+          + "sent to a data stream.\n All documents sent to a data stream needs an``@timestamp`` "
+          + " field with values of type ``date`` or ``data_nanos``. Otherwise, the document "
+          + " would not be sent. If multiple fields are provided, the first field listed that "
+          + "also appears in the record will be used. If this configuration is left empty,"
+          + " all of the documents will use the record timestamp as the ``@timestamp`` field "
+          + "value. Note that the ``@timestamp`` field needs to be explicitly listed if records "
+          + "already contain this field. This configuration can only be set if ``%s`` and ``%s`` "
+          + "are set.",
```
levzem, line 342:
```suggestion
  private static final String DATA_STREAM_TIMESTAMP_DOC = String.format(
      "The Kafka record field to use as the timestamp for the ``@timestamp`` field in documents "
          + "sent to a data stream.\n All documents sent to a data stream needs an``@timestamp`` "
          + " field with values of type ``date`` or ``data_nanos``. Otherwise, the document "
          + " will not be sent. If multiple fields are provided, the first field listed that "
          + "also appears in the record will be used. If this configuration is left empty,"
          + " all of the documents will use the Kafka record timestamp as the ``@timestamp`` field "
          + "value. Note that ``@timestamp``still  needs to be explicitly listed if records "
          + "already contain this field. This configuration can only be set if ``%s`` and ``%s`` "
          + "are set.",
```

levzem (commented):
LGTM for what we have right now, let's wait on the performance tests to finalize this

PR #546, opened 2021-06-04
Add logging for retries

ilanjiR:
## Problem
When we retry failed requests, we don't log the cause of the failure until we exhaust the retries. If the consumer leaves the group before we exhaust the retries, we will never see the cause of the error.

## Solution
Log every retry attempt with the exception

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
@@ -160,6 +160,7 @@ public class RetryUtil {
         );
         return function.call();
       } catch (IOException e) {
+        log.debug("Failed to {} due to {}. Retrying attempt ({}/{})", description, e, attempt, maxAttempts);
         if (attempt >= maxAttempts) {
           throw new ConnectException("Failed to " + description, e);
         }
--- a/src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
@@ -161,7 +161,8 @@ public class RetryUtil {
         return function.call();
       } catch (IOException e) {
         if (attempt >= maxAttempts) {
-          log.error("Failed to {} due to {} after total of {} attempt(s)", description, e.getCause(), maxAttempts, e);
+          log.error("Failed to {} due to {} after total of {} attempt(s)",
+              description, e.getCause(), maxAttempts, e.getMessage());
           throw new ConnectException("Failed to " + description, e);
         }
         // Otherwise it is retriable and we should retry
--- a/src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
@@ -160,7 +160,7 @@ public class RetryUtil {
         );
         return function.call();
       } catch (IOException e) {
-        log.debug("Failed to {} due to {}. Retrying attempt ({}/{})", description, e, attempt, maxAttempts);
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
```diff
@@ -160,6 +160,7 @@ public static long computeRetryWaitTimeInMillis(int retryAttempts,
         );
         return function.call();
       } catch (IOException e) {
+        log.debug("Failed to {} due to {}. Retrying attempt ({}/{})", description, e.getCause(), attempt, maxAttempts);
```
kpatelatwork, line 163: My 2 cents we should change this to warn.  The reason is that customer already told us that if he changes log to TRACE then it generates too much log, as this situation is now occurring once every few days we need to catch it in the act and that's why we should do it as warning.  The old PR also had this as warn https://github.com/confluentinc/kafka-connect-elasticsearch/pull/116/files#diff-b929a6768a74899db0afa5ad8eba4f2a0fafb65c9df9372085dfb01e840c02eeR361

kpatelatwork:
LGTM from my side but please check with Konstantine also if we need more logs(I vaguely remember there were more than 1 places but I may be wrong)

ilanjiR:
@kkonstantine Are there any other logs we should add before shipping this out to the customer?

src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
```diff
@@ -160,6 +160,7 @@ public static long computeRetryWaitTimeInMillis(int retryAttempts,
         );
         return function.call();
       } catch (IOException e) {
+        log.warn("Failed to {} due to {}. Retrying attempt ({}/{})", description, e.getCause(), attempt, maxAttempts);
```
kkonstantine, line 163: First the error log here won't be accurate if the attempt is equal to max attempts. 

I think we can distinguish between a warn message while we keep retrying and a similar error message when we exhaust retries and before we throw (I think exception will be printed but should be fine to have the symmetric error message like [before](https://github.com/confluentinc/kafka-connect-elasticsearch/pull/116/files#diff-b929a6768a74899db0afa5ad8eba4f2a0fafb65c9df9372085dfb01e840c02eeR367))

kkonstantine: Then I think it might be useful to print the backoff along with the warn message when we retry to know how long (in ms) the task will wait. So, that means that the log should be added in two cases (error and warn) and should be moved further down. Wdyt?

src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
```diff
@@ -160,13 +160,14 @@ public static long computeRetryWaitTimeInMillis(int retryAttempts,
         );
         return function.call();
       } catch (IOException e) {
-        log.warn("Failed to {} due to {}. Retrying attempt ({}/{})", description, e.getCause(), attempt, maxAttempts);
         if (attempt >= maxAttempts) {
+          log.error("Failed to {} due to {} after total of {} attempt(s)", description, e.getCause(), maxAttempts, e);
```
kkonstantine, line 164: let's do use `e.getMessage()` instead of `e` if you agree that `e` would results in printing the stack trace twice. To me it seems that the stack trace will be printed as part of `ConnectException`.

kkonstantine (approved):
Thanks @ilanjiR 
One last minor comment which I'd appreciate if you can confirm, but otherwise LGTM

kkonstantine:
Are we ready to merge here?

PR #550, opened 2021-06-08
CCMSG-1137: Handle ElasticsearchStatusException in validateConnection.

avocader:
## Problem
We call `org.elasticsearch.client.RestHighLevelClient#ping` in `io.confluent.connect.elasticsearch.Validator#validateConnection` and we are catching `IOException` only.
But `RestHighLevelClient` can throw `ElasticsearchStatusException` as well, which results in the validate endpoint returning HTTP 500.

## Solution
Catch `ElasticsearchStatusException` to avoid throwing `500`.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/Validator.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/Validator.java
@@ -269,7 +269,7 @@ public class Validator {
     String exceptionMessage = "";
     try {
       successful = client.ping(RequestOptions.DEFAULT);
-    } catch (IOException | RuntimeException e) {
+    } catch (Exception e) {
       successful = false;
       exceptionMessage = String.format("Error message: %s", e.getMessage());
     }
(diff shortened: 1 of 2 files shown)


dosvath (approved):
LGTM

yanglei99 (approved):
Look good for me to address current issue. However should we also try to catch all Runtime exception to ensure we never throw any exception out of validation? The caller should never get internal error of 500 on config/validate…

avocader:
> Look good for me to address current issue. However should we also try to catch all Runtime exception to ensure we never throw any exception out of validation? The caller should never get internal error of 500 on config/validate…

Makes sense to me, let me add a broader catch to make sure we never throw a `500`.

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -269,11 +269,10 @@ private void validateConnection(RestHighLevelClient client) {
     String exceptionMessage = "";
     try {
       successful = client.ping(RequestOptions.DEFAULT);
-    } catch (IOException e) {
+    } catch (IOException | RuntimeException e) {
```
levzem, line 272:
```suggestion
    } catch (Exception e) {
```

we might as well make it the most generic

levzem (commented):
LGTM

PR #552, opened 2021-06-11
CCMSG-1181: Updated validateConnection to not fail if user doesn't have monitor permission.

avocader:
## Problem
After we introduced `io.confluent.connect.elasticsearch.Validator#validateConnection` method, the validation started to fail with the following exception:
```
Caused by: ElasticsearchStatusException[method [HEAD], host [https://XXX:1234], URI [/], status line [HTTP/1.1 403 Forbidden]]; nested: ResponseException[method [HEAD], host [https://XXX:1234], URI [/], status line [HTTP/1.1 403 Forbidden]];
	at org.elasticsearch.client.RestHighLevelClient.parseResponseException(RestHighLevelClient.java:1680)
	at org.elasticsearch.client.RestHighLevelClient.internalPerformRequest(RestHighLevelClient.java:1446)
	at org.elasticsearch.client.RestHighLevelClient.performRequest(RestHighLevelClient.java:1403)
	at org.elasticsearch.client.RestHighLevelClient.ping(RestHighLevelClient.java:677)
	at io.confluent.connect.elasticsearch.Validator.validateConnection(Validator.java:271)
	at io.confluent.connect.elasticsearch.Validator.validate(Validator.java:100)
	at io.confluent.connect.elasticsearch.ElasticsearchSinkConnector.validate(ElasticsearchSinkConnector.java:80)
	at org.apache.kafka.connect.runtime.AbstractHerder.validateConnectorConfig(AbstractHerder.java:406)
	at org.apache.kafka.connect.runtime.AbstractHerder.lambda$validateConnectorConfig$2(AbstractHerder.java:354)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	... 1 more
Caused by: org.elasticsearch.client.ResponseException: method [HEAD], host [XXX:1234], URI [/], status line [HTTP/1.1 403 Forbidden]
	at org.elasticsearch.client.RestClient.convertResponse(RestClient.java:260)
	at org.elasticsearch.client.RestClient.performRequest(RestClient.java:238)
	at org.elasticsearch.client.RestClient.performRequest(RestClient.java:212)
	at org.elasticsearch.client.RestHighLevelClient.internalPerformRequest(RestHighLevelClient.java:1433)
	... 12 more
```  
which can be a false-negative, since a cluster is actually running, but a user is not authorized to call this endpoint.

## Solution

- Treat `403` status as a cluster, that we can establish a connection to.
- Do not use an Elastic superuser in integration tests(ES super user is still used in `ElasticsearchClientTest`, which not an integration test).
- Update docs(https://github.com/confluentinc/docs-kafka-connect-elasticsearch/pull/45) and README to have an accurate list of required Elasticsearch privileges.

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/Validator.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/Validator.java
@@ -16,22 +16,25 @@
 
 package io.confluent.connect.elasticsearch;
 
-import io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.SecurityProtocol;
-import java.io.IOException;
-import java.util.List;
-import java.util.Map;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 import org.apache.http.HttpHost;
 import org.apache.kafka.common.config.Config;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.config.ConfigValue;
 import org.apache.kafka.common.config.SslConfigs;
+import org.elasticsearch.ElasticsearchStatusException;
 import org.elasticsearch.client.RequestOptions;
 import org.elasticsearch.client.RestClient;
 import org.elasticsearch.client.RestHighLevelClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.SecurityProtocol;
 
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.BATCH_SIZE_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.CONNECTION_PASSWORD_CONFIG;
@@ -269,6 +272,16 @@ public class Validator {
     String exceptionMessage = "";
     try {
       successful = client.ping(RequestOptions.DEFAULT);
+    } catch (ElasticsearchStatusException e) {
+      switch (e.status()) {
+        case FORBIDDEN:
+          // ES is up, but user is not authorized to ping server
+          successful = true;
+          break;
+        default:
+          successful = false;
+          exceptionMessage = String.format("Error message: %s", e.getMessage());
+      }
     } catch (Exception e) {
       successful = false;
       exceptionMessage = String.format("Error message: %s", e.getMessage());
--- a/src/test/java/io/confluent/connect/elasticsearch/helper/ElasticsearchContainer.java
+++ b/src/test/java/io/confluent/connect/elasticsearch/helper/ElasticsearchContainer.java
@@ -15,6 +15,19 @@
 
 package io.confluent.connect.elasticsearch.helper;
 
+import org.elasticsearch.client.security.user.User;
+import org.elasticsearch.client.security.user.privileges.Role;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.ContainerLaunchException;
+import org.testcontainers.containers.output.OutputFrame;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.images.RemoteDockerImage;
+import org.testcontainers.images.builder.ImageFromDockerfile;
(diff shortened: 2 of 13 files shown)


src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -269,6 +270,16 @@ private void validateConnection(RestHighLevelClient client) {
...
       successful = client.ping(RequestOptions.DEFAULT);
+    } catch (ElasticsearchStatusException e) {
+      switch(e.status()) {
+        case FORBIDDEN:
+          // ES is up, but user is not authorized to ping server
+          successful = true;
+          break;
+        default:
+          successful = false;
+          exceptionMessage = String.format("Error message: %s", e.getMessage());
```
levzem, line 283:
```suggestion
          throw e;
```

this will be recaught below and we can reuse the same logic

src/test/java/io/confluent/connect/elasticsearch/helper/ElasticsearchContainer.java
```diff
@@ -218,6 +291,12 @@ protected void configure() {
         .withFileFromClasspath("instances.yml", getFullResourcePath("instances.yml"))
         .withDockerfileFromBuilder(this::buildImage);
 
+    if (isBasicAuthEnabled()) {
+      log.info("Setting up basic authentication in a Docker image");
+      withEnv("ELASTICSEARCH_USERNAME", ELASTIC_SUPERUSER_NAME);
+      withEnv("ELASTIC_PASSWORD", ELASTIC_PASSWORD);
```
levzem, line 297: basicAuth is mutually exclusive
```suggestion
      withEnv("ELASTIC_PASSWORD", ELASTIC_PASSWORD);
      return;
```

src/test/java/io/confluent/connect/elasticsearch/helper/ElasticsearchContainer.java
```diff
@@ -185,13 +243,28 @@ public boolean isKerberosEnabled() {
...
+    if (isCreated()) {
+      throw new IllegalStateException(
+          "enableBasicAuth can only be used before the container is created."
+      );
+    }
+    this.users = users;
+  }
+
+  public boolean isBasicAuthEnabled() {
+    return this.users != null && !this.users.isEmpty();
```
levzem, line 256:
```suggestion
    return users != null && !users.isEmpty();
```

levzem (commented):
nice tests @avocader

it would make more sense to run all of our tests against the minimal permissions to catch regressions like these, so maybe the default container is brought up is the `basicAuth` container for all tests instead of explicitly enabling it. Then in the `setup()` we create a minimal user using the helper client as mentioned below and other tests can create and use other users as necessary, that should reduce some of the code following that approach and make it more readable

yanglei99 (approved):
Lgtm. Thanks!

src/main/java/io/confluent/connect/elasticsearch/Validator.java
kpatelatwork: IMHO it won't be recaught below because its a cascading exception block and not a nested try/catch

src/test/java/io/confluent/connect/elasticsearch/helper/ElasticsearchContainer.java
avocader: We can't return here, we have to do `setImage(image);`. But let me restructure the code to make these options mutually exclusive indeed in case some test sets multiple of them.

src/main/java/io/confluent/connect/elasticsearch/Validator.java
avocader: It won't be recaught, +1 @kpatelatwork .

kpatelatwork:
I am new to the code but the test looks ok 😁 to me logically.

levzem (commented):
this needs to target `11.0.x`

levzem (approved):
LGTM

PR #556, opened 2021-06-17
CCMSG-1203: Update Data Stream Happy Path IT to include monitor privilege.

amuamushu:
## Problem
After the latest merge to 11.1.x that specified role privileges, the data stream happy path IT failed with the error message 

`org.apache.kafka.connect.runtime.rest.errors.ConnectRestException: Could not execute PUT request. Error response: {"error_code":500,"message":"Elasticsearch exception [type=security_exception, reason=action [cluster:monitor/main] is unauthorized for user [frank]]"}`

## Solution
In the IT, a user with the `monitor` privilege needs to be added and used when data streams is enabled.

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


levzem (approved):
LGTM

PR #557, opened 2021-06-23
Upgrade ES client to latest version

ilanjiR:
## Problem
The ES client we are on is from 2019

## Solution
Upgrade the client to the latest version

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -292,7 +292,7 @@
                 <configuration>
                     <compilerArgs>
                         <arg>-Xlint:all,-try</arg>
-<!--                        <arg>-Werror</arg>-->
+                        <arg>-Werror</arg>
                     </compilerArgs>
                 </configuration>
             </plugin>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -292,7 +292,6 @@
                 <configuration>
                     <compilerArgs>
                         <arg>-Xlint:all,-try</arg>
-                        <arg>-Werror</arg>
```
ilanjiR, line 295: This flag is causing the build to fail due to a deprecation warning. Not sure if there's a more elegant way to avoid the warning other than switching to the new builder (which looks complicated)

dosvath (approved):
LGTM, @ilanjiR do go through the breaking changes between the last and new version here https://www.elastic.co/guide/en/elasticsearch/reference/7.13/es-release-notes.html, to make sure we're safe to upgrade.

Issue #223, opened 2018-07-13
Support elastic-search routing for kafka-connect-elasticsearch sink.


rkalluri-clgx:
I would like to know if we support routing of records to certain shards of elastic search as explained below.

https://www.elastic.co/guide/en/elasticsearch/reference/6.2/mapping-routing-field.html

Looks like we need to be able to specify routing param in the elastic url, and can vary from message to message. Looks like this needs to be dynamic with a SMT of some sort to pick routing from the message just like we pick the index. Either that or have a per partition route from Kafka for performance reasons. Just wanted to kickstart the discussions.

DevonPeroutky:
+1 Would love to see this.

robgryn:
+1 Would also benefit from this

maxsel:
+1
This feature would enable parent-child relationship in target Elasticsearch cluster, otherwise I have no idea how to achieve this.

matpersonne:
+1
I realy need this feature, I changed all my ES mapping to join / routing. I'm now stuck with kafka connect to add the routing={id}

mikelsanvi:
+1 This would be really useful

fubhy:
I also need this feature for defining parent/child join relations: https://www.elastic.co/guide/en/elasticsearch/reference/6.3/parent-join.html

MottiniMauro:
+1
Having issues finding a way to implement parent-child relationship without this

frankkoornstra:
+1
Would be great to make our cluster more efficient

dgthomugo:
@levzem Is there a plan to support this feature? we really need a way to flush parent-child relationship records in kafka-connect-elasticsearch.

mIkhail-zaretsky:
Is there any workaround??

arungitan:
same here.  Would like to be able to insert parent-child records via elasticsearch connector.   Any updates on this?

frankkoornstra:
That leaks a details to Kafka Connect, like how the shard number is calculated and how many shards the target index has. Besides, it would involve storing state about which ids are still available, creating an algo to come up with the next id for a targeted shard...  all things that shouldn't be necessary if routing would be supported.

arungitan:
for those still trying to solve this, I found this cool workaround (if you have some control on the elasticsearch side):  Use an ingest processor! 
https://www.elastic.co/guide/en/elasticsearch/reference/current/accessing-data-in-pipelines.html
The line of interest is this: 
[...]

PR #558, opened 2021-06-24
Adds routing support, configurable via 'routing.field.name'

hartmut-co-uk:
Refs:
- https://github.com/confluentinc/kafka-connect-elasticsearch/issues/223
- https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-routing-field.html
- https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-routing
- https://www.elastic.co/guide/en/elasticsearch/reference/current/parent-join.html

## Problem
ElasticSearch `_routing` field currently is not supported.

- [x] yes
- [ ] no

##### If yes, where?
Parent/Child relationships -> join field type

## Test Strategy
`DataConverterTest` has been added.   
Integration test might require refactoring of existing tests.
`ElasticsearchConnectorIT` / `ElasticsearchConnectorBaseIT` seems to use a very basic schema / payload.
Also `ElasticsearchHelperClient` would need to be improved to also support __routing_.

**Please advise** how to approach.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan
* New feature
* Non-breaking change
* New config fields are optional
* Does not require any change or migration upon upgrade

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.BehaviorOnNullValues;
+import org.apache.http.util.TextUtils;
 import org.apache.kafka.connect.data.ConnectSchema;
 import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
@@ -29,6 +30,7 @@ import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.data.Time;
 import org.apache.kafka.connect.data.Timestamp;
+import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.errors.DataException;
 import org.apache.kafka.connect.json.JsonConverter;
 import org.apache.kafka.connect.sink.SinkRecord;
@@ -160,6 +162,9 @@ public class DataConverter {
         ? String.format("%s+%d+%d", record.topic(), record.kafkaPartition(), record.kafkaOffset())
         : convertKey(record.keySchema(), record.key());
 
+    // routing
+    String routing = getRouting(record);
+
     // delete
     if (record.value() == null) {
       return maybeAddExternalVersioning(new DeleteRequest(index).id(id), record);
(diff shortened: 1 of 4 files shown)


hartmut-co-uk:
I noticed for parent-join use case - in addition to adding the routing, the payload also potentially needs to be enriched with the `my_join_field.name` and `my_join_field.parent` (for children).
https://www.elastic.co/guide/en/elasticsearch/reference/current/parent-join.html
```
PUT my-index-000001/_doc/3?routing=1&refresh 
{
  "my_id": "3",
  "parent_id": "1",
  "text": "This is an answer",
  "my_join_field": {
    "name": "answer", 
    "parent": "1" 
  }
}
```

I wonder if it would be worth to also natively build into this connector, instead of forcing the user to enrich data upfront, or build+require a custom SMT.

Note: I tried with _InsertField_ SMT - but since it only supports flat fields it's impossible to enrich the struct for children.
https://docs.confluent.io/platform/current/connect/transforms/insertfield.html

So one option I could see is to add following config options:
- `join.field.name` in above example would be 'my_join_field'
- `join.relation.name` in above example would be 'answer'
- `join.parent.id.field.name` would essentially be the value field (_path_), -> e.g. same as would be configured for `routing.field.name`

hartmut-co-uk:
A custom SMT might be the best fit to create the ‘join-field struct’ for ES parent/child relationship on the fly.
I wonder if it's possible and would be an acceptable pattern to just add + ship a custom SMT for this along with the connector jar?

hartmut-co-uk:
Further room for improvement: allow to use data from `$key` (not only from a `$value` field)

hartmut-co-uk:
I've implemented a simple SMT (also works just fine having it embedded in the connector jar).

sticking with the ES [example](https://www.elastic.co/guide/en/elasticsearch/reference/current/parent-join.html) it's configured like:
```
  "transforms.insertParentJoin.join.field.name": "my_join_field",
  "transforms.insertParentJoin.relation.type.name": "answer",
  "transforms.insertParentJoin.parent.id.field.path": "questionId"
```
(with the actual msg value having a field `questionId` as the 'question' parent ID of the answer doc - this would also be used for routing...)

Please provide some feedback on how to proceed further.

frankkoornstra:
Amazing work @hartmut-co-uk :heart_eyes: This is going to help us a lot!

hartmut-co-uk:
@levzem could you please provide initial feedback?

Also
* advise how to proceed with integration tests?
  > Integration test might require refactoring of existing tests.
ElasticsearchConnectorIT / ElasticsearchConnectorBaseIT seems to use a very basic schema / payload.
Also ElasticsearchHelperClient would need to be improved to also support _routing.

* Do you think it would make sense to add an SMT to the connector? I have an implementation of a 
  `public class InsertParentJoin<R extends ConnectRecord<R>> implements Transformation<R> {`
  (I've had to add dependency `org.apache.kafka:connect-transforms:${kafka.version}`)
  * Or would/should it make more sense to embed the functionality directly into the connector?
  * Or do not add but maybe create + provide as a separate maven artifact?

hartmut-co-uk:
I'm still keen to help wrap this up, write more tests, .. if we can agree how to proceed.

frankkoornstra:
@connect-team1 anyone? This has been dead in the water without comments from the team for ~2 months now...

hartmut-co-uk:
ping

frankkoornstra:
Pong @yanglei99 @kkonstantine @levzem @dosvath

frankkoornstra:
Went on a sabbatical for 3 months and still :joy: Is this project abandoned all together?

hartmut-co-uk:
I don't think it's been abandoned in general. I'm currently busy with other things. 
(I've not yet come to use the proposed changes beyond a POC)

Still happy to contribute if someone confirms that the feature/changes are welcome and will be merged + how to proceed with adding tests / SMT.

dania-ru:
Any plans to merge this PR?

hartmut-co-uk:
Dear maintainers, is this feature/PR welcome, and are further changes required?
Is the config option fine as is, what about the new private methods added (are there better alternatives/libraries to use)?

I'm still keen to help wrap this up and write tests, ... if we can agree on how to proceed.

frankkoornstra:
If you're still searching for a platform-type of solution, take a look at Apache Flink. You get pretty much access to the bare-bones Elasticsearch API so you can do whatever you want, including routing, timeouts per bulk action, etc. Plus an actual Web UI :star_struck:

hartmut-co-uk:
> If you're still searching for a platform-type of solution, take a look at Apache Flink. You get pretty much access to the bare-bones Elasticsearch API so you can do whatever you want, including routing, timeouts per bulk action, etc. Plus an actual Web UI 🤩

Use [flink kafka source](https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/connectors/datastream/kafka/#kafka-source) -> [flink elasticsearch sink](https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/connectors/datastream/elasticsearch/#elasticsearch-sink) -- instead of kafka connect? 🤔
This doesn't unblock getting this PR over the line ^^ but it's a great suggestion of an alternative solution way easier to customise.  👍 👍 

Many thanks for bringing this up, given I've been considering Apache Flink for other use cases in my project, too...

sandyplace:
I am no expert but I think this approach can be difficult, routing based upon a value in a record is tricky when it comes to deleting. Typically deleting is triggered by a null record. We noticed when that is the case the delete request does not route the delete to the correct shard because the routing key is missing from the record. We had to take a different approach to get this to work.

PR #559, opened 2021-06-24
MINOR: Re-add Guava as compile time dependency

kkonstantine:
## Problem
Guava is incorrectly marked as scope dependency. The issue won't surface as long as guava is included with CP

## Solution
Add the compile time dep

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -134,12 +134,10 @@
             <version>${lucene.version}</version>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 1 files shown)


dosvath (approved):
LGTM, saw this with team yesterday

PR #560, opened 2021-06-25
MINOR: Remove guava as a dependency

kkonstantine:
## Problem
Guava was a dependency of the jest client.

## Solution
We can remove in 11.0.x and newer

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -149,11 +149,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


dosvath (approved):
LGTM

PR #561, opened 2021-06-28
Log server version on sink task startup

ilanjiR:
## Problem
We need to know what version of ES server our customers are running in cloud so we can determine if it's safe to upgrade
## Solution
Log ES server version on sink task startup

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -160,7 +160,7 @@ public class ElasticsearchSinkTask extends SinkTask {
       try {
         highLevelClient.close();
       } catch (Exception e) {
-        log.error("Failed to close high level client");
+        log.warn("Failed to close high level client");
       }
     }
     return esVersionNumber;
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -117,6 +117,11 @@ public class ElasticsearchSinkTask extends SinkTask {
   public void stop() {
     log.debug("Stopping Elasticsearch client.");
     client.close();
+    try {
+      highLevelClient.close();
+    } catch (Exception e) {
+      log.error("Failed to close high level rest client", e);
+    }
   }
 
   @Override
@@ -154,7 +159,8 @@ public class ElasticsearchSinkTask extends SinkTask {
     try {
       response = highLevelClient.info(RequestOptions.DEFAULT);
       esVersionNumber = response.getVersion().toString();
-    } catch (IOException | ElasticsearchStatusException e) {
+    } catch (Exception e) {
+      log.info("Failed to get ES server version", e);
       // Same error messages as from validating the connection for IOException.
       // Insufficient privileges to validate the version number if caught
       // ElasticsearchStatusException.
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -45,7 +45,6 @@ public class ElasticsearchSinkTask extends SinkTask {
 
   private DataConverter converter;
   private ElasticsearchClient client;
-  private RestHighLevelClient highLevelClient;
   private ElasticsearchSinkConnectorConfig config;
   private ErrantRecordReporter reporter;
   private Set<String> existingMappings;
@@ -115,12 +114,6 @@ public class ElasticsearchSinkTask extends SinkTask {
   public void stop() {
     log.debug("Stopping Elasticsearch client.");
     client.close();
-    try {
-      highLevelClient.close();
-      highLevelClient = null;
-    } catch (Exception e) {
-      log.error("Failed to close high level rest client", e);
-    }
   }
 
   @Override
@@ -141,7 +134,7 @@ public class ElasticsearchSinkTask extends SinkTask {
 
   private String getServerVersion() {
     ConfigCallbackHandler configCallbackHandler = new ConfigCallbackHandler(config);
-    this.highLevelClient = highLevelClient != null ? highLevelClient : new RestHighLevelClient(
+    RestHighLevelClient highLevelClient = new RestHighLevelClient(
         RestClient
             .builder(
                 config.connectionUrls()
@@ -163,6 +156,11 @@ public class ElasticsearchSinkTask extends SinkTask {
       // Insufficient privileges to validate the version number if caught
       // ElasticsearchStatusException.
       log.warn("Failed to get ES server version", e);
+    }
+    try {
+      highLevelClient.close();
+    } catch (Exception e) {
+      log.error("Failed to close high level client");
     }
     return esVersionNumber;
   }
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -38,7 +38,7 @@
     </scm>
 
     <properties>
-        <es.version>7.9.3</es.version>
+        <es.version>7.0.1</es.version>
```
ilanjiR, line 41: Reverted the upgrade here: https://github.com/confluentinc/kafka-connect-elasticsearch/pull/557/files

It looks like the MappingMetaData change does not need to be reverted

ilanjiR, line 41: We reverted the version upgrade from here: https://github.com/confluentinc/kafka-connect-elasticsearch/pull/557/files until we can confirm that it will not break compatibility

kkonstantine (commented):
@ilanjiR I'd recommend reverting on a separate PR to make changes more clear. Also, let's mention on the title that the logging change is `MINOR:`

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -125,6 +135,33 @@ private void checkMapping(SinkRecord record) {
...
+            )
+            .setHttpClientConfigCallback(configCallbackHandler)
+            .setRequestConfigCallback(configCallbackHandler)
+    );
+    MainResponse response;
+    String esVersionNumber = "Unknown";
+    try {
+      response = highLevelClient.info(RequestOptions.DEFAULT);
+      esVersionNumber = response.getVersion().toString();
+    } catch (IOException | ElasticsearchStatusException e) {
```
kpatelatwork, line 157: we should catch Exception e and also log the exception.  This way getting version is an optional thing and we should start the connector regardless

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -15,29 +15,39 @@
...
+
+import io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.BehaviorOnNullValues;
 
 public class ElasticsearchSinkTask extends SinkTask {
 
   private static final Logger log = LoggerFactory.getLogger(ElasticsearchSinkTask.class);
 
   private DataConverter converter;
   private ElasticsearchClient client;
+  private RestHighLevelClient highLevelClient;
```
kpatelatwork, line 50: we should close the highLevelClient and set it to null in stop.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -72,7 +81,8 @@ protected void start(Map<String, String> props, ElasticsearchClient client) {
 
     this.client = client != null ? client : new ElasticsearchClient(config, reporter);
 
-    log.info("Started ElasticsearchSinkTask.");
+    log.info("Started ElasticsearchSinkTask. Connecting to ES server version: {}",
```
kpatelatwork, line 82: in the future scapping logs wont work. We should file another enhancement ticket to log this version as a metric this way we can plot it in datadog

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -125,6 +140,34 @@ private void checkMapping(SinkRecord record) {
...
+            .setHttpClientConfigCallback(configCallbackHandler)
+            .setRequestConfigCallback(configCallbackHandler)
+    );
+    MainResponse response;
+    String esVersionNumber = "Unknown";
+    try {
+      response = highLevelClient.info(RequestOptions.DEFAULT);
+      esVersionNumber = response.getVersion().toString();
+    } catch (Exception e) {
+      log.info("Failed to get ES server version", e);
```
kpatelatwork, line 162: log.error

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -125,6 +140,34 @@ private void checkMapping(SinkRecord record) {
...
+      // Same error messages as from validating the connection for IOException.
```
kpatelatwork, line 163: please move the comments above the log line :)

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -107,6 +117,11 @@ public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
   public void stop() {
     log.debug("Stopping Elasticsearch client.");
     client.close();
+    try {
+      highLevelClient.close();
```
kpatelatwork, line 119: please set it to null also because in start we are doing `this.highLevelClient = highLevelClient != null ? highLevelClient : new RestHighLevelClient(`

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
kkonstantine: error is quite strict for this soft information. 
`log.warn` seems like a good compromise

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
ilanjiR: https://confluentinc.atlassian.net/browse/CCMSG-1218

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
kkonstantine: Also, if we use the high level client only to get the version, should we close it (and discard it) here?
What's the life cycle of this client? It might utilized threads, etc.

kpatelatwork: This is even better idea.  Agree on removing the instance level field and closing it here.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -125,6 +132,39 @@ private void checkMapping(SinkRecord record) {
...
+    } catch (Exception e) {
+      // Same error messages as from validating the connection for IOException.
+      // Insufficient privileges to validate the version number if caught
+      // ElasticsearchStatusException.
+      log.warn("Failed to get ES server version", e);
+    }
+    try {
+      highLevelClient.close();
+    } catch (Exception e) {
+      log.error("Failed to close high level client");
```
kpatelatwork, line 163: warn

kpatelatwork:
@ilanjiR  LGTM. just resolve the last minor comment before merge

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
kpatelatwork: also put this in finally of the above try?

ilanjiR: Shouldn't it be an error? Since leaving a client open could cause issues if it happens many times over the life of a cluster.

ilanjiR:
Will merge on green build

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
ilanjiR: I was thinking that leaving the client open could be potentially problematic over the long life span of a cluster

PR #564, opened 2021-06-28
Upgrade ES client to 7.9.3

ilanjiR:
## Problem
The existing ES client has a known deadlock issue: https://github.com/elastic/elasticsearch/issues/47599
## Solution
Upgrade to a newer version of the client with the fix

- [ ] yes
- [ ] no

## Test Strategy
The client upgrade doesn't seem to bring any incompatible changes based on the client's changelogs.  We will do further testing to validate this before releasing the connector.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,7 @@
     </scm>
 
     <properties>
-        <es.version>7.0.1</es.version>
+        <es.version>7.9.3</es.version>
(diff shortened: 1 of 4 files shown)


kkonstantine (approved):
LGTM. But I agree. Let's do some more testing before we release

PR #565, opened 2021-06-30
Change version logging to work with upgraded client

ilanjiR:
## Problem
Once we upgraded ES client to `7.9.3`, the method we were using for logging ES server versions was changed

## Solution
Update to the newer way to getting ES server version

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -150,7 +150,7 @@ public class ElasticsearchSinkTask extends SinkTask {
     String esVersionNumber = "Unknown";
     try {
       response = highLevelClient.info(RequestOptions.DEFAULT);
-      esVersionNumber = response.getVersion().toString();
+      esVersionNumber = response.getVersion().getNumber();
     } catch (Exception e) {
       // Same error messages as from validating the connection for IOException.
       // Insufficient privileges to validate the version number if caught


kkonstantine (commented):
Do we have a unit test for this method? I'd expect it to fail if the underlying implementation changes

kpatelatwork:
> Do we have a unit test for this method? I'd expect it to fail if the underlying implementation changes

writing a unit test for this will be tricky unless we test it against a specific docker container version.

kkonstantine:
> > Do we have a unit test for this method? I'd expect it to fail if the underlying implementation changes
> 
> writing a unit test for this will be tricky unless we test it against a specific docker container version.

I think we need at least to assert what this method returns with whatever mocks we use in this repo. 
I feel this might have caught the fact that we got a class object instead of a version number. But I might be missing something

kpatelatwork:
> > > Do we have a unit test for this method? I'd expect it to fail if the underlying implementation changes
> > 
> > 
> > writing a unit test for this will be tricky unless we test it against a specific docker container version.
> 
> I think we need at least to assert what this method returns with whatever mocks we use in this repo.
> I feel this might have caught the fact that we got a class object instead of a version number. But I might be missing something

Got it so what you are saying is extract the line response.getVersion().getNumber() to a method getVersion(MainResonse response). and make it package protected and then mock the call response.getVersion().getNumber() returns a string, That should be easy and much faster then trying to run a container and test.   @ilanjiR  let me know if you need help.

kkonstantine:
> > > > Do we have a unit test for this method? I'd expect it to fail if the underlying implementation changes
> > > 
> > > 
> > > writing a unit test for this will be tricky unless we test it against a specific docker container version.
> > 
> > 
> > I think we need at least to assert what this method returns with whatever mocks we use in this repo.
> > I feel this might have caught the fact that we got a class object instead of a version number. But I might be missing something
> 
> Got it so what you are saying is extract the line response.getVersion().getNumber() to a method getVersion(MainResonse response). and make it package protected and then mock the call response.getVersion().getNumber() returns a string, That should be easy and much faster/simple than trying to run a container and test. @ilanjiR let me know if you need help.

Probably test `getServerVersion` as a whole (which is now `private`). Which should have the same effect with what you mention.

Issue #193, opened 2018-04-07
[KafkaConnect to ES ]org.apache.kafka.connect.errors.DataException: STRUCT is not supported as the document id.


JanardhanVarikuti:
Hi,

Getting the below error while trying to connect to Elastic Search (dokcer image - elasticsearch:2.4.3) using Confluent Kafka connect (v4.0)

===================================================================
[2018-04-07 17:12:51,083] ERROR WorkerSinkTask{id=elasticsearch-sink-0} Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restarted. (org.apache.kafka.connect.runtime.WorkerSinkTask:515)
org.apache.kafka.connect.errors.DataException: STRUCT is not supported as the document id.
	at io.confluent.connect.elasticsearch.DataConverter.convertKey(DataConverter.java:88)
	at io.confluent.connect.elasticsearch.DataConverter.convertRecord(DataConverter.java:105)
	at io.confluent.connect.elasticsearch.ElasticsearchWriter.tryGetIndexableRecord(ElasticsearchWriter.java:254)
	at io.confluent.connect.elasticsearch.ElasticsearchWriter.write(ElasticsearchWriter.java:232)
	at io.confluent.connect.elasticsearch.ElasticsearchSinkTask.put(ElasticsearchSinkTask.java:159)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:495)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:288)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:198)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:166)
	at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:170)
	at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:214)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[...]

PR #567, opened 2021-07-16
support STRUCT message keys

deuscapturus:
## Problem

There is currently no way to sink messages with composite keys without ignoring those keys.

Error "STRUCT is not supported as the document id."

 https://github.com/confluentinc/kafka-connect-elasticsearch/issues/193

## Solution

Convert message key STRUCT to a string for the doucment id.

- [ ] yes
- [ ] no

## Test Strategy

unit test provided for DataConverter

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -108,6 +108,8 @@ public class DataConverter {
       case INT64:
       case STRING:
         return String.valueOf(key);
+      case STRUCT:
+        return key.toString();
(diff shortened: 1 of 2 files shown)


deuscapturus:
[clabot:check]

OneCricketeer:
I feel like a better solution is to just use a Transform on the Key

For example - https://jcustenborder.github.io/kafka-connect-documentation/projects/kafka-connect-transform-common/transformations/ToJSON.html

PR #575, opened 2021-08-09
Add timeout for retries

ilanjiR:
## Problem
The retry util does not have a timeout. If the ES server is not reachable, the request never completes and the connector hands indefinitely

## Solution
Add a timeout to the retry util. 

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -367,7 +367,8 @@ public class ElasticsearchClient {
           description,
           function,
           config.maxRetries(),
-          config.retryBackoffMs()
+          config.retryBackoffMs(),
+          config.connectionTimeoutMs()
       );
     } catch (Exception e) {
       throw new ConnectException("Failed to " + description + ".", e);
--- a/src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
@@ -15,10 +15,15 @@
 
 package io.confluent.connect.elasticsearch;
 
-import java.io.IOException;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
 import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.slf4j.Logger;
@@ -106,6 +111,7 @@ public class RetryUtil {
    * @param function         the function to call; may not be null
    * @param maxAttempts      maximum number of attempts
    * @param initialBackoff   the initial backoff in ms before retrying
+   * @param connectionTimeout The length of time to wait for a response from the server
    * @param <T>              the return type of the function to retry
    * @return the function's return value
    * @throws ConnectException        if the function failed after retries
@@ -115,9 +121,11 @@ public class RetryUtil {
       String description,
       Callable<T> function,
       int maxAttempts,
-      long initialBackoff
+      long initialBackoff,
+      long connectionTimeout
   ) throws Exception {
-    return callWithRetries(description, function, maxAttempts, initialBackoff, Time.SYSTEM);
+    return callWithRetries(description, function, maxAttempts, initialBackoff,
+        connectionTimeout, Time.SYSTEM);
   }
 
   /**
@@ -136,39 +144,43 @@ public class RetryUtil {
    * @param <T>              the return type of the function to retry
    * @return the function's return value
    * @throws ConnectException        if the function failed after retries
-   * @throws Exception               if the function throws an exception
    */
   protected static <T> T callWithRetries(
       String description,
       Callable<T> function,
       int maxAttempts,
       long initialBackoff,
+      long connectionTimeout,
       Time clock
   ) throws Exception {
     assert description != null;
     assert function != null;
-
+    log.warn("in call with retries");
     int attempt = 0;
+    long backoff = computeRandomRetryWaitTimeInMillis(attempt, initialBackoff);
     while (true) {
       ++attempt;
+      ExecutorService executorService = Executors.newSingleThreadExecutor();
       try {
-        log.trace(
+        log.warn(
             "Try {} (attempt {} of {})",
             description,
             attempt,
             maxAttempts
         );
-        return function.call();
-      } catch (IOException e) {
+        Future<T> future = executorService.submit(function);
+        // Wait for connectionTimeout duration until the result is returned
+        return future.get(connectionTimeout, TimeUnit.MILLISECONDS);
+
+      } catch (Exception e) {
         if (attempt >= maxAttempts) {
           log.error("Failed to {} due to {} after total of {} attempt(s)",
               description, e.getCause(), maxAttempts, e.getMessage());
           throw new ConnectException("Failed to " + description, e);
         }
-        // Otherwise it is retriable and we should retry
-        long backoff = computeRandomRetryWaitTimeInMillis(attempt, initialBackoff);
         log.warn("Failed to {} due to {}. Retrying attempt ({}/{}) after backoff of {} ms",
             description, e.getCause(), attempt, maxAttempts, backoff);
+        executorService.shutdown();
         clock.sleep(backoff);
       }
     }
--- a/src/test/java/io/confluent/connect/elasticsearch/RetryUtilTest.java
+++ b/src/test/java/io/confluent/connect/elasticsearch/RetryUtilTest.java
@@ -72,7 +72,7 @@ public class RetryUtilTest {
     MockTime mockClock = new MockTime();
     long expectedTime = mockClock.milliseconds();
 
-    assertTrue(RetryUtil.callWithRetries("test", () -> testFunction(0), 3, 100, mockClock));
+    assertTrue(RetryUtil.callWithRetries("test", () -> testFunction(0), 3, 100,1000, mockClock));
     assertEquals(expectedTime, mockClock.milliseconds());
   }
 
@@ -80,7 +80,7 @@ public class RetryUtilTest {
   public void testCallWithRetriesSomeRetries() throws Exception {
     MockTime mockClock = spy(new MockTime());
 
-    assertTrue(RetryUtil.callWithRetries("test", () -> testFunction(2), 3, 100, mockClock));
+    assertTrue(RetryUtil.callWithRetries("test", () -> testFunction(2), 3, 100, 1000, mockClock));
     verify(mockClock, times(2)).sleep(anyLong());
   }
 
@@ -88,7 +88,7 @@ public class RetryUtilTest {
   public void testCallWithRetriesExhaustedRetries() throws Exception {
     MockTime mockClock = new MockTime();
 
-    assertTrue(RetryUtil.callWithRetries("test", () -> testFunction(4), 3, 100, mockClock));
+    assertTrue(RetryUtil.callWithRetries("test", () -> testFunction(4), 3, 100, 1000, mockClock));
     verify(mockClock, times(3)).sleep(anyLong());
   }
 
--- a/src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorBaseIT.java
+++ b/src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorBaseIT.java
@@ -29,6 +29,7 @@ import org.junit.AfterClass;
 import org.junit.Before;
 
 import java.io.IOException;
+import java.net.ConnectException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -87,8 +88,13 @@ public class ElasticsearchConnectorBaseIT extends BaseConnectorIT {
     stopConnect();
 
     if (helperClient != null) {
-      helperClient.deleteIndex(TOPIC);
-      helperClient.close();
+      try {
+        helperClient.deleteIndex(TOPIC);
+        helperClient.close();
+      } catch (ConnectException e) {
+        // Server is already down. No need to close
+      }
+
     }
   }
 
--- a/src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
+++ b/src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
@@ -18,6 +18,7 @@ package io.confluent.connect.elasticsearch.integration;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.BATCH_SIZE_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.CONNECTION_PASSWORD_CONFIG;
+import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.CONNECTION_TIMEOUT_MS_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.CONNECTION_USERNAME_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.IGNORE_KEY_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.LINGER_MS_CONFIG;
@@ -40,9 +41,12 @@ import org.junit.experimental.categories.Category;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.net.ConnectException;
+import java.sql.Time;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 @Category(IntegrationTest.class)
 public class ElasticsearchConnectorIT extends ElasticsearchConnectorBaseIT {
@@ -65,6 +69,24 @@ public class ElasticsearchConnectorIT extends ElasticsearchConnectorBaseIT {
     props.put(CONNECTION_USERNAME_CONFIG, ELASTIC_MINIMAL_PRIVILEGES_NAME);
     props.put(CONNECTION_PASSWORD_CONFIG, ELASTIC_MINIMAL_PRIVILEGES_PASSWORD);
     return props;
+  }
+
+  @Test
+  public void testStopESConnectorAndRestart() throws Exception {
+    log.info("Running testStopESConnectorAndRestart");
+    // run connector and write
+    runSimpleTest(props);
+    // stop container
+    container.stop();
+    // write some more
+    writeRecords(NUM_RECORDS);
+
+    // Wait for retry mechanism to finish
+    TimeUnit.SECONDS.sleep(20);
+
+    // Connector should fail since the server is down
+    assertEquals(connect.connectorStatus(CONNECTOR_NAME).tasks().get(0).state(), "FAILED");
+
   }
 
   @Test


src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
```diff
@@ -136,39 +144,43 @@ public static long computeRetryWaitTimeInMillis(int retryAttempts,
...
       Callable<T> function,
       int maxAttempts,
       long initialBackoff,
+      long connectionTimeout,
       Time clock
   ) throws Exception {
     assert description != null;
     assert function != null;
-
+    log.warn("in call with retries");
```
kpatelatwork, line 158: change to debug?

src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
```diff
@@ -136,39 +144,43 @@ public static long computeRetryWaitTimeInMillis(int retryAttempts,
...
             "Try {} (attempt {} of {})",
             description,
             attempt,
             maxAttempts
         );
-        return function.call();
-      } catch (IOException e) {
+        Future<T> future = executorService.submit(function);
+        // Wait for connectionTimeout duration until the result is returned
+        return future.get(connectionTimeout, TimeUnit.MILLISECONDS);
```
kpatelatwork, line 173: @ilanjiR  its a right find that the bulk isn't setting timeout but its a fix in the wrong place.  We need to instead find and fix why we are passing RequestOptions.DEFAULT. instead of a real RequestOptions with timeout and we need to fix this in other places also that are doing the same.  CC: @erdody

```
BulkResponse bulkResponse = callWithRetries(
    "retrying bulk request",
    () -> client.bulk(request, RequestOptions.DEFAULT)
);
```

kpatelatwork: It seems as per docs RequestOptions here are to override the timeouts defined when creating the client and passing DEFAULT means no override, in that case, we should find out why the timeouts defined when creating the client aren't working here.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -362,23 +347,28 @@ public void afterBulk(long executionId, BulkRequest request, Throwable failure)
...
-          config.maxRetries(),
-          config.retryBackoffMs()
-      );
-    } catch (Exception e) {
-      throw new ConnectException("Failed to " + description + ".", e);
-    }
+    return RetryUtil.callWithRetries(
+        description,
+        function,
+        config.maxRetries() + 1,
```
kpatelatwork, line 353: why +1 ?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -88,38 +89,23 @@
...
   ) {
-    ConfigCallbackHandler configCallbackHandler = new ConfigCallbackHandler(config);
-    NHttpClientConnectionManager cm = configCallbackHandler.connectionManager();
-    /*
-     * Handles closing any idle or expired connections to avoid SocketTimeoutExceptions. Expired
-     * connections occur when the server closes their half of the connection without notifying
-     * the client.
-     */
-    this.executorService = Executors.newSingleThreadScheduledExecutor();
-    executorService.scheduleAtFixedRate(
```
kpatelatwork, line 106: so where are the expired connections handled now?

src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
```diff
@@ -194,9 +177,6 @@ private PoolingNHttpClientConnectionManager configureConnectionManager() {
         cm = new PoolingNHttpClientConnectionManager(ioReactor);
       }
 
-      cm.setDefaultMaxPerRoute(config.maxInFlightRequests());
```
kpatelatwork, line 197: is this deprecated and what is the impact of removing it?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
kpatelatwork: I wonder why we needed a schedued thread for this and shouldn't HTTP connection pool be doing it automatically?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -144,13 +144,12 @@ private String getServerVersion() {
...
             )
             .setHttpClientConfigCallback(configCallbackHandler)
-            .setRequestConfigCallback(configCallbackHandler)
     );
     MainResponse response;
     String esVersionNumber = "Unknown";
     try {
       response = highLevelClient.info(RequestOptions.DEFAULT);
-      esVersionNumber = response.getVersion().toString();
+      esVersionNumber = response.getVersion().getNumber();
```
kpatelatwork, line 152: Thanks for this fix

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: The argument, maxAttempts is the overall number of attempts, including the first call.
maxRetries can be 0 (no retries needed).

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: We don't explicitly call these methods anymore. With correctly configured request/socket timeouts, this should be handled automatically by the pool.

src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
erdody: The idea here was to use defaults. Defaults should be more than enough for cloud, but now I'm thinking that there could be on-prem cases where the number of maxInFlightRequests is bigger than that (not likely, but possible). I'll update this to the max + some buffer (setting it to exactly maxInFlightRequests is not a good idea).

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
kpatelatwork: got it, if maxAttempts isnt documented , could we please document it that it includes the first call

kpatelatwork:
@erdody could we please also fix the PR description and title, given that this fix is no longer about timeout for retries but it is about making bulk calls retry outside of bulk processor

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
kpatelatwork: could you please double-check why they were added in the first place by looking at git history?  Even I think HTTP client should handle idle/broken connections but I wonder how the pool will test, in MySQL connection pool we have to give a test query.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -130,22 +116,30 @@ public ElasticsearchClient(
...
             .setHttpClientConfigCallback(configCallbackHandler)
-            .setRequestConfigCallback(configCallbackHandler)
     );
     this.bulkProcessor = BulkProcessor
-        .builder((req, lis) -> client.bulkAsync(req, RequestOptions.DEFAULT, lis), buildListener())
+        .builder(buildConsumer(), buildListener())
         .setBulkActions(config.batchSize())
         .setConcurrentRequests(config.maxInFlightRequests() - 1) // 0 = no concurrent requests
         .setFlushInterval(TimeValue.timeValueMillis(config.lingerMs()))
-        .setBackoffPolicy(
```
kpatelatwork, line 140: is this removed because we dont want bulk processor to retry at all (even the non IO exceptions?)

kpatelatwork: like what if ES is temporarily sending 429 and the bulk processor was retrying after backingoff, will afterbulk be called in that case?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: I did, it was part of the initial Jest -> BulkProcessor migration, no extra information. 
What do you mean by "how the pool will test" ? Both methods refer to connections that are not in use.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: Correct, our retry mechanism is a superset of what BulkProcessor does. 
Added a test.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: Extended the current documentation and renamed the parameter for more clarity

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
kpatelatwork: I mean we were calling closeExpiredConnections regularly but in the new code we aren't what does expiry has to do with correctly configured request/socket timeouts.

isn't expiry has to do with idle connections in pool that expired and now a new request uses that connection.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
kpatelatwork: Got it but is there any downside to not using the built in retry?

erdody: I can't think of any.
The clear downside is to have retry logic split in two different places.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -251,6 +253,7 @@ public void index(SinkRecord record, DocWriteRequest<?> request) {
         close();
       } catch (ConnectException e) {
         // if close fails, want to still throw the original exception
```
nicolasguyomar, line 255:
```suggestion
        // if close fails, want to still log the original exception
```

erdody: I think the comment is correct, we're logging the close exception, but in the line below, we're throwing the original exception.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: I'm referring to the section where we say that we have this background tasks "to avoid SocketTimeoutExceptions". 
I meant that if timeouts are correctly configured, we should be getting these exceptions during the use of the connections, and we shouldn't have to deal with extra cleanups.

BTW, if a connections is expired, it should be leased by the pool.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
nicolasguyomar: works for me indeed ! thank you

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorNetworkIT.java
```diff
@@ -0,0 +1,184 @@
...
+  public void testReadTimeout() throws Exception {
+    wireMockRule.stubFor(post(urlPathEqualTo("/_bulk"))
+            .willReturn(ok().withFixedDelay(2_000)));
+
+    connect.configureConnector(CONNECTOR_NAME, props);
+    waitForConnectorToStart(CONNECTOR_NAME, TASKS_MAX);
+    writeRecords(NUM_RECORDS);
+
+    // Connector should fail since the request takes longer than request timeout
+    await().atMost(Duration.ofMinutes(1)).untilAsserted(() ->
```
kpatelatwork, line 81: we should add a test where first retry will fail but second one will succeed

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
kpatelatwork: got it

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorNetworkIT.java
```diff
@@ -0,0 +1,184 @@
...
+            .contains("Failed to execute bulk request due to 'java.net.SocketTimeoutException: " +
+                    "1,000 milliseconds timeout on connection")
+            .contains("after 3 attempt(s)");
+
+    // 1 + 2 retries
+    verify(3, postRequestedFor(urlPathEqualTo("/_bulk")));
+  }
+
+  @Test
+  public void testTooManyRequests() throws Exception {
```
kpatelatwork, line 95: we should also add a new test for a scenario where the request to Elastic are taking a long time and the new ExecutorService has more requests than maxInFlightRequests

kpatelatwork:
@erdody Very good work overall.  I think we need to test the few cases I added and also do a performance test.  Also if possible we should document how the new request flow looks either in the Readme.md with a diagram as it took some time to figure out whats going on here.

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorNetworkIT.java
```diff
@@ -68,6 +78,123 @@ public void cleanup() {
...
+   * Transformer that blocks all incoming requests until {@link #release(int)} is called
+   * to fairly unblock a given number of requests.
+   */
+  public static class ConcurrencyTransformer extends ResponseTransformer {
+
+    private final Semaphore s = new Semaphore(0, true);
+    private final AtomicInteger requestCount = new AtomicInteger();
+
+    public ConcurrencyTransformer() {
+      s.drainPermits();
```
kpatelatwork, line 91: could you please educate me on if we have started with 0 permits then why is drainPermits required here?

erdody: We don't, leftover from a previous version where we started with a few permits. Removed.

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorNetworkIT.java
erdody: Added

src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
```diff
@@ -194,8 +177,13 @@ private PoolingNHttpClientConnectionManager configureConnectionManager() {
...
-      cm.setMaxTotal(config.maxInFlightRequests());
+      int maxPerRoute = Math.max(10, config.maxInFlightRequests() * 2);
```
kpatelatwork, line 180: @erdody could you comment on the new logic for the maxPerRoute and maxTotal calculation?

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorNetworkIT.java
```diff
@@ -0,0 +1,359 @@
...
+import static org.apache.kafka.connect.runtime.SinkConnectorConfig.TOPICS_CONFIG;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+
+public class ElasticsearchConnectorNetworkIT extends BaseConnectorIT {
+
+  @Rule
+  public WireMockRule wireMockRule = new WireMockRule(options().dynamicPort(), false);
+
+  protected static final int NUM_RECORDS = 5;
```
kpatelatwork, line 62: should we use private scope for all the variables?

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorNetworkIT.java
```diff
@@ -0,0 +1,359 @@
...
+            .willReturn(aResponse().withStatus(500))
+            .willSetStateTo("Failed"));
+
+    wireMockRule.stubFor(post(urlPathEqualTo("/_bulk"))
+            .inScenario("bulkRetry1")
+            .whenScenarioStateIs("Failed")
+            .withRequestBody(WireMock.containing("{\"doc_num\":0}"))
+            .willSetStateTo("Fixed")
+            .willReturn(okJson("{\n" +
+                    "   \"took\": 30,\n" +
```
kpatelatwork, line 150: this seems a brittle way to construct a json, should we use JsonObject or is this a pattern among the existing code?

kpatelatwork (approved):
LGTM, Thanks a lot Diego this looks awesome.

src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
```diff
@@ -104,24 +104,23 @@ public static long computeRetryWaitTimeInMillis(int retryAttempts,
    * @param description      present tense description of the action, used to create the error
    *                         message; may not be null
    * @param function         the function to call; may not be null
-   * @param maxAttempts      maximum number of attempts
+   * @param maxTotalAttempts      maximum number of total attempts, including the first call
```
rhauch, line 107: Nit: fix the indentation
```suggestion
   * @param maxTotalAttempts maximum number of total attempts, including the first call
```

src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
```diff
@@ -130,45 +129,47 @@ public static long computeRetryWaitTimeInMillis(int retryAttempts,
    * @param description      present tense description of the action, used to create the error
    *                         message; may not be null
    * @param function         the function to call; may not be null
-   * @param maxAttempts      maximum number of attempts
+   * @param maxTotalAttempts      maximum number of attempts
```
rhauch, line 132: Nit: fix indentation

src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
```diff
@@ -130,45 +129,47 @@ public static long computeRetryWaitTimeInMillis(int retryAttempts,
...
+  ) {
     assert description != null;
     assert function != null;
-
     int attempt = 0;
     while (true) {
       ++attempt;
       try {
-        log.trace(
+        log.warn(
```
rhauch, line 152: Should this still be warn?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -52,18 +52,29 @@
...
+/**
+ * Implementation notes:
+ * <ul>
+ *   <li>Based on Elasticsearch's BulkProcessor, which is responsible for batching based on size
+ *   and linger. It also limits the concurrency (max number of in-flight requests).</li>
+ *   <li>Batches are run asynchronously on a separate thread</li>
+ *   <li>Retries are processed synchronously in the original batch thread</li>
+ * </ul>
+ */
+@SuppressWarnings("checkstyle:ClassDataAbstractionCoupling")
```
rhauch, line 77: I don't think this description helps to really explain how this client is now performing the behavior. IIUC, the bulk processor was being used to actually execute the requests, but now that's not the case. Instead, the bulk processor is being used to synchronously submit batches (with retry logic) to the executor service. IOW, the bulk processor is basically just building batches and submit requests to the executor service, which actually sends the requests to ES and retries upon any failure. If the retries all fail, then the exception is reported via the atomic reference to an error, which is checked and thrown from a subsequent call to the task's `put(...)` method and that results in failure of the task.

Overall this sounds like a solid approach, but it's odd this is not described even slightly in the code or JavaDoc.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -130,22 +126,36 @@ public ElasticsearchClient(
...
+        // Disabling bulk processor retries, because they only cover a small subset of errors
+        // (see https://github.com/elastic/elasticsearch/issues/71159)
+        // We are doing retries in the async thread instead.
+        .setBackoffPolicy(BackoffPolicy.noBackoff())
         .build();
   }
 
+  private BiConsumer<BulkRequest, ActionListener<BulkResponse>> buildConsumer() {
+    return (req, lis) ->
+      bulkExecutorService.submit(() -> {
```
rhauch, line 144: The result of this submit call is not used, and this merely enqueues the operation and returns. Should this method use `bulkExecutorService.submit(...).get();` to block until the request is actually processed? If not, what is providing the backpressure to the Connect call to `SinkTask.put(...)` if the bulk processor doesn't fill up?

rhauch (commented):
Thanks, @erdody. A few comments inline below.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -362,23 +364,28 @@ public void afterBulk(long executionId, BulkRequest request, Throwable failure)
...
+    bulkExecutorService.shutdown();
+    try {
+      if (!bulkExecutorService.awaitTermination(config.flushTimeoutMs(), TimeUnit.MILLISECONDS)) {
+        bulkExecutorService.shutdownNow();
+      }
+    } catch (InterruptedException e) {
+      bulkExecutorService.shutdownNow();
+      Thread.currentThread().interrupt();
+      log.warn("Interrupted while awaiting for executor service shutdown.", e);
+    }
```
rhauch, line 388: The `config.flushTimeoutMs()` could be set to a large value, and if so that could prevent interrupting executor threads that are blocked in backoffs before a retry.

Instead, I'd suggest using a constant duration (e.g., 5 seconds) that is not configurable. This will prevent users from accidentally configuring the connector one way and then having it becomes really slow to terminate the tasks.

rhauch (approved):
Overall, most of my comments are just requests for better documentation. I'll approve pending the correction of a few areas, including the log warning and the timeout when shutting down the executor service.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: As discussed, concurrency is limited by BulkProcessor, and the result is reported via the ActionLister. Added documentation to explain all this.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: Expanded documentation.

pom.xml
```diff
@@ -207,6 +207,24 @@
...
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.20.2</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
```
Prathibha-m, line 217: Just for consistency sake and to avoid adding the overhead of managing possibly more CVE's from new deps- 
since junit is already imported in the pom, I suppose we can use that instead of assertj-core ?

also, I see that we have used `awaitility` here. I've generally seen usage of TestUtils for the same functionality https://github.com/confluentinc/ce-kafka/blob/970dc45276ba77f606cd162bfb60fdbdb0f26a02/clients/src/test/java/org/apache/kafka/test/TestUtils.java#L280

https://github.com/confluentinc/kafka-connect-kinesis/blob/770d244822fa04b4270ef5cd9ac0a6de0d862f64/src/test/java/io/confluent/connect/kinesis/integration/BaseConnectorIT.java#L138-L140

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorNetworkIT.java
```diff
@@ -0,0 +1,337 @@
...
+
+    return props;
+  }
+
+  protected void writeRecords(int numRecords) {
+    for (int i = 0; i < numRecords; i++) {
+      connect.kafka().produce(TOPIC, String.valueOf(i), String.format("{\"doc_num\":%d}", i));
+    }
+  }
+
```
Prathibha-m, line 319: Love the comprehensive tests, learning a ton from it 🙇🏽‍♀️

Prathibha-m (approved):
Thanks @Diego, LGTM! 🚢

pom.xml
erdody: Thanks. Are test dependencies included in CVE scans? 
I saw `TestUtils.waitForCondition` but having the possibility of using assertions makes the test more concise.
Otherwise, if you want useful information about why a condition was not met, you would need to use the one that has a `conditionDetailsSupplier` argument and duplicate some of the test condition logic.

Prathibha-m: Ah you're right, I doubt CVE's can come from here since it's a test dependency. Thanks @erdody 👍🏽

PR #579, opened 2021-08-18
Flush not waiting for async batches to finish

erdody:
## Problem

Record processing is asynchronous (BulkProcessor) and flush doesn't wait for batches to finish.

## Solution

Maintain the asynchronous behavior, keep track of processed records and use preCommit instead of flush to be able to report the safe offsets.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

Patch latest minor release.
Rollback should be a safe operation.


src/main/java/io/confluent/connect/elasticsearch/OffsetTracker.java
```diff
@@ -0,0 +1,180 @@
...
+  private final Map<TopicPartition, Long> maxOffsetByPartition
+          = new ConcurrentHashMap<>();
+
+  private final AtomicLong numEntries = new AtomicLong();
+
+  private static final long LIMIT = 1_000; // TODO calculate / config
+
+  // fairness is not important, things can only be blocked from one thread
+  private final ReentrantLock lock = new ReentrantLock();
+  private final Condition notFull = lock.newCondition();
```
ncliang, line 88: Any reason not to use the intrinsic lock and wait/notify API? Seems like you are locking the entire methods anyways, might be more readable to just have them as synchronized methods.

src/main/java/io/confluent/connect/elasticsearch/OffsetTracker.java
```diff
@@ -0,0 +1,180 @@
...
+      markProcessed(TRUE);
+    }
+
+    /**
+     * Provides a supplier that will determine if the record was processed or not.
+     * It does nothing is markProcessed was called before.
+     * Intended for async operations that return a Future and don't provide a callback we can
+     * hook to. (e.g. {@link ErrantRecordReporter#report(SinkRecord, Throwable)})
+     */
+    public void markProcessed(Supplier<Boolean> processedSupplier) {
```
ncliang, line 66: I am not a fan of this interface. It seems overly complex to have `markProcessed` either mark the state or provide a guarantee to mark the state some time in the future. Essentially, it is trying to do the job of marking state + Futures. Since it seems to exist for the sole purpose of `ErrantRecordReporter` , would it be easier to wrap that `Future` in a `CompletableFuture` and chain the call to mark state after it has completed?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -519,19 +549,24 @@ private void removeFromInFlightRequests(long executionDd) {
...
-        reporter.report(
-            original,
+        Future<Void> reportFuture = reporter.report(
+            original.sinkRecord,
             new ReportingException("Indexing failed: " + response.getFailureMessage())
         );
+        record.offsetState.markProcessed(reportFuture::isDone);
       }
     }
+    record.offsetState.markProcessed();
```
ncliang, line 569: Doesn't this line end up always getting called even if we went into `record.offsetState.markProcessed(reportFuture::isDone);` ? You guard against the duplicate call in the `markProcessed` method, but the code flow is still awkward.

ncliang (commented):
Good catch @erdody . Surprised this did not get caught earlier. I have a few comments, but overall approach looks good.

src/main/java/io/confluent/connect/elasticsearch/OffsetTracker.java
erdody: Thanks Nigel, I thought about that but the main problem is that the only way to wrap a `Future` in a `CompletableFuture` you either wait for the future in a separate thread, which in this case is not reasonable (this is per record) or you need to create a [separate event loop to schedule the result checking](http://www.thedevpiece.com/converting-old-java-future-to-completablefuture/), which also adds to the complexity.

src/main/java/io/confluent/connect/elasticsearch/OffsetTracker.java
```diff
@@ -0,0 +1,151 @@
...
+      processed = true;
+    }
+
+    public boolean isProcessed() {
+      return processed;
+    }
+  }
+
+  /**
+   * Partitions are no longer owned, we should release all related resources.
```
rhauch, line 73: Nit: complete the JavaDoc

src/main/java/io/confluent/connect/elasticsearch/OffsetTracker.java
```diff
@@ -0,0 +1,151 @@
...
+      }
+      maxOffsetByPartition.remove(tp);
+    });
+  }
+
+  /**
+   * This method assumes that new records are added in offset order.
+   * Older records can be re-added, and the same Offset object will be return if its
+   * offset hasn't been reported yet.
+   */
```
rhauch, line 89: Nit: complete the JavaDoc

src/main/java/io/confluent/connect/elasticsearch/OffsetTracker.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  /**
+   * @return offsets to commit
+   */
+  public synchronized Map<TopicPartition, OffsetAndMetadata> offsets() {
+    return maxOffsetByPartition.entrySet().stream()
+            .collect(toMap(
+                Map.Entry::getKey,
+                e -> new OffsetAndMetadata(e.getValue() + 1)));
```
rhauch, line 148: Add comment or mention in JavaDoc why "+1"?

src/main/java/io/confluent/connect/elasticsearch/OffsetTracker.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  /**
+   * Partitions are no longer owned, we should release all related resources.
+   */
+  public synchronized void closePartitions(Collection<TopicPartition> topicPartitions) {
+    topicPartitions.forEach(tp -> {
+      Map<Long, OffsetState> offsets = offsetsByPartition.remove(tp);
+      if (offsets != null) {
+        numEntries.getAndAdd(offsets.size());
```
rhauch, line 79: Why increment `numEntries` by the number of items in the map for the removed partition?

src/main/java/io/confluent/connect/elasticsearch/OffsetTracker.java
```diff
@@ -0,0 +1,151 @@
...
+              .computeIfAbsent(sinkRecord.kafkaOffset(), OffsetState::new);
+    } else {
+      return new OffsetState(sinkRecord.kafkaOffset());
+    }
+  }
+
+  /**
+   * @return overall number of entries
+   */
+  public long numEntries() {
```
rhauch, line 108: Can this be package protected since it's only used in tests?

Also, what is meant by "number of entries"? The total number of entries that have be processed? The number of offsets that are in `offsetsByPartition`?

Since this is only used by tests, is it worth it to maintain a separate field, especially if that field is ambiguous in its purpose? Plus, if there is a bug in computing it, then this additional complexity is definitely not worth it. Why not just make the method synchronized (like all of the others) and compute the total number of offsets?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -76,38 +89,90 @@ protected void start(Map<String, String> props, ElasticsearchClient client) {
...
+      tryWriteRecord(record, offsetState);
     }
+    partitionPauser.tryPausePartitions();
   }
 
   @Override
-  public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
-    log.debug("Flushing data to Elasticsearch with the following offsets: {}", offsets);
+  public Map<TopicPartition, OffsetAndMetadata> preCommit(Map<TopicPartition,
+          OffsetAndMetadata> currentOffsets) {
```
rhauch, line 168: Nit: the lines splits here are unconventional. Should be:
```suggestion
  public Map<TopicPartition, OffsetAndMetadata> preCommit(
          Map<TopicPartition, OffsetAndMetadata> currentOffsets
  ) {
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -263,12 +273,30 @@ public boolean hasMapping(String index) {
...
    *
    * @param record the record to index
    * @param request the associated request to send
    * @throws ConnectException if one of the requests failed
    */
-  public void index(SinkRecord record, DocWriteRequest<?> request) {
+  public void index(SinkRecord record, DocWriteRequest<?> request, OffsetState offsetState) {
+    throwIfFailed();
+
+    // TODO should we just pause partitions instead of blocking and failing the connector?
```
rhauch, line 291: Are we or are we not pausing partitions?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -263,12 +273,30 @@ public boolean hasMapping(String index) {
    * document id when either the DLQ is configured or
    * {@link ElasticsearchSinkConnectorConfig#IGNORE_KEY_CONFIG} is set to <code>false</code> because
    * they require the use of a map keyed by document id.
+   * <br>
```
rhauch, line 276: It's standard practice to use `<p>` rather than `<br>`.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -263,12 +273,30 @@ public boolean hasMapping(String index) {
...
+   * This call is usually asynchronous, but can block if:
```
rhauch, line 277: Nit: Please clarify whether this call can block if *both* or *either* of the following conditions are true.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -76,38 +89,90 @@ protected void start(Map<String, String> props, ElasticsearchClient client) {
...
 
-    this.client = client != null ? client : new ElasticsearchClient(config, reporter);
+    this.client = client != null ? client
+            : new ElasticsearchClient(config, reporter, offsetTracker);
 
     log.info("Started ElasticsearchSinkTask. Connecting to ES server version: {}",
         getServerVersion());
   }
 
+  private static class PartitionPauser {
```
rhauch, line 99: What purpose does this class provide?

rhauch (commented):
Thanks, @erdody. A few comment/questions below.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
rhauch: Nevermind. I totally missed where the `PartitionPauser` logic is.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
rhauch, line 99: It's probably better to move this to the beginning (before the member fields) or end (after the member methods) of this class.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -76,38 +89,90 @@ protected void start(Map<String, String> props, ElasticsearchClient client) {
...
+
+    /**
+     * Pause partitions if they are paused and resume condition is met.
+     * Has to be run in the connector thread.
+     */
+    private void tryPausePartitions() {
+      if (!partitionsPaused && pauseCondition.getAsBoolean()) {
+        log.debug("Pausing all partitions");
+        context.pause(context.assignment().toArray(new TopicPartition[0]));
+        context.timeout(100);
```
rhauch, line 138: Add a comment that mentions that we're setting the timeout so that the `put(...)` method will be called, even with an empty list of records.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -76,38 +89,90 @@ protected void start(Map<String, String> props, ElasticsearchClient client) {
...
+     * Has to be run in the connector thread.
+     */
+    private void tryResumePartitions() {
+      if (partitionsPaused) {
+        if (resumeCondition.getAsBoolean()) {
+          log.debug("Resuming all partitions");
+          context.resume(context.assignment().toArray(new TopicPartition[0]));
+          partitionsPaused = false;
+        } else {
+          context.timeout(100);
```
rhauch, line 125: Add a comment that mentions that we're setting the timeout so that the `put(...)` method will be called, even with an empty list of records.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -76,38 +89,90 @@ protected void start(Map<String, String> props, ElasticsearchClient client) {
...
       }
 
       logTrace("Writing {} to Elasticsearch.", record);
-
-      ensureIndexExists(convertTopicToIndexName(record.topic()));
-      checkMapping(record);
-      tryWriteRecord(record);
+      tryWriteRecord(record, offsetState);
     }
+    partitionPauser.tryPausePartitions();
```
rhauch, line 163: WDYT about calling these methods `maybePausePartitions()` and `maybeResumePartitions()`?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -76,38 +89,90 @@ protected void start(Map<String, String> props, ElasticsearchClient client) {
...
+      this.context = context;
+      this.pauseCondition = pauseCondition;
+      this.resumeCondition = resumeCondition;
+    }
+
+    /**
+     * Resume partitions if they are paused and resume condition is met.
+     * Has to be run in the connector thread.
+     */
+    private void tryResumePartitions() {
```
rhauch, line 118: Would be good to add unit tests for this class.

rhauch (commented):
Thanks, @erdody. Sorry about my previous review. A few minor suggestions and questions in addition to my earlier (unresolved) ones, but otherwise looks good.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: We are pausing partitions in case we reach the limit of offset entries.

This is another scenario, where currently we block for some time and fail the connector if the overall number of in-flight records reaches the predefined limit. 
I marked this as a TODO to limit the scope of this PR.

src/main/java/io/confluent/connect/elasticsearch/OffsetTracker.java
erdody: Good catch! We should be decrementing here. Added asserts to cover it.

src/main/java/io/confluent/connect/elasticsearch/OffsetTracker.java
erdody: Added

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: Added

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
erdody: Extracted a constant and added a comment.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
erdody: Moved to the end

src/main/java/io/confluent/connect/elasticsearch/OffsetTracker.java
erdody: Renamed and updated the javadoc

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
erdody: Added

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -325,4 +280,54 @@ private static String recordString(SinkRecord record) {
   public void close(Collection<TopicPartition> partitions) {
     offsetTracker.closePartitions(partitions);
   }
+
+  // Visible for testing
+  static class PartitionPauser {
+
+    // Kafka consumer poll timeout to set when partitions are paused, to avoid waiting for a long
+    // time (default poll timeout) to resume it.
+    private static final int PAUSE_POLL_TIMEOUT_MS = 100;
```
rhauch, line 289: `SinkTaskContext.timeout(...)` takes a `long`, so let's use a `long` here:
```suggestion
    private static final long PAUSE_POLL_TIMEOUT_MS = 100;
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -325,4 +280,54 @@ private static String recordString(SinkRecord record) {
...
+                           BooleanSupplier pauseCondition,
+                           BooleanSupplier resumeCondition) {
+      this.context = context;
+      this.pauseCondition = pauseCondition;
+      this.resumeCondition = resumeCondition;
+    }
+
+    /**
+     * Resume partitions if they are paused and resume condition is met.
+     * Has to be run in the connector thread.
```
rhauch, line 306:
```suggestion
     * Has to be run in the task thread.
```

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -325,4 +280,54 @@ private static String recordString(SinkRecord record) {
...
+          partitionsPaused = false;
+        } else {
+          context.timeout(PAUSE_POLL_TIMEOUT_MS);
+        }
+      }
+    }
+
+    /**
+     * Pause partitions if they are paused and resume condition is met.
+     * Has to be run in the connector thread.
```
rhauch, line 322:
```suggestion
     * Has to be run in the task thread.
```

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTaskTest.java
```diff
@@ -23,12 +23,7 @@
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
rhauch, line 26: We don't use wildcards for imports.

rhauch (approved):
LGTM. Thanks, @erdody.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -314,54 +352,54 @@ public boolean indexExists(String index) {
...
 
       @Override
       public void afterBulk(long executionId, BulkRequest request, BulkResponse response) {
+        List<DocWriteRequest<?>> requests = request.requests();
+
+        int itemNumber = 0;
         for (BulkItemResponse bulkItemResponse : response) {
-          handleResponse(bulkItemResponse, executionId);
+          SinkRecordAndOffset record = itemNumber < requests.size()
+                  ? requestToSinkRecord.get(requests.get(itemNumber))
```
ncliang, line 380: Am I missing something or is this always getting the first request only? I guess it doesn't really matter since you always call `markProcessed` on all requests bellow. Maybe no need to pipe through the record to `handleResponse` then?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -249,4 +275,59 @@ private static String recordString(SinkRecord record) {
...
+          context.resume(context.assignment().toArray(new TopicPartition[0]));
+          partitionsPaused = false;
+        } else {
+          context.timeout(PAUSE_POLL_TIMEOUT_MS);
+        }
+      }
+    }
+
+    /**
+     * Pause partitions if they are paused and resume condition is met.
```
ncliang, line 321:
```suggestion
     * Pause partitions if they are not paused and pause condition is met.
```

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -36,10 +36,7 @@
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.*;
```
ncliang, line 39: nit: remove star import

ncliang (approved):
Thanks @erdody . LGTM with minor comments.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
erdody: Good catch, we don't have tests for javadoc :)

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: Correct, that's leftover code from before we realized we didn't have to wait for reporter futures (because preCommit already does that). Removed.

PR #587, opened 2021-10-01
CCMSG-1376: offsets are marked as processed for individual record failures

erdody:
## Problem

When individual records in a bulk batch fail to be indexed, we still mark them as processed in OffsetTracker. 
These batches are going to be retried but offsets should not be committed past the failing records.

## Solution

Update logic to avoid marking them as processed.

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

Patch latest minor release (11.1)
Safe to rollback.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -363,12 +363,15 @@ public class ElasticsearchClient {
       public void afterBulk(long executionId, BulkRequest request, BulkResponse response) {
         List<DocWriteRequest<?>> requests = request.requests();
 
+        int idx = 0;
         for (BulkItemResponse bulkItemResponse : response) {
-          handleResponse(bulkItemResponse, executionId);
+          boolean failed = handleResponse(bulkItemResponse, executionId);
+          if (!failed && idx < requests.size()) {
+            requestToSinkRecord.get(requests.get(idx)).offsetState.markProcessed();
+          }
+          idx++;
         }
 
-        requests.forEach(req ->
-                requestToSinkRecord.get(req).offsetState.markProcessed());
         offsetTracker.updateOffsets();
 
         bulkFinished(executionId, request);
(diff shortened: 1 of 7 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -484,15 +488,18 @@ private boolean createIndex(String index) {
...
                               long executionId) {
     if (response.isFailed()) {
       for (String error : MALFORMED_DOC_ERRORS) {
         if (response.getFailureMessage().contains(error)) {
-          handleMalformedDocResponse(response);
-          reportBadRecord(response, executionId);
-          return;
+          boolean failed = handleMalformedDocResponse(response);
+          if (!failed) {
+            reportBadRecord(response, executionId);
```
erdody, line 500: Note for reviewer: I'm also fixing DLQ reporting in this scenario: only send to DLQ if we are not failing the task (and eventually retry the record).

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -363,12 +364,15 @@ public void beforeBulk(long executionId, BulkRequest request) {
       public void afterBulk(long executionId, BulkRequest request, BulkResponse response) {
         List<DocWriteRequest<?>> requests = request.requests();
 
+        int idx = 0;
         for (BulkItemResponse bulkItemResponse : response) {
-          handleResponse(bulkItemResponse, executionId);
+          boolean failed = handleResponse(bulkItemResponse, executionId);
```
ncliang, line 369: nit: I feel like this could be done more elegantly by pairing up the request and responses before iterating through them.

ncliang (approved):
Thanks @erdody . LGTM with minor nit.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
erdody: Yeah, I considered using guava's `Streams.forEachPair` but it doesn't have quite the same behavior, and also it's just a couple of lines of difference.

PR #589, opened 2021-10-22
CCMSG: Add a synchronous flushing mode to support topic mutating SMTs.

erdody:
##### Does this solution apply anywhere else?
- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

PR #592, opened 2021-11-04
Use a synchronous flush by default.

avocader:
## Problem
This change https://github.com/confluentinc/kafka-connect-elasticsearch/pull/589 introduced a new configuration parameter - `flush.synchronously`, the default value was set to `true` to avoid a breaking change.
We released this change in a minor release - `v11.1.4`.
But, the synchronous mode has worse performance than asynchronous and it's only needed for a subset of users, who use ES sink with routing SMTs(those that mutate the topic's name).

Original discussion - https://github.com/confluentinc/kafka-connect-elasticsearch/pull/589/files#r735739916

## Solution
Make asynchronous mode default(`flush.synchronously`=`false`) and have a major release, since this is a breaking change for users, who have routing SMTs.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan
Doing a major `12.0.0` release.


kpatelatwork (approved):
LGTM

mdespriee:
PS: in our case, we need to change the destination index name because we use a sharded pattern (name-YYYY-MM)

kpatelatwork:
@mdespriee it was one of the alternatives suggested and is in consideration in the roadmap for the future, we had to do it this way to be backward compatible with existing released versions.

PR #597, opened 2021-11-08
Support Elasticsearch aliases

goran-rumin:
## Problem
The connector currently (version 11.1.3) doesn't support writing to Elasticsearch alias. Aliases are a convenient way to make the management of the underlying indices easy - like mapping updates, index rollover, etc. That, combined with the inability to use topic name transforms due to async commit makes alias usage very attractive.

## Solution
Expanded index existence checks with alias existence check to prevent new index creation (and subsequent failing) when topic name corresponds to an alias.

- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [x] System tests
- [x] Manual tests

## Release Plan

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -15,7 +15,26 @@
 
 package io.confluent.connect.elasticsearch;
 
-import io.confluent.connect.elasticsearch.OffsetTracker.OffsetState;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.BiConsumer;
+
 import org.apache.http.HttpHost;
 import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.connect.data.Schema;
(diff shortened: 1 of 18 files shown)


goran-rumin:
The 3 build steps that are failing all have the same error message `We have disabled builds for pull requests created by the open source communitydue to build system maintenance.We are working to finish it soon.`, so the build is not failing due to the code change. If this is expected, please ignore this comment, if not, I would kindly ask for your assistance with the issue.

kpatelatwork:
@goran-rumin Thanks for the contribution, I will request to consider this in the pipeline for Q1 planning.

Bhashit:
@kpatelatwork was this ever included in the planning?

kpatelatwork:
@Bhashit this was considered but did not make it to the next quarter.
CC: @rajat-agarwal

Lowess:
@kpatelatwork this is also a feature we are looking for, are there any considerations to put this ticket on the roadmap so that it can be released soon (Q2/Q3)? Thank you.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -354,6 +358,39 @@ public boolean indexExists(String index) {
...
+   *
+   * @param target the target to check
+   * @return actual index name
+   * @throws ConnectException if alias does not have write index
+   */
+  public String resolveIndexIfAlias(String target) {
+    if (config.isDataStream()) {
+      return target;
+    }
+    String actualIndex = aliasWriteIndex(target);
```
snehashisp, line 387: If the target is not an alias, what will this return?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -146,8 +146,9 @@ public String version() {
 
   private void checkMapping(String index, SinkRecord record) {
     if (!config.shouldIgnoreSchema(record.topic()) && !existingMappings.contains(index)) {
-      if (!client.hasMapping(index)) {
-        client.createMapping(index, record.valueSchema());
+      String targetIndex = client.resolveIndexIfAlias(index);
```
snehashisp, line 149: There might be an issue w.r.t how the mappings are being caches here (in `existingMappings`). If the alias write index is changed then this will not create the mapping in the new write index as we are caching the name of the alias in that case (end of this method). 

Adding `targetIndex` to the cache is not that great either as it means the checkMapping will result in a call to the current write index on each put. 

Also, in hindsight there can be a race condition between a write index update for an alias and a mapping update here i.e. by the time we write the data using an alias the backing write index might have changes. Hence we might end up writing a document to an index without a mapping at that point.

snehashisp:
Thanks for the contribution @goran-rumin. I have added a couple of comments. Primarily my concern is the comment above.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
goran-rumin: Yes, ElasticSearch API will return the same index name if the target is not an alias since the target is sent as an index query parameter in the GetAliasesRequest. 
Basically:
- if the target is an index, _alias API will return only one entry
- if the target is an alias, _alias API will return related index entries

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
goran-rumin: Yes, I see, my goal here was to support the async write to index with a name different from the topic name - basically a static alias. 
Since full support is not possible as there is no efficient way to create mappings in case of full alias usage (would need to check the current write index with Elasticsearch on every write), I would suggest disabling that feature in case of alias usage. If the topic name points the connector to an alias, it needs to manage underlying indices and mappings manually. What do you think?

goran-rumin:
Thanks for reviewing @snehashisp, I added my responses to the comments. I kindly request your input on 2nd one, since it affects one of the features of the connector in case of alias usage.

PR #600, opened 2021-11-15
External version (topic offset) conflict records should skip DLQ

cjolivier01:
## Problem
Resent or out-of-order records for a given topic who have their versions set externally by the ES connector to the topic offset of the message are generating warnings and being added to the DLQ.  This isn't desired because the nature of the offset mismatch dictates that the record is obsolete anyway and may simply being resent when connect is restarted (which is the case which was seen and reproduced).

## Solution
For externally-versioned topic records, simply log to debug log and do not send to DLQ.

- [ ] yes
- [x] no

## Test Strategy
Reproduction via rolling back offset to zero and bringing connect back up, in which case, topics will be resent with "old" versions, which are <= the version in ES.  dlq topic record s are generated.
After change, no dlq topic records are generated.
This re-send of "old" topic record versions are simulated in the added unit test.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

Merge to 11.1.x, 12.x, master.  There will be an extra test in >= 11.1.x

Backporting to 11.0.x and then will ping merge to higher versions and finally master
 
N/A

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -189,7 +189,7 @@ public class DataConverter {
 
   /**
    * In many cases, we explicitly set the record version using the topic's offset.
-   * This version will, in turn, be checked by ElasticSearch and will throw a versioning
+   * This version will, in turn, be checked by Elasticsearch and will throw a versioning
    * error if the request represents an equivalent or older version of the record.
    *
    * @param request the request currently being constructed for `record`
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -187,6 +187,15 @@ public class DataConverter {
     return new String(rawJsonPayload, StandardCharsets.UTF_8);
   }
 
+  /**
+   * In many cases, we explicitly set the record version using the topic's offset
+   * This version will, in turn, be checked by ElasticSearch and will throw a versioning
+   * error if the request represents an equivalent or older version of the record.
+   *
+   * @param request the request currently being constructed for `record`
+   * @param record the record to be processed
+   * @return the (possibly modified) request which was passed in
+   */
   private DocWriteRequest<?> maybeAddExternalVersioning(
       DocWriteRequest<?> request,
       SinkRecord record
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -188,7 +188,7 @@ public class DataConverter {
   }
 
   /**
-   * In many cases, we explicitly set the record version using the topic's offset
+   * In many cases, we explicitly set the record version using the topic's offset.
    * This version will, in turn, be checked by ElasticSearch and will throw a versioning
    * error if the request represents an equivalent or older version of the record.
    *
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -40,6 +40,7 @@ import org.elasticsearch.client.indices.GetMappingsResponse;
 import org.elasticsearch.client.indices.PutMappingRequest;
 import org.elasticsearch.cluster.metadata.MappingMetadata;
 import org.elasticsearch.common.unit.TimeValue;
+import org.elasticsearch.index.VersionType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -348,10 +349,12 @@ public class ElasticsearchClient {
 
       @Override
       public void afterBulk(long executionId, BulkRequest request, BulkResponse response) {
+        List<DocWriteRequest<?>> requests = request.requests();
+        int idx = 0;
         for (BulkItemResponse bulkItemResponse : response) {
-          handleResponse(bulkItemResponse, executionId);
+          handleResponse(bulkItemResponse, requests.get(idx), executionId);
+          idx++;
         }
-
         removeFromInFlightRequests(executionId);
         numRecords.addAndGet(-response.getItems().length);
       }
@@ -412,9 +415,11 @@ public class ElasticsearchClient {
    * according to configuration (ignore or fail). Version conflicts are ignored.
    *
    * @param response    the response to process
+   * @param request     the request which generated the response
    * @param executionId the execution id of the request
    */
-  private void handleResponse(BulkItemResponse response, long executionId) {
+  private void handleResponse(BulkItemResponse response, DocWriteRequest<?> request,
+                              long executionId) {
     if (response.isFailed()) {
       for (String error : MALFORMED_DOC_ERRORS) {
         if (response.getFailureMessage().contains(error)) {
@@ -423,17 +428,35 @@ public class ElasticsearchClient {
           return;
         }
       }
-
       if (response.getFailureMessage().contains(VERSION_CONFLICT_EXCEPTION)) {
-        log.warn(
-            "Ignoring version conflict for operation {} on document '{}' version {} in index '{}'.",
-            response.getOpType(),
-            response.getId(),
-            response.getVersion(),
-            response.getIndex()
+        // For external version conflicts, response.getVersion() will be returned as -1,
+        // but we have the actual version number for this record because we set it in
+        // the request.
+        long thisRecordVersion = request.versionType() == VersionType.EXTERNAL
+                ? request.version() : response.getVersion();
+        String message = String.format("Ignoring %s version conflict for operation %s on "
+                        + "document '%s' version %d in index '%s'.",
+                request.versionType(),
+                response.getOpType(),
+                response.getId(),
+                thisRecordVersion,
+                response.getIndex()
         );
-
-        reportBadRecord(response, executionId);
+        // Now check if this version conflict is caused by external version number
+        // which was set by us (set explicitly to the topic's offset), in which case
+        // the version conflict is due to a repeated or out-of-order message offset
+        // and thus can be ignored, since the newer value (higher offset) should
+        // remain the key's value in any case.
+        if (request.versionType() != VersionType.EXTERNAL) {
+          log.warn(message);
+          // Maybe this was a race condition?  Put it in the DLQ in case someone
+          // wishes to investigate.
+          reportBadRecord(response, executionId);
+        } else {
+          // This is an out-of-order or (more likely) repeated topic offset.  Allow the
+          // higher offset's value for this key to remain.
+          log.debug(message);
+        }
         return;
       }
 
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -454,7 +454,7 @@ public class ElasticsearchClient {
           // but we have the actual version number for this record because we set it in
           // the request.
           log.debug("Ignoring EXTERNAL version conflict for operation {} on"
-                  + " document '{}' version {} in index '{}'.",
+                          + " document '{}' version {} in index '{}'.",
                   response.getOpType(),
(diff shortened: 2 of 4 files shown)


src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -187,6 +187,15 @@ private String getPayload(SinkRecord record) {
     return new String(rawJsonPayload, StandardCharsets.UTF_8);
   }
 
+  /**
+   * In many cases, we explicitly set the record version using the topic's offset
```
kpatelatwork, line 191: minor nit: missing . after offset

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -423,17 +428,35 @@ private void handleResponse(BulkItemResponse response, long executionId) {
...
-            response.getOpType(),
-            response.getId(),
-            response.getVersion(),
-            response.getIndex()
+        // For external version conflicts, response.getVersion() will be returned as -1,
+        // but we have the actual version number for this record because we set it in
+        // the request.
+        long thisRecordVersion = request.versionType() == VersionType.EXTERNAL
+                ? request.version() : response.getVersion();
+        String message = String.format("Ignoring %s version conflict for operation %s on "
```
kpatelatwork, line 437: Performance Nit: we have seen millions of records going to DLQ and VersionType is internal then we would do this string concatenation only to do log.debug

Should we instead extract just the template ` "Ignoring version conflict for operation {} on document '{}' version {} in index '{}'."`  to a variable and call it using log.warn and log.debug with replacement params

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -1,18 +1,17 @@
...
  *
- * Licensed under the Apache License, Version 2.0 (the "License"); you may not
- * use this file except in compliance with the License. You may obtain a copy of
- * the License at
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.confluent.io/confluent-community-license
```
kpatelatwork, line 8: Thanks for fixing this .

kpatelatwork (changes requested):
@cjolivier01 Good work on the PR and Thanks a lot for adding a test for this. 

Looks good to me except some minor nits, Could you please resolve them ?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
cjolivier01: Sorry, not sure what you are suggesting. Can you please clarify?

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -505,6 +505,41 @@ public void testReporterNotCalled() throws Exception {
     client.close();
   }
 
+  /**
+   * If the record version is set to VersionType.EXTERNAL (normal case for non-streaming),
+   * then same or less
+   * @throws Exception will be thrown if the test fails
+   */
+  @Test
+  public void testExternalVersionConflictReporterNotCalled() throws Exception {
```
erdody, line 514: Should we also test the code path where version type is not EXTERNAL? (data stream or ignore key)

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -505,6 +505,41 @@ public void testReporterNotCalled() throws Exception {
...
+    // Sequentially increase out record version (which comes from the offset)
+    writeRecord(sinkRecord("key0", 0), client);
+    writeRecord(sinkRecord("key0", 1), client);
+    writeRecord(sinkRecord("key0", 2), client);
+    client.flush();
+
+    // At the end of the day, it's just one record being overwritten
+    waitUntilRecordsInES(3);
+
+    // Now duplcate the last and then the one before that
```
erdody, line 532: Typo: duplcate -> duplicate

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
kpatelatwork: Something like

```
String message ="Ignoring {} version conflict for operation {} on document '{}' version {} in index '{}'.";
if (request.versionType() != VersionType.EXTERNAL) {
          log.warn(message,  request.versionType(),
                response.getOpType(),
                response.getId(),
                thisRecordVersion,
                response.getIndex());
                ........
} else {
          log.debug(message,  request.versionType(),
                response.getOpType(),
                response.getId(),
                thisRecordVersion,
                response.getIndex());	
}

```

avocader:
Thanks for the PR @cjolivier01, looks good overall!

Can we also add an integration test, something like:
- Configure an ES sink connector, produce data to a topic
- Make sure data lands to ES
- Reset topic offsets to 0
- Assert that DLQ doesn't have any records.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
kpatelatwork: +1 if the test isnt already present

cjolivier01: Note that "data stream" check wasn't until 11.1.x (I did the initial fix on master).  In fact, the data stream config value of "log" seems not introduced until 11.1.x.  But yes, let me look into how to cause this version conflict -- it should be impossible to cause on a blind insert, but maybe not on an read+update, so will probably need to manually read+write+out_of_date_update on that with max retries of 0 (see upsert in DataConverter.java): 
```java
      case UPSERT:
        return new UpdateRequest(index, id)
            .doc(payload, XContentType.JSON)
            .upsert(payload, XContentType.JSON)
            .retryOnConflict(Math.min(config.maxInFlightRequests(), 5));
```

cjolivier01:
> Thanks for the PR @cjolivier01, looks good overall!
> 
> Can we also add an integration test, something like:
> 
> * Configure an ES sink connector, produce data to a topic
> * Make sure data lands to ES
> * Reset topic offsets to 0
> * Assert that DLQ doesn't have any records.



> Thanks for the PR @cjolivier01, looks good overall!
> 
> Can we also add an integration test, something like:
> 
> * Configure an ES sink connector, produce data to a topic
> * Make sure data lands to ES
> * Reset topic offsets to 0
> * Assert that DLQ doesn't have any records.

Sure, I think this will be a good exercise for me.  I noticed that when using the docker containers in playground, bringing down connect was enough to allow the offset reset,.  However, when using platform, i have to bring down both connect and schema-registry.  And even then, bring schema-registry up and down a few times sometimes before it will allow me to reset the offset.  It is not clear to me yet why that is.  I do understand that schema-registry keeps the topic "alive", but why I need to shut it down and restart a couple of times, I don't know yet.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
kpatelatwork: @cjolivier01 if its going to make the test or code too complex then we can punt it.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -505,6 +505,41 @@ public void testReporterNotCalled() throws Exception {
...
+    writeRecord(sinkRecord("key0", 2), client);
+    writeRecord(sinkRecord("key0", 1), client);
+    client.flush();
```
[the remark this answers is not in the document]
kpatelatwork: @cjolivier01 I see changes done to testNoVersionConflict  method did you mean to add MAX_IN_FLIGHT_REQUESTS_CONFIG prop to testExternalVersionConflictReporterNotCalled test instead?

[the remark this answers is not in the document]
cjolivier01: that one is still WIP, since it's not trivial to cause the version error now

kpatelatwork:
> > Thanks for the PR @cjolivier01, looks good overall!
> > Can we also add an integration test, something like:
> > 
> > * Configure an ES sink connector, produce data to a topic
> > * Make sure data lands to ES
> > * Reset topic offsets to 0
> > * Assert that DLQ doesn't have any records.
> 
> Sure, I think this will be a good exercise for me. I noticed that when using the docker containers in playground, bringing down connect was enough to allow the offset reset. However, when using platform, I have to bring down both connect and schema-registry. And even then, bring schema-registry up and down a few times sometimes before it will allow me to reset the offset. It is not clear to me yet why that is. I do understand that schema-registry keeps the topic "alive", but why I need to shut it down and restart a couple of times, I don't know yet.

@avocader please guide if you know an easy way to reset offsets to 0?

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
cjolivier01: did a positive and negative test, the positive test being that it simply hijacks the version type and tests that handleResponse() behaves correctly by reporting the error.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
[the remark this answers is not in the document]
cjolivier01: both in now

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -505,6 +508,96 @@ public void testReporterNotCalled() throws Exception {
...
+        // Note that we don't make any attempt to reset the response version number,
+        // which will be -1 here.
+        request.versionType(VersionType.INTERNAL);
+        super.handleResponse(response, request, executionId);
+      }
+    };
+
+    causeExternalVersionConflictError(client);
+
+    // Make sure that no error was reported for either offset [1, 2] record(s)
```
kpatelatwork, line 595: minor nit: change "no error was reported" to "error was reported"

kpatelatwork (approved):
LGTM

cjolivier01:
Adding integration test in this JIRA: https://confluentinc.atlassian.net/browse/CC-16096

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
cjolivier01: done

avocader:
> > > Thanks for the PR @cjolivier01, looks good overall!
> > > Can we also add an integration test, something like:
> > > 
> > > * Configure an ES sink connector, produce data to a topic
> > > * Make sure data lands to ES
> > > * Reset topic offsets to 0
> > > * Assert that DLQ doesn't have any records.
> > 
> > 
> > Sure, I think this will be a good exercise for me. I noticed that when using the docker containers in playground, bringing down connect was enough to allow the offset reset. However, when using platform, I have to bring down both connect and schema-registry. And even then, bring schema-registry up and down a few times sometimes before it will allow me to reset the offset. It is not clear to me yet why that is. I do understand that schema-registry keeps the topic "alive", but why I need to shut it down and restart a couple of times, I don't know yet.
> 
> @avocader please guide if you know an easy way to reset offsets to 0?

We can use this method to set offsets - https://docs.confluent.io/5.5.0/clients/javadocs/org/apache/kafka/clients/admin/Admin.html#alterConsumerGroupOffsets-java.lang.String-java.util.Map-.
And this is how you can create admin client - https://github.com/confluentinc/kafka-connect-elasticsearch/blob/7120b4240505ca125f870386c6f03cefd4e23cb7/src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java#L120

PR #602, opened 2021-11-22
Add logic check seen in 11.1.x IT for # responses > # requests

cjolivier01:
## Problem

It was seen in a couple of 11.1.x integration tests that # responses can be larger than # requetss.  This check already exists in the 11.1.x codebase, so backported this "idx" logic to 11.0.x.  

I looked into trying to bring in the IT from 11.1.x, but it would pull in a lot of other changes, which seemed too drastic, since the logic change in this PR is isolated.

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


kpatelatwork (approved):
LGTM

PR #604, opened 2021-12-10
Bump up the log4j dependency

rajdangwal:
## Problem
Security team uncovered a vulnerability for log4j v2.14.0 (RCCA-5119). versions 2.15.0 or more are good to use.
The log4j dependency is coming as transitive dependency from `org.elasticsearch:elasticsearch`.

## Solution
Overridden the log4j dependency coming from `org.elasticsearch:elasticsearch` by explicitly adding v2.15.0.

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,12 +83,6 @@
             <groupId>org.elasticsearch</groupId>
             <artifactId>elasticsearch</artifactId>
             <version>${es.version}</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.apache.logging.log4j</groupId>
-                    <artifactId>log4j-api</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -83,6 +83,17 @@
             <groupId>org.elasticsearch</groupId>
             <artifactId>elasticsearch</artifactId>
             <version>${es.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.logging.log4j</groupId>
+                    <artifactId>log4j-api</artifactId>
+                </exclusion>
+            </exclusions>
```
pbadani, line 91: @rajdangwal This exclusion is redundant, if we are pinning a dependency explicitly. that will override the transitive dependency.

pbadani (approved):

PR #622, opened 2022-03-27
CCMSG-1612: Update ElasticSearch client to 7.15.2 with associated fixes

cjolivier01:
## Problem

ElasticSearch client < 7.14.0 have vulnerabilities (CVE).

## Solution

Upgrade client to 7.15.2 ("oldest" branch tag >= 7.14.0 still on elasticsearch-java repo).

This is intentionally cut against master, which we can cut a new 13.0.x branch.

Changes to the backing elasticsearch client are nontrivial and we may want to let it "bake" in a new version (13.0.x) before backporting.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
+++ b/src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
@@ -284,6 +284,21 @@ public class ElasticsearchConnectorIT extends ElasticsearchConnectorBaseIT {
   }
 
   @Test
+  public void testBackwardsCompatibilityDataStream2() throws Exception {
+    container.close();
+    container = ElasticsearchContainer.withESVersion("7.9.3");
+    container.start();
+    setupFromContainer();
+
+    runSimpleTest(props);
+
+    helperClient = null;
+    container.close();
+    container = ElasticsearchContainer.fromSystemProperties();
+    container.start();
+  }
+
+  @Test
   public void testRoutingSmtSynchronousMode() throws Exception {
     index = addRoutingSmt("YYYYMM", "route-it-to-here-${topic}-at-${timestamp}");
     props.put(FLUSH_SYNCHRONOUSLY_CONFIG, "true");
--- a/pom.xml
+++ b/pom.xml
@@ -5,10 +5,9 @@
     <parent>
         <groupId>io.confluent</groupId>
         <artifactId>common</artifactId>
-        <version>6.0.3</version>
+        <version>6.1.4</version>
     </parent>
 
-    <groupId>io.confluent</groupId>
     <artifactId>kafka-connect-elasticsearch</artifactId>
     <version>12.1.0-SNAPSHOT</version>
     <packaging>jar</packaging>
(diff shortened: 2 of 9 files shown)


kpatelatwork:
@cjolivier01 Thanks a lot for the change.

Agree with cutting 13.0.x.  are we going to test that this connector change works against ES server 7.9+ via integration tests with docker containers?

arunx2:
This will break connectivity to OpenSearch clusters.

cjolivier01:
> This will break connectivity to OpenSearch clusters.

@arunx2  Can you please elaborate?

kpatelatwork:
> This will break connectivity to OpenSearch clusters.

@arunx2 this connector does not officially support OpenSearch yet.

arunx2:
> > This will break connectivity to OpenSearch clusters.
> 
> @arunx2 Can you please elaborate?

After V7.13 the java libraries are compatible with AWS OpenSearch. https://docs.aws.amazon.com/opensearch-service/latest/developerguide/samplecode.html#client-compatibility

arunx2:
> 



> > 
> 
> @arunx2 this connector does not officially support OpenSearch yet.

I agree, But until now, it works with OpenSearch.

cjolivier01:
> > > This will break connectivity to OpenSearch clusters.
> > 
> > 
> > @arunx2 Can you please elaborate?
> 
> After V7.13 the java libraries are compatible with AWS OpenSearch. https://docs.aws.amazon.com/opensearch-service/latest/developerguide/samplecode.html#client-compatibility

@arunx2 thank you for the heads-up!  As it turns out, the checks added in 7.14+, I believe, are directly related to ES mitigating the CVE's against the elasticsearch-java client (hence the new test class *ElasticSearchMockUtil* in this PR).

I suppose for OpenSearch, you'll need to use versions < 13.x.y until such time as OpenSearch mitigates this issue.

cjolivier01:
> @cjolivier01 Thanks a lot for the change.
> 
> Agree with cutting 13.0.x. are we going to test that this connector change works against ES server 7.1+ via integration tests with docker containers?

There are already integration test backwards compatibility to as far back as 7.0.1.
All of the integration tests have been run against both 7.9.3 and 7.15.2
In the latest commit, I have switched most IT's to now use 7.15.2, there is still a back-check for 7.9.3.

arunx2:
> I suppose for OpenSearch, you'll need to use versions < 13.x.y until such time as OpenSearch mitigates this issue.

@cjolivier, I don't think it can be solved by OpenSearch. Even though OpenSearch is forked from Elasticsearch, the new client libraries checks the distribution coming from elastic.co. 
In my opinion, They are two different flavors of same search engine (v7.10) we may need a separate sink connector for OpenSearch.

kpatelatwork:
@arunx2  you are right the search engines are diverging and we will soon evaluate the possibility of a new connector or refactor this codebase.

For now, OpenSearch users can use the 11.x version or https://github.com/dmathieu/kafka-connect-opensearch is an alternative they can test and 13.x of this connector will only support the Elastic cloud as we need to upgrade the library to fix CVEs in 7.9.3 client.  

cc: @snehashisp

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
```diff
@@ -283,6 +283,21 @@ public void testBackwardsCompatibilityDataStream() throws Exception {
     container.start();
   }
 
+  @Test
+  public void testBackwardsCompatibilityDataStream2() throws Exception {
+    container.close();
+    container = ElasticsearchContainer.withESVersion("7.9.3");
```
kpatelatwork, line 289: could we refactor this method and accept the esVersion as argument and then create two tests, one for 7.1 and another for 7.9.3?

kpatelatwork:
@cjolivier01 Thanks a lot for fixing the CVEs and making the connector secure. Could you please update the PR title or description and mentioning the CVEs fixed by this PR?

Overall the change looks good, I sent a minor nit and  the build is failing, could you please fix it and it would be ready for final review?

src/test/java/io/confluent/connect/elasticsearch/integration/ElasticsearchConnectorIT.java
cjolivier01: Not following -- we already have two tests -- one for 7.0.1 and one for 7.9.3 (it may show off the screen).  I can consolidate a ferw of the lines, I suppose, although they're pretty trivial to begin with.

kpatelatwork: if we have test for 7.0.1 then no need to do what I am saying, I might not have realized this.

PR #626, opened 2022-04-12
FIX: Invalid configs can lead to validation throwing a 500 error. 

snehashisp:
## Problem
The `ConfigCallbackHandler` has a set of initialisation steps in `customizeHttpClient` which uses some of the configs. If any of the configs are incorrect It will bubble up an uncaught exception in the validator while creating the `HighLevelRestClient`, which internally initialises the handler. The validation will fail with an internal server error. Example
```
{"error_code":500,"message":"Failed to load SSL keystore  of type JKS"}
```
Although the error message is returned this should ideally have the correct response code and proper formatted output.
```
{"error_code":400,"message":"Connector configuration is invalid and contains the following 1 error(s):\nFailed to create client to verify connection. Error: Failed to load SSL keystore  of type JKS\nYou can also find the above list of errors at the endpoint `/connector-plugins/{connectorType}/config/validate`"}
```

## Solution
Capture any exceptions thrown while creating the client and return a proper exception to the user while simultaneously logging the stack trace. 

Also I have moved some of the validation that do not require the client to be initialised out of the client creation try catch block .

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/Validator.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/Validator.java
@@ -117,9 +117,9 @@ public class Validator {
       } catch (IOException e) {
         log.warn("Closing the client failed.", e);
       } catch (Throwable e) {
-        log.error("Validation failed ", e);
+        log.error("Failed to create client to verify connection. ", e);
         addErrorMessage(CONNECTION_URL_CONFIG, "Failed to create client to verify connection. "
-            + "Error: " + e.getMessage());
+            + e.getMessage());
       }
     }
 
--- a/src/main/java/io/confluent/connect/elasticsearch/Validator.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/Validator.java
@@ -100,25 +100,27 @@ public class Validator {
       return new Config(validations);
     }
 
-    try (RestHighLevelClient client = clientFactory.client()) {
-      validateCredentials();
-      validateDataStreamConfigs();
-      validateIgnoreConfigs();
-      validateKerberos();
-      validateLingerMs();
-      validateMaxBufferedRecords();
-      validateProxy();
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -100,25 +100,27 @@ public Config validate() {
...
+
+    if (!hasErrors()) {
+      // no point in connection validation if previous ones fails
+      try (RestHighLevelClient client = clientFactory.client()) {
         validateConnection(client);
-      }
-      if (!hasErrors()) {
         validateVersion(client);
+      } catch (IOException e) {
+        log.warn("Closing the client failed.", e);
```
cjolivier01, line 118: nit: I know this was here before, but seems kind of weird that it assumes that closinbg the client was the problem, doesn't it?

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -100,25 +100,27 @@ public Config validate() {
...
+      } catch (Throwable e) {
+        log.error("Validation failed ", e);
+        addErrorMessage(CONNECTION_URL_CONFIG, "Failed to create client to verify connection. "
```
cjolivier01, line 121: nit: can this have the same error message as the log in case we try to cross-check/search one of the other?

src/main/java/io/confluent/connect/elasticsearch/Validator.java
snehashisp: This is because the other exceptions in the client methods are captured in their respective validations. So only close can throw an IOException here.

PR #635, opened 2022-06-08
CCMSG-1958: Elasticsearch 8 support through compatibility mode.

snehashisp:
## Problem
The current version of the connector does not support  ES 8 and above. 

## Solution
The High level rest client 7.17 has a compatibility mode which adds some headers to work with ES 8. When using ES 8 and above set this compatibility mode as true. 

Although elastic does not explicitly say that the HLRC in compatibility mode won't work with older elastic versions (7.11 and lower) they do mention [here](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/java-rest-high-compatibility.html) that, under compatibility mode HLRC can also work with 7.11 and higher versions. Its best to only enable compatibility mode only with major version as 8 and above. 

The connector already the ability to retrieve the version so we are using it to selectively turn on compatibility mode support (or otherwise) based on the version.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -53,6 +53,7 @@ import org.elasticsearch.client.RequestOptions;
 import org.elasticsearch.client.RestClient;
 import org.elasticsearch.client.RestHighLevelClient;
 import org.elasticsearch.client.RestHighLevelClientBuilder;
+import org.elasticsearch.client.core.MainResponse;
 import org.elasticsearch.client.indices.CreateDataStreamRequest;
 import org.elasticsearch.client.indices.CreateIndexRequest;
 import org.elasticsearch.client.indices.GetIndexRequest;
@@ -101,6 +102,7 @@ public class ElasticsearchClient {
           "action_request_validation_exception"
       )
   );
+  private static final String UNKNOWN_VERSION_TAG = "Unknown";
 
   protected final AtomicInteger numBufferedRecords;
   private final AtomicReference<ConnectException> error;
@@ -114,6 +116,7 @@ public class ElasticsearchClient {
   private final Time clock;
   private final Lock inFlightRequestLock = new ReentrantLock();
   private final Condition inFlightRequestsUpdated = inFlightRequestLock.newCondition();
+  private final String esVersion;
 
   @SuppressWarnings("deprecation")
   public ElasticsearchClient(
@@ -139,7 +142,16 @@ public class ElasticsearchClient {
                 .collect(toList())
                 .toArray(new HttpHost[config.connectionUrls().size()])
         ).setHttpClientConfigCallback(configCallbackHandler).build();
-    this.client = new RestHighLevelClientBuilder(client).setApiCompatibilityMode(true).build();
+
+    esVersion = getServerVersion(client);
+
+    RestHighLevelClientBuilder clientBuilder = new RestHighLevelClientBuilder(client);
+
+    if (shouldSetCompatibilityToES8()) {
+      clientBuilder.setApiCompatibilityMode(true);
+    }
+
+    this.client = clientBuilder.build();
 
     this.bulkProcessor = BulkProcessor
         .builder(buildConsumer(), buildListener(afterBulkCallback))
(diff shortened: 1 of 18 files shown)


kpatelatwork:
Looks good so far, lets add some ITs to test with ES8 docker VM and some unit tests

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -154,6 +167,32 @@ public ElasticsearchClient(
...
+    return !version().equals(UNKNOWN_VERSION_TAG)
+        && Integer.parseInt(version().split("\\.")[0]) >= 8;
+  }
+
+  private String getServerVersion(RestClient client) {
+    RestHighLevelClient highLevelClient = new RestHighLevelClientBuilder(client).build();
+    MainResponse response;
+    String esVersionNumber = UNKNOWN_VERSION_TAG;
+    try {
+      response = highLevelClient.info(RequestOptions.DEFAULT);
```
sudeshwasnik, line 185: highLevelClient.info itself internally will check useAPICompatibility ([here](https://github.com/elastic/elasticsearch/blob/312ddd6b6b41ce0dd516b7b3a197792ea3c77cf9/client/rest-high-level/src/main/java/org/elasticsearch/client/RestHighLevelClient.java#L1371-L1373)), 
just checking if `info` depends on useAPICompatibility flag in client itself.

sudeshwasnik: nit - there is a slight chance we may be checking only one ES server's version - [here](https://github.com/elastic/elasticsearch/blob/312ddd6b6b41ce0dd516b7b3a197792ea3c77cf9/client/rest/src/main/java/org/elasticsearch/client/RestClient.java#L296), the server corresponding to one of `connection.url`.

sudeshwasnik: QQ - what if we set ApiCompatibility = true by default for every request? would those requests be incompatible with ES version <7.11 ?

snehashisp: > highLevelClient.info itself internally will check useAPICompatibility ([here](https://github.com/elastic/elasticsearch/blob/312ddd6b6b41ce0dd516b7b3a197792ea3c77cf9/client/rest-high-level/src/main/java/org/elasticsearch/client/RestHighLevelClient.java#L1371-L1373)),
just checking if info depends on useAPICompatibility flag in client itself.

From what I understand compatibility mode sends some extra headers so that ES8 recognises and works in ES7 mode. Not all calls requires the use of headers like the info call which is most likely the same between versions. 

> nit - there is a slight chance we may be checking only one ES server's version - [here](https://github.com/elastic/elasticsearch/blob/312ddd6b6b41ce0dd516b7b3a197792ea3c77cf9/client/rest/src/main/java/org/elasticsearch/client/RestClient.java#L296), the server corresponding to one of connection.url

I expect that if the customer is running ES in a cluster it won't be a mismatch of different ES versions. 

> QQ - what if we set ApiCompatibility = true by default for every request? would those requests be incompatible with ES version <7.11 

Elastic does not officially state that its incompatible though they do mention that it will work with 7.11 and above which says to me that older ones may not have been fully tested to work.

src/test/java/io/confluent/connect/elasticsearch/helper/ElasticsearchContainer.java
```diff
@@ -375,9 +376,10 @@ private void buildImage(DockerfileBuilder builder) {
       ArrayList<Integer> versionsInt = getImageVersion();
       log.info("Building Elasticsearch image with SSL configuration");
       builder
+          .user("root")
           .copy("instances.yml", CONFIG_SSL_PATH + "/instances.yml")
           .copy("start-elasticsearch.sh", CONFIG_SSL_PATH + "/start-elasticsearch.sh");
-      if (versionsInt.get(0) >= 7 && versionsInt.get(1) >= 15) {
+      if (versionsInt.get(0) == 8 || versionsInt.get(0) == 7 && versionsInt.get(1) >= 15) {
```
sudeshwasnik, line 382: minor nit - 
```suggestion
      if (versionsInt.get(0) == 8 || (versionsInt.get(0) == 7 && versionsInt.get(1) >= 15)) {
```

sudeshwasnik (approved):
Looks good! 

QQ - would this be a candidate for minor or major release?

src/test/java/io/confluent/connect/elasticsearch/helper/ElasticsearchContainer.java
snehashisp: Yes. This will be a major release.

PR #639, opened 2022-06-22
Add Version Type & Field Configuration

jokefaker:
## Problem
* Not able to custom version type 
* Not able to set custom version when use external version type

## Solution
* add version.type configuration: internal, external, external_gte
* add version.type.field configuration to support read version value from payload 
* use version.type and version.type.field in data convert 

- [ ] yes
- [x] no

## Test Strategy
Testing if version.type and version.type.field is set, and check if IndexRequest is correct 

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan
This feature is backwards compatible.
Merging to master
Rolling back might result in validation error for people that use version.type or version.type.field


blackmad:
is there a possibility of this being merged?

PR #640, opened 2022-06-30
RCCA-7507 : Adding some more logs to understand the exception.

pkuntalteam1:
## Problem
lcc-gq1xgm Elasticsearch Sink Connector getting document already exists errors

## Solution
Enabling more logging to understand the NPE 

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -538,6 +538,16 @@ public class ElasticsearchClient {
                   response.getVersion(),
                   response.getIndex()
           );
+
+          log.trace("{} version conflict for operation {} on document '{}' version {}"
+                          + " in index '{}' and stacktrace '{}'",
+                  request != null ? request.versionType() : "UNKNOWN",
+                  response.getOpType(),
+                  response.getId(),
+                  response.getVersion(),
+                  response.getIndex(),
+                  response.getFailure().getCause().getStackTrace()
+          );
           // Maybe this was a race condition?  Put it in the DLQ in case someone
           // wishes to investigate.
           reportBadRecord(response, executionId);
--- a/src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
@@ -170,9 +170,9 @@ public class RetryUtil {
 
         log.warn("Failed to {} due to {}. Retrying attempt ({}/{}) after backoff of {} ms",
             description, e.getCause(), attempt, maxTotalAttempts, backoff);
-        log.trace("Failed to {} due to {} and stacktrace {}. "
-                         + "Retrying attempt ({}/{}) after backoff of {} ms",
-                         description, e.getCause(), e.getStackTrace(), attempt,
+        log.trace("Failed to {} due to exception :: {} and its stacktrace {}  "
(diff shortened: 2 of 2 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -538,6 +538,16 @@ protected boolean handleResponse(BulkItemResponse response,
                   response.getVersion(),
                   response.getIndex()
           );
+
+          log.trace("{} version conflict for operation {} on document '{}' version {}"
```
pbadani, line 542: let's have this at info level. otherwise trace logging needs to be enabled.

pkuntalteam1: In this case, we can modify warn itself but idea is to switch it off immediately after collecting logs.

pbadani: ok

pbadani (approved):

src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
```diff
@@ -170,6 +170,10 @@ protected static <T> T callWithRetries(
 
         log.warn("Failed to {} due to {}. Retrying attempt ({}/{}) after backoff of {} ms",
             description, e.getCause(), attempt, maxTotalAttempts, backoff);
+        log.trace("Failed to {} due to {} and stacktrace {}. "
+                         + "Retrying attempt ({}/{}) after backoff of {} ms",
+                         description, e.getCause(), e.getStackTrace(), attempt,
+                         maxTotalAttempts, backoff);
```
snehashisp, line 176:
```suggestion
        log.trace("Retrying {} attempt ({}/{}) after backoff of {} ms. Failed due to {}",
                         description, attempt, maxTotalAttempts, backoff, e);
```
Im not sure if `e.getStackTrace()` will print the traces of of causes. Best to just log e.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -538,6 +538,16 @@ protected boolean handleResponse(BulkItemResponse response,
...
+                          + " in index '{}' and stacktrace '{}'",
+                  request != null ? request.versionType() : "UNKNOWN",
+                  response.getOpType(),
+                  response.getId(),
+                  response.getVersion(),
+                  response.getIndex(),
+                  response.getFailure().getCause().getStackTrace()
```
snehashisp, line 549: Are we sure that getFailure won't be null here. Better to just print `response.getFailure()` instead of `response.getFailure().getCause().getStackTrace()` and change the log from `stacktrace` to `failures`.

src/main/java/io/confluent/connect/elasticsearch/RetryUtil.java
pkuntalteam1: Ok, agree we can get more info too.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
pkuntalteam1: if (response.isFailed())  is being checked at the start of the code block which will ensure failure is not null.

PR #647, opened 2022-07-22
CCMSG-2073: Throw data exception when non json object payload is present.

snehashisp:
## Problem
If a non json object is preset in the payload, during data stream insert we end up throwing a `ClassCastException`, while trying to insert the record timestamp if not user given timestamp field is not used. 

Note that if the user is trying to obtain the timestamp from the payload the `jsonNode.has` condition returns null if `jsonNode` is not an `ObjectNode` so the cast is not executed. In such cases, it will most likely fail to be inserted as ES only supports Json Objects as top level elements. This is the current behaviour and is left unchanged. 

## Solution
Do an explicit check and throw a data exception if not an object node. Also in the previous condition if the json node does not have the user given timestamp field the resulting record may be skipped entirely so added some debug statements for future debugging ease.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -207,8 +207,8 @@ public class DataConverter {
       JsonNode jsonNode = objectMapper.readTree(payload);
 
       if (!jsonNode.isObject()) {
-        throw new DataException("Top level payload contains data of Json type " +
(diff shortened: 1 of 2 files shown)


sidd1809 (approved):
LGTM! Thanks

PR #655, opened 2022-09-26
CCMSG-1698: Remove snakeyml dependency

snehashisp:
## Problem
Snakeyml has a vulnerability in the way it loads which can lead to RCEs. More info here https://securitylab.github.com/research/swagger-yaml-parser-vulnerability/. 

## Solution
Since it's a transitive dependency on a submodule of ES seeing if we can exclude this from the connector. 

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/licenses.html
+++ b/licenses.html
@@ -1,4 +1,4 @@
- <HTML><HEAD><style>
+<HTML><HEAD><style>
 table {
     border-collapse: collapse;
     width: 100%;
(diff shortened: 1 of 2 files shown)


arihant-confluent (approved):
LGTM. Assuming Unit tests and integration tests are working fine.

PR #667, opened 2022-12-14
CCMSG-2371: Remove key being logged.

snehashisp:
## Problem
Key is being logged which can contain PII.

## Solution
Remove key being logged.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -144,10 +144,9 @@ public class DataConverter {
         default:
           throw new DataException(
               String.format(
-                  "%s with key of %s and null value encountered (to ignore future records like"
+                  "%s has a null value (to ignore future records like"
                       + " this change the configuration property '%s' from '%s' to '%s')",
                   recordString(record),
-                  record.key(),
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -144,10 +144,9 @@ public DocWriteRequest<?> convertRecord(SinkRecord record, String index) {
         default:
           throw new DataException(
               String.format(
-                  "%s with key of %s and null value encountered (to ignore future records like"
+                  "%s has a null value (to ignore future records like"
```
pkuntalteam1, line 147: @snehashisp : we are still logging entire record .. ig we should avoid logging records too

snehashisp: do you meant `recordString`. That only records the metadata of the record not the entire record. 
```private static String recordString(SinkRecord record) {
    return String.format(
        "record from topic=%s partition=%s offset=%s",
        record.topic(),
        record.kafkaPartition(),
        record.kafkaOffset()
    );
  }````

pkuntalteam1 (approved):

PR #669, opened 2023-01-18
[CCMSG-2395] Restrict Customer data logging to Confluent infra

sp-gupta:
## Problem

ElasticSearch Connector is reporting exceptions with customer data. We need to find mechanism to prevent customer data to be reported in our logs.
https://confluentinc.atlassian.net/browse/CCMSG-2395

## Solution
Connect Runtime [Log Reporter](https://github.com/confluentinc/ce-kafka/blob/master/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/errors/LogReporter.java#L66) logs everything under the context. This is why we are even seeing customer DLQ logs in OpenSearch.

Checked all the APIs under `response.getFailure()` for what they are returning.

Screenshot 2023-01-24 at 12 24 20 PM

The APIs don't provide enough capability to drill down and log the exception explicitly without previewing any customer data.
After careful consideration, we are opting Status, Index and Doc Id (from the ElasticSearch response) to be logged for our infra and Customer. This will help customer to identify the document for which the error occurred. 

**Note:** We cannot consider `cause()` and `message()` APIs since they still log the value of the customer data field.

This shows how we are reporting error today:
Screenshot 2023-01-25 at 11 01 16 AM

This shows how we'll report error after this change:
Screenshot 2023-01-25 at 12 33 31 PM

**Drawback:** Customer doesn't know the exact field in the record for which the error happened.

- [ ] yes
- [x] no

## Test Strategy
Tested it via docker playground. Screenshots above

##### Testing done:
- [ ] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -480,8 +480,12 @@ public class ElasticsearchClient {
    */
   private void handleMalformedDocResponse(BulkItemResponse response) {
     String errorMsg = String.format(
-        "Encountered an illegal document error '%s'. Ignoring and will not index record.",
-        response.getFailureMessage()
+        "Encountered an illegal document error -> Response status: '%s',\n"
+                + "Index: '%s',\n Document Id: '%s'. \n"
+                + "Ignoring and will not index record.",
+        response.getFailure().getStatus(),
+        response.getFailure().getIndex(),
+        response.getFailure().getId()
     );
     switch (config.behaviorOnMalformedDoc()) {
       case IGNORE:
@@ -493,15 +497,23 @@ public class ElasticsearchClient {
       case FAIL:
       default:
         log.error(
-            "Encountered an illegal document error '{}'. To ignore future records like this,"
-                + " change the configuration '{}' to '{}'.",
-            response.getFailureMessage(),
+            "Encountered an illegal document error -> Response status: '{}',\n "
+                    + "Index: '{}',\n Document Id: '{}'\n"
+                    + " To ignore future records like this,"
+                    + " change the configuration '{}' to '{}'.",
+            response.getFailure().getStatus(),
+            response.getFailure().getIndex(),
+            response.getFailure().getId(),
             ElasticsearchSinkConnectorConfig.BEHAVIOR_ON_MALFORMED_DOCS_CONFIG,
             BehaviorOnMalformedDoc.IGNORE
         );
         error.compareAndSet(
             null,
-            new ConnectException("Indexing record failed.", response.getFailure().getCause())
+            new ConnectException(
+                    "Indexing record failed -> Response status: "
+                    + response.getFailure().getStatus()
+                    + ",\n Index: " + response.getFailure().getIndex()
+                    + ",\n Document Id: " + response.getFailure().getId())
         );
     }
   }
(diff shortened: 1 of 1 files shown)


tunnelshade (approved):

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -493,16 +497,16 @@ private void handleMalformedDocResponse(BulkItemResponse response) {
...
+                    + "Index: '{}',\n Document Id: '{}'\n"
+                    + " To ignore future records like this,"
+                    + " change the configuration '{}' to '{}'.",
+            response.getFailure().getStatus(),
+            response.getFailure().getIndex(),
+            response.getFailure().getId(),
             ElasticsearchSinkConnectorConfig.BEHAVIOR_ON_MALFORMED_DOCS_CONFIG,
             BehaviorOnMalformedDoc.IGNORE
         );
-        error.compareAndSet(
```
subhashiyer9, line 502: What is the impact of removing this statement?
Do we not need it?

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -554,9 +558,13 @@ private synchronized void reportBadRecord(BulkItemResponse response, long execut
           ? sinkRecords.get(response.getItemId())
           : null;
       if (original != null) {
+        // log only status, index and document id for record failure
         reporter.report(
             original,
-            new ReportingException("Indexing failed: " + response.getFailureMessage())
+            new ReportingException(
```
subhashiyer9, line 564: Is there a possibility of `response.getFailure()` being null?
If yes, then better to check for null before accessing its status/index/id..

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -554,9 +558,13 @@ private synchronized void reportBadRecord(BulkItemResponse response, long execut
...
+                    "Indexing failed -> Response status: " + response.getFailure().getStatus()
+                    + ",\n Index: " + response.getFailure().getStatus()
```
subhashiyer9, line 566:
```suggestion
                    + ",\n Index: " + response.getFailure().getIndex()
```

subhashiyer9 (commented):
@sp-gupta Thanks for the changes. 
Changes looks good except few minor fixes. Please have a look at them. Thanks

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
sp-gupta: Thanks for pointing this

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
sp-gupta: This is being handled [here](https://github.com/confluentinc/kafka-connect-elasticsearch/blob/2604b9929bfe8c743b5dde2ae2a92330738ecea5/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java#L562) & [here](https://github.com/confluentinc/kafka-connect-elasticsearch/blob/2604b9929bfe8c743b5dde2ae2a92330738ecea5/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java#L570).

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
sp-gupta: Thanks Subhash. After careful consideration again, I think we should keep this. I missed the fact that this error might be getting used somewhere else in the code. It is necessary to set the connect exception properly. I have made the necessary changes. Please review.

subhashiyer9 (approved):

PR #697, opened 2023-05-09
Introduce optional external.version.header config

rjosal-indeed:
Introduce optional external.version.header config to use for ES external version instead of kafka offset for non-datastream indices

## Problem

After changing an index mapping, we need to reindex.  To do this, we create a new index, run a batch job with all the existing input documents, while at the same time streaming realtime updates, ensuring we do not miss any documents.  Today both can't run at the same time because we might overwrite a new streamed update with a batch index.

## Solution

ES external versioning can be used to compare document versions for idempotency.  We will be able to pull a version out of the document into a kafka header and use the same version in the batch job.  We choose header so that deletes with null values are also idempotent.

- [ x] yes
- [ ] no

##### If yes, where?
There are a couple of other uses cases that may be solved by this control.
https://github.com/confluentinc/kafka-connect-elasticsearch/issues?q=is%3Aissue+is%3Aopen+external+version

##### Testing done:
- [x ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

If merged, we will run this connector in Confluent Cloud.  Else, we will run the fork.  As an optional feature, with a default that does not affect existing usages, it is backwards compatible.

master
 
n/a

Thank you, and please pick this apart!

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -25,10 +25,14 @@ import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.data.Time;
 import org.apache.kafka.connect.data.Timestamp;
+import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.errors.DataException;
+import org.apache.kafka.connect.header.Header;
 import org.apache.kafka.connect.json.JsonConverter;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.storage.Converter;
+import org.apache.kafka.connect.storage.HeaderConverter;
+import org.apache.kafka.connect.storage.SimpleHeaderConverter;
 import org.elasticsearch.action.DocWriteRequest;
 import org.elasticsearch.action.delete.DeleteRequest;
 import org.elasticsearch.action.index.IndexRequest;
@@ -52,6 +56,7 @@ public class DataConverter {
   private static final Logger log = LoggerFactory.getLogger(DataConverter.class);
 
   private static final Converter JSON_CONVERTER;
+  private static final HeaderConverter HEADER_CONVERTER = new SimpleHeaderConverter();
   protected static final String MAP_KEY = "key";
   protected static final String MAP_VALUE = "value";
 
@@ -202,7 +207,24 @@ public class DataConverter {
   ) {
     if (!config.shouldIgnoreKey(record.topic())) {
       request.versionType(VersionType.EXTERNAL);
-      request.version(record.kafkaOffset());
+      if (config.hasExternalVersionHeader()) {
+        final Header versionHeader = record.headers().lastWithName(config.externalVersionHeader());
+        final byte[] versionValue = HEADER_CONVERTER.fromConnectHeader(
+                record.topic(),
+                versionHeader.key(),
+                versionHeader.schema(),
+                versionHeader.value()
+        );
+        try {
+          //fromConnectHeader byte output is UTF_8
+          request.version(Long.parseLong(new String(versionValue, StandardCharsets.UTF_8)));
+        } catch (NumberFormatException e) {
+          throw new ConnectException("Error converting to long: "
+                  + new String(versionValue, StandardCharsets.UTF_8), e);
+        }
+      } else {
+        request.version(record.kafkaOffset());
+      }
     }
 
     return request;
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
```diff
@@ -249,7 +249,12 @@ private DocWriteRequest<?> maybeAddExternalVersioning(
   ) {
     if (!config.isDataStream() && !config.shouldIgnoreKey(record.topic())) {
       request.versionType(VersionType.EXTERNAL);
-      request.version(record.kafkaOffset());
+      if (config.hasExternalVersionHeader()) {
+        request.version((Long) record.headers().lastWithName(
+                config.externalVersionHeader()).value());
```
rjosal-indeed, line 254: When we put this in service, we noticed that even though we used an SMT to write a `long` type header, it didn't come out that way in this code.  We got it to work by using `org.apache.kafka.connect.storage.SimpleHeaderConverter #fromConnectHeader` and then reading it as a String and parsing it to a Long.  Does that sound correct or did we misunderstand something about kafka connect headers?

sp-gupta, line 254: @rjosal-indeed 
Thanks for putting changes in PR.

While our testing, we faced this issue, where kafka header values were being treated as bytes instead of string, due to which we were getting this exception:

`java.lang.ClassCastException: class java.lang.Byte cannot be cast to class java.lang.Long (java.lang.Byte and java.lang.Long are in module java.base of loader 'bootstrap')`

Some people may resolve it by using an SMT but we can also generalise the changes. 

Few suggestions:

- Keep `request.version` method inside a try/catch block and throwing error wrapped under `ConnectException`. This can ensure graceful failure of the connector.

- Generalise `Object` to `Long` conversion (like this):
`request.version(Long.parseLong(
            String.valueOf(
                record.headers().lastWithName(
                    config.externalVersionHeader()).value().toString())));`

- Change the base branch of the PR to `10.0.x`.
                         
Let us know if you have any comments. Thanks.

rjosal-indeed: Hey thanks for testing and reviewing!  I actually had a comment above about using `SimpleHeaderConverter#fromConnectHeader` to help ensure parsing without assuming schema - can I request guidance/confirmation to use the parseLong approach, or the formatted SimpleHeaderCoverter utility?

sp-gupta: Hey @rjosal-indeed 
Apologies, I missed this before.

Would it be possible to elaborate it? I am not sure if I quite understood how you used SimpleHeaderConverter?

Thanks!

rjosal-indeed: yes, the code for this lines with the connect header converter looks like:
```
        final Header versionHeader = record.headers().lastWithName(config.externalVersionHeader());
        final byte[] versionValue = HEADER_CONVERTER.fromConnectHeader(
                record.topic(),
                versionHeader.key(),
                versionHeader.schema(),
                versionHeader.value()
        );
        request.version(Long.parseLong(new String(versionValue)));
```
where HEADER_CONVERTER is an constant `new SimpleHeaderConverter()`

sp-gupta: Hey @rjosal-indeed 
Great! Thank you so much, this makes sense. 

I think we can go ahead with this change. Also, can you refer to the suggestions mentioned [here](https://github.com/confluentinc/kafka-connect-elasticsearch/pull/697#discussion_r1400510474) (ignore 2nd point) and make the changes accordingly?

rjosal-indeed: @sp-gupta  I worked for awhile on trying to backport this to 10.0.x and ended up failing.  10.0.x is missing a critical element in external versioning usage for offset which was extended here.  I would have to backport significant new functionality from 11.0.x to get it to work, which presumably wasn't backported for some reason, so feels a bit risky.  I can successfully apply this patch to 11.0.x.

I wonder generally though if backporting is expected for new features?  CONTRIBUTING.md indicates this is generally for bugfixes
> If your code changes are essentially bug fixes that make sense to backport to existing releases

This feature I'd expect to have documentation updates on old releases as well as making initial release announcements retroactively incomplete.  Thoughts?

sp-gupta: Hey @rjosal-indeed 
Thanks for the changes. 

As per discussion with the team, we can change the base branch to 11.0.x and we don't need to backport the changes to 10.0.x (since this is not a bug fix).

Yes, Backporting is **not** expected for new features. Documentation updates should cover informing customers about this feature.

Please go ahead and change the base branch to 11.0.x, after which I'll approve the PR.

rjosal-indeed: @sp-gupta ready for review, thanks!

sp-gupta: Thanks @rjosal-indeed 
There was one import missing due to which the build was failing. I have corrected it. 

Again, Thank you so much for your contribution. Really appreciate!!

sp-gupta (approved):

PR #742, opened 2024-01-01
add 'data.stream.namespace' configuration property

fbaligand:
## Problem
Currently, we cannot fully customize the target data stream.
More precisely, the namespace part of the data stream is necessarily the topic.
This is especially a problem, when you want that a connector consumes several topics for only one target data stream.
More simply, I want to be able to define target data stream name, regardless of the input topic name.

## Solution
This PR aims to add a new configuration property named 'data.stream.namespace'.
This property aims to customize data stream "namespace" part (as its name says).
This is an optional setting, that have `${topic}` as default value ; so that, by default, it works as currently, no breaking change.
By the way, user is able to define a custom namespace that contains topic name like that:
`my_prefix_${topic}`

- [ ] yes
- [x ] no

##### Testing done:
- [ x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

I guess that you (Confluent Inc.) is responsible for release plan.

I am merging to master
 
It is safe to revert or rollback

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -520,7 +520,7 @@ public class ElasticsearchClient {
   /**
    * Creates a data stream. Will not recreate the data stream if it already exists.
    *
-   * @param dataStream the data stream to create given in the form {type}-{dataset}-{topic}
+   * @param dataStream the data stream to create given in the form {type}-{dataset}-{namespace}
    * @return true if the data stream was created, false if it already exists
    */
   private boolean createDataStream(String dataStream) {
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -332,29 +332,46 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
   private static final String KERBEROS_KEYTAB_PATH_DEFAULT = null;
 
   // Data stream configs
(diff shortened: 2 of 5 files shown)


michele-brambilla:
I might be a bit naive, but can't the same result be achieved using the RegExRouter SMT?

fbaligand:
@michele-brambilla 
I know this feature and I use it.
But there is a blocking issue doing that: 
Because the Elasticsearch document id is computed with "topic-partition-offset", if you want to send several topics to one Elasticsearch index, then you can have conflicts with document ID (two records that are from different topics but have the same partition and offset).
Because of that, I have errors in my connector.

michele-brambilla:
@fbaligand thanks for the explanation. I stumbled on your ticket while looking for the exact opposite: the possibility of leveraging information about the data stream already stored in the message.

fbaligand:
@michele-brambilla 
In the meantime, I realized yesterday that this PR has been copied (exactly the same code) and merged here:
https://github.com/confluentinc/kafka-connect-elasticsearch/pull/802

I'm not fan of the practice (copy/paste my code in another PR), but I'm happy to see that the feature has been added and is now released.

fbaligand:
As the feature is now available (thanks to https://github.com/confluentinc/kafka-connect-elasticsearch/pull/802 PR), I close this PR.

srishti-saraswat:
@fbaligand Thanks for this PR . Changes from this PR have been cherry-picked and merged here - https://github.com/confluentinc/kafka-connect-elasticsearch/pull/802

fbaligand:
Hi @srishti-saraswat, indeed,  I noticed it. 
Is there a special reason to have cherry-picked the code in another PR, instead of merging this one?

srishti-saraswat:
Yes, due to our internal tools and regulations, we were facing difficulties in pushing changes to the external fork, which prevented us from using the original PR.

fbaligand:
Thanks for the explanation.

Issue #751, opened 2024-02-19
add support for index templates other than logs and metrics as types when using data streams


maxgruber19:
In most real world scenarios a user is not in control over the default index templates "logs" and "metrics". There are many usecases where the user wants to define a custom template. In my case I want to use the data retention feature to delete messages stored in elasticsearch after a certain amount of time.

Currently this connector is able to use the predefined index templates "logs" and "metrics" when data streams are enabled. User created templates are not supported yet.

I'd like to contribute some lightweight lines to make the connector support custom templates as well.

I guess issue #700 tried to do the same but I didn't find a related pullrequest or reactions on that thread.

maxgruber19:
closed with pr https://github.com/confluentinc/kafka-connect-elasticsearch/pull/804

PR #752, opened 2024-02-19
index templates other than logs and metrics when using data streams

maxgruber19:
index templates other than logs and metrics when using data streams #751 

## Problem
In most real world scenarios a user is not in control over the default index templates "logs" and "metrics". There are many usecases where the user wants to define a custom template. In my case I want to use the data retention feature to delete messages stored in elasticsearch after a certain amount of time.

Currently this connector is able to use the predefined index templates "logs" and "metrics" when data streams are enabled. User created templates are not supported yet.

## Solution
Defining the type of the data stream as String to make it flexible for custom inputs. The enum will not be removed to still be able to use the recommender when deploying the connector via control-center.

- [ ] yes
- [ ] no

## Test Strategy
Started a local cluster via docker-compose (cp-quickstart and elastic+kibana) and installed the connector.zip in the cluster. Both connectors with "logs" and connectors with "custom" as data stream types were running perfectly fine after creating the "custom" index template in Kibana. 

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [x] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorConfig.java
@@ -350,7 +350,8 @@ public class ElasticsearchSinkConnectorConfig extends AbstractConfig {
           + "The default is %s which indicates the connector will write "
           + "to regular indices instead. If set, this configuration will "
           + "be used alongside %s to construct the data stream name in the form of "
-          + "{``%s``}-{``%s``}-{topic}.",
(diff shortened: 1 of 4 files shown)


ge1sterfahrer:
Having this feature would massively help us in day to day business. What's your opinion on this? @sp-gupta

srishti-saraswat:
@maxgruber19 Thanks for this PR . Changes from this PR have been cherry-picked and merged here - https://github.com/confluentinc/kafka-connect-elasticsearch/pull/804

PR #776, opened 2024-04-23
[CC-26337] Version bump for Jackson Databind package

sp-gupta:
## Problem
https://confluentinc.atlassian.net/browse/CC-26337

Few issues with updating `common` package:

1. Usually in `common` repo, artifacts are created as soon as a commit is merged to a feature branch. But 6.0.x is not migrated to Semaphore. Hence, no artifacts are generated for this branch. Also, there is no provision to run connect-releases job on this repo. Hence migrated the common package to 6.1.x branch.

2. The converters (Avro, Json etc) use `confluent.version` parameter which translates to the common package version (6.1.16-32 in this case). Upon checking the maven packages available for [Kafka Avro converter](https://mvnrepository.com/artifact/io.confluent/kafka-connect-avro-converter), it is available till 6.1.15 
So we would need to pin the converter artifacts to 6.1.15 version. 

## Solution
We can pin the jackson version to 2.16.0 in the pom. Created a new ticket [here](https://confluentinc.atlassian.net/browse/CC-26830) to bump to new common version and remove jackson changes from this pom once the CP release is done at the end of the quarter.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

Refer to the comment: https://confluentinc.atlassian.net/browse/CC-26337?focusedCommentId=2591491

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.confluent</groupId>
         <artifactId>common</artifactId>
-        <version>6.1.16-32</version>
+        <version>6.0.0</version>
     </parent>
 
     <groupId>io.confluent</groupId>
@@ -52,14 +52,14 @@
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
         <commons.codec.version>1.15</commons.codec.version>
         <snakeyaml.version>1.28</snakeyaml.version>
-        <confluent.version>6.1.15</confluent.version>
+        <jackson.version>2.16.0</jackson.version>
     </properties>
 
     <repositories>
         <repository>
             <id>confluent</id>
             <name>Confluent</name>
-            <url>https://packages.confluent.io/maven/</url>
+            <url>http://packages.confluent.io/maven/</url>
         </repository>
     </repositories>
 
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -52,6 +52,7 @@
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
         <commons.codec.version>1.15</commons.codec.version>
         <snakeyaml.version>1.28</snakeyaml.version>
+        <jackson.version>2.16.0</jackson.version>
```
ngupta0205, line 55: As discussed over the call, let's make this change in the parent pom only. That would facilitate others to not pin their individual versions to fix this CVE, rather just upgrade the parent pom version.

src/test/java/io/confluent/connect/elasticsearch/MappingTest.java
```diff
@@ -38,6 +38,7 @@
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+@SuppressWarnings("deprecation")
```
ngupta0205, line 41: why is this needed?

ngupta0205 (approved):
Left a minor comment. LGTM otherwise.

src/test/java/io/confluent/connect/elasticsearch/MappingTest.java
sp-gupta: ObjectNode is getting deprecated so the build fails with some warnings.

aniketshrimal (approved):
LGTM @sp-gupta! Lets create a ticket to remove this pinning once common version is released with the change.

ngupta0205 (approved):

PR #788, opened 2024-06-06
[CC-27487] Error Reporting in Elasticsearch via Errant Reporter

sp-gupta:
Changes: 

- Only exceptions from Elasticsearch will be logged in DLQ and all connector level errors will be logged as before. 

- When behavior.on.malformed.documents is set to:
IGNORE - Report errors to DLQ and the connector keep running and processes other records.
FAIL - Report errors to DLQ and the connector fails.

- Support for log.sensitive.data is removed. In case of any errors, the log points the customer to check DLQ topic for errors.

- Since this affect UX for on-prem customers, it will be released as a minor version change.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -696,6 +696,13 @@ public class ElasticsearchClient {
    * @param executionId the execution id of the request associated with the response
    */
   private synchronized void reportBadRecordAndError(BulkItemResponse response, long executionId) {
+
+    // RCCA-7507 : Don't push to DLQ if we receive Internal version conflict on data streams
+    if (response.getFailureMessage().contains(VERSION_CONFLICT_EXCEPTION)
+            && config.isDataStream()) {
+      log.debug("Skipping DLQ insertion for DataStream type.");
+      return;
+    }
     if (reporter != null) {
       List<SinkRecordAndOffset> sinkRecords =
           inFlightRequests.getOrDefault(executionId, new ArrayList<>());
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
@@ -104,8 +104,6 @@ public class ElasticsearchClient {
       )
   );
   private static final String UNKNOWN_VERSION_TAG = "Unknown";
-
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -715,20 +690,12 @@ private void removeFromInFlightRequests(long executionId) {
...
    */
-  private synchronized void reportBadRecord(BulkItemResponse response,
-                                            long executionId) {
-
-    // RCCA-7507 : Don't push to DLQ if we receive Internal version conflict on data streams
-    if (response.getFailureMessage().contains(VERSION_CONFLICT_EXCEPTION)
-            && config.isDataStream()) {
-      log.debug("Skipping DLQ insertion for DataStream type.");
-      return;
-    }
```
aniketshrimal, line 731: Did we check why this condition was added and if it is safe to remove this and report error for VERSION_CONFLICT_EXCEPTION as well now?

aniketshrimal (approved):
Overall LGTM @sp-gupta! Just have one minor comment - please address before merging!

aniketshrimal:
@sp-gupta Please update the description with more details and remove the confluent wiki link as this is an open-source connector and external folks won't have access to internal wikis.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
sp-gupta: I think this was missed while rebasing. Thanks for pointing this out. I have reverted this change.

aniketshrimal (approved):

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
```diff
@@ -663,17 +638,17 @@ private boolean handleMalformedDocResponse(BulkItemResponse response) {
         return false;
       case FAIL:
```
ngupta0205, line 639: nit. do we need `case FAIL:`?

ngupta0205 (approved):
LGTM.

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchClient.java
sp-gupta: Yes, it is not needed, but since this is part of old code, I didn't modify it.

PR #801, opened 2024-10-11
auth api key

alozano3:
## Problem
There is no support for ApiKey Header authentication which is used in ES cloud version

## Solution
Add a new property to put the API Key and use it to authenticate. It also checks if basic authentication is configured.

- [ ] yes
- [ ] no

##### Testing done:
- [ X] Unit tests
- [ X] Integration tests
- [ ] System tests
- [ X] Manual tests


giulio1979:
Hi all, anyone looking into this ? Do you need help ?

PR #842, opened 2025-02-26
Convert org.apache.kafka.connect.data.Date to epoch millis before sending to Elasticsearch

aleksanderlech:
## Problem

Lets consider the following kafka schema:

```json
{
      "name": "dateOfBirth",
      "type": [ "null", {"type": "int", "logicalType": "date"}],
      "default": null
}
```

And the target index field:

```json
  "dateOfBirth": {
      "type": "date"
    }
```

Currently the date type (org.apache.kafka.connect.data.Date) is being converted to int (epoch days) and ends up in Elasticsearch being interpreted as epoch millis turning it into some datetime long in the past.

Elasticsearch does not suport epoch days type for `date` fields.

## Solution

Assuming the field is properly indexed as `date` we should pass some of default supported formats like `epoch_millis`.

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [x] Integration tests
- [x] System tests
- [x ] Manual tests

## Release Plan

Bugfix, it might be have some incompatibles with existing user workarounds though.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/DataConverter.java
@@ -295,10 +295,11 @@ public class DataConverter {
       switch (schemaName) {
         case Decimal.LOGICAL_NAME:
           return copySchemaBasics(schema, SchemaBuilder.float64()).build();
-        case Date.LOGICAL_NAME:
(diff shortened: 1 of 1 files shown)


aleksanderlech:
Hello any chance of merging?

PR #859, opened 2025-04-30
Split test class to solve flakiness

sharanrudresh:
## Problem
The tests in this class were failing frequently, and it seemed to be due to test setup.
The test setup creates an Elasticsearch container, used by all the tests.
Except, one of the tests `testSsl`, so it re-creates the container with SSL enabled.
There seems to be some issue with this - the other tests sometimes end up using the container with SSL
Example run - https://semaphore.ci.confluent.io/jobs/084e28ec-fad2-43eb-a8a2-e40d0fcc0a05/summary?report_id=17990af8-cb17-371c-9a8e-215e0e201902&test_id=4a7aece7-5cab-3669-bf09-3c2e177c2c7a

## Solution
Move `testSsl` to a different class for isolation.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
+++ b/src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
@@ -18,9 +18,7 @@ package io.confluent.connect.elasticsearch;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.BATCH_SIZE_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.BEHAVIOR_ON_MALFORMED_DOCS_CONFIG;
 import static io.confluent.connect.elasticsearch.ElasticsearchSinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG;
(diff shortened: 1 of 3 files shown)


src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -804,73 +737,4 @@ public void testWriteDataStreamInjectTimestamp() throws Exception {
...
-        : name;
-  }
-
-  @Test
-  public void testConnectionUrlExtraSlash() {
-    props.put(CONNECTION_URL_CONFIG, container.getConnectionUrl() + "/");
-    config = new ElasticsearchSinkConnectorConfig(props);
-    ElasticsearchClient client = new ElasticsearchClient(config, null, () -> offsetTracker.updateOffsets());
-    client.close();
-  }
```
sp-gupta, line 820: Is this test getting added anywhere?

sharanrudresh: I must have removed it by accident. I'll add it back. Thanks for pointing it out!

sp-gupta (approved):

PR #869, opened 2025-06-30
CC-32762: Set threadName with "taskId" prefix

jjain1259:
## Problem
The Elasticsearch connector spawns additional threads (via bulkExecutorService) to offload bulk processing from the primary task thread. The overall resource consumption is underestimated as these additional threads are not being tracked. This can lead to inaccurate task placement decisions, causing some workers to become overloaded while others remain underutilized.

Jira Ticket: https://confluentinc.atlassian.net/browse/CC-32762

## Solution
Implement a thread naming convention that allows the runtime to accurately track CPU time across all threads associated with a specific task by setting the thread name with a `connectorName + taskId` prefix. The thread naming pattern follows: {connectorName}-{taskId}-elasticsearch-bulk-{threadNumber}

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [x] System tests
- [x] Manual tests

## Release Plan
This change is backward compatible and is targeted for both CP and CC

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -42,7 +42,7 @@ public class ElasticsearchSinkTask extends SinkTask {
 
   private DataConverter converter;
   private ElasticsearchClient client;
-  private ElasticsearchSinkConnectorConfig config;
+  private ElasticsearchSinkTaskConfig config;
   private ErrantRecordReporter reporter;
   private Set<String> existingMappings;
   private Set<String> indexCache;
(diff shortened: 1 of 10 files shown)


src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -80,7 +80,8 @@ protected void start(Map<String, String> props, ElasticsearchClient client) {
     }
     Runnable afterBulkCallback = () -> offsetTracker.updateOffsets();
     this.client = client != null ? client
-        : new ElasticsearchClient(config, reporter, afterBulkCallback);
+        : new ElasticsearchClient(config, reporter, afterBulkCallback,
+            config.taskNumber(), config.connectorName());
```
sp-gupta, line 84: nit: Rename to `taskId()`

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkConnectorTest.java
```diff
@@ -33,9 +33,14 @@ public void shouldCatchInvalidConfigs() {
   public void shouldGenerateValidTaskConfigs() {
```
sp-gupta, line 33: We can add some tests to verify that the thread generated has the expected name.

sp-gupta (approved):

PR #924, opened 2026-07-15
CC-42849: stop logging raw Kafka record key in version-conflict WARN/DEBUG logs

jjain1259:
## Summary
- Fixes [CC-42849](https://confluentinc.atlassian.net/browse/CC-42849): the version-conflict handler in `ElasticsearchClient.handleResponse` logged `response.getId()` at WARN and DEBUG, which for `key.ignore=false` is the raw Kafka record key (customer-controlled data such as user ids, emails, order ids) written to connector logs in cleartext.
- Removes the record key from the WARN (L588) and DEBUG (L615) log lines, keeping only safe fields (`opType`, `version`, `index`).
- Leaves the existing TRACE log line (L597) unchanged, since it retains the document-id detail for troubleshooting when explicitly enabled, per the ticket's recommendation.

## Test plan
- [x] `mvn compile` succeeds
- [ ] `mvn test` — `ElasticsearchClientTest` requires a Docker-backed Elasticsearch testcontainer, unavailable in the local dev environment; will run in CI
- [x] Confirmed no test asserts on the exact WARN/DEBUG log message text

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[CC-42849]: https://confluentinc.atlassian.net/browse/CC-42849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #925, opened 2026-07-15
CC-42849: redact remaining sensitive-data logging paths in the ES sink

jjain1259:
## Summary
Stacks on top of #924. A log-leak audit of this repo (comparable to the fix in #924) surfaced three more real-but-currently-unfired leak paths, closed here:

None of these were confirmed as active incidents — verified against 30 days of production logs (`logs.json.connect`) with no sensitive content observed for any of the three — but all three are genuine leak paths worth closing as defense in depth.

## Test plan
- [x] `mvn test -Dtest=DataConverterTest,ElasticsearchSinkTaskTest` — 80 tests pass
- [x] Confirmed no test asserts on the exact log/exception message text changed here


jjain1259:
Superseded by #926, which combines this and #924 into a single PR against a simply-named branch.

PR #926, opened 2026-07-15
CC-42849: stop logging sensitive data in the Elasticsearch sink

jjain1259:
## Summary
Combines #924 and #925 into one PR. A log-leak audit of this repo found several places where sensitive data (Kafka record keys, header values, and a connection URL) could reach connector logs:

- `ElasticsearchClient`: the version-conflict handler in `handleResponse` logged `response.getId()` at WARN and DEBUG, which for `key.ignore=false` is the raw Kafka record key (customer-controlled data such as user ids, emails, order ids). Removed the record key from the WARN (L588) and DEBUG (L615) log lines, keeping only safe fields (`opType`, `version`, `index`). The existing TRACE log line (L597) is left unchanged, since it retains the document-id detail for troubleshooting when explicitly enabled.

Only the first item (`ElasticsearchClient`) was a confirmed pattern already known to leak (per [CC-42849](https://confluentinc.atlassian.net/browse/CC-42849)); the other three were found via static analysis and verified against 30 days of production logs (`logs.json.connect`) with no sensitive content observed, but are real leak paths worth closing as defense in depth.

## Test plan
- [x] `mvn test -Dtest=DataConverterTest,ElasticsearchSinkTaskTest` — 80 tests pass
- [ ] `mvn test` — `ElasticsearchClientTest` requires a Docker-backed Elasticsearch testcontainer, unavailable in the local dev environment; will run in CI
- [x] Confirmed no test asserts on the exact log/exception message text changed here

[CC-42849]: https://confluentinc.atlassian.net/browse/CC-42849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
@@ -22,7 +22,10 @@ import java.security.PrivilegedExceptionAction;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
 import javax.security.auth.Subject;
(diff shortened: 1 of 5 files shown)


jjain1259:
Closing in favor of #927 — this fix should land on 14.1.x first and forward-merge through 15.0.x → 15.1.x → master per this repo's usual workflow, rather than being applied directly here. #927 now has all 3 commits from this PR (including the NumberFormatException cause-leak fix added after this PR was opened) and is diff-equivalent.

PR #927, opened 2026-07-16
CC-42849: stop logging sensitive data in the Elasticsearch sink (14.1.x)

jjain1259:
## Summary
Backport of #926 (originally targeted at 15.1.x, now closed in favor of this PR) to 14.1.x. A log-leak audit of this repo found several places where sensitive data (Kafka record keys, header values, and a connection URL) could reach connector logs:

- `ElasticsearchClient`: the version-conflict handler in `handleResponse` logged `response.getId()` at WARN and DEBUG, which for `key.ignore=false` is the raw Kafka record key (customer-controlled data such as user ids, emails, order ids). Removed the record key from the WARN and DEBUG log lines, keeping only safe fields (`opType`, `version`, `index`). The existing TRACE log line is left unchanged, since it retains the document-id detail for troubleshooting when explicitly enabled.
- `DataConverter` (follow-up): the above fix still chained the original `NumberFormatException` as the cause, and that exception's own message embeds the invalid header value. Since the outer exception propagates uncaught and Connect logs the full cause chain on task failure, the value was still leaking one level down. The cause is now dropped entirely.

This is a cherry-pick of the three commits from #926 (e06086af, d5602d72, 7626c5fe) onto 14.1.x, with no conflicts. Diffed against #926 to confirm equivalence (the only difference is a git blob-hash line, not the actual changes).

## Test plan
- [x] `mvn test -Dtest=DataConverterTest` passes on the equivalent 15.1.x branch (27/27, including a new regression test for the NumberFormatException fix) — surefire on this 14.1.x branch currently discovers 0 tests locally due to a pre-existing missing `junit-vintage-engine`/`junit-jupiter-engine` dependency in this branch's `pom.xml`, unrelated to this change; will run in CI
- [x] All 4 leak paths verified end-to-end against a real Elasticsearch Cloud cluster via kafka-docker-playground, using canary values to confirm they no longer appear in connector logs

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,7 @@
         <es.version>7.17.24</es.version>
         <lz4.version>1.10.1</lz4.version>
         <hamcrest.version>1.3</hamcrest.version>
-        <mockito.version>2.28.2</mockito.version>
+        <mockito.version>4.6.1</mockito.version>
         <gson.version>2.9.0</gson.version>
         <test.containers.version>1.16.3</test.containers.version>
         <kafka.connect.maven.plugin.version>0.11.1</kafka.connect.maven.plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -175,6 +175,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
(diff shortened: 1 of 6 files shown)


jjain1259:
Closing in favor of #928 — same branch/commits, pushed directly to this repo via git push-external instead of from a fork, which is the correct convention here.

PR #928, opened 2026-07-16
CC-42849: stop logging sensitive data in the Elasticsearch sink (14.1.x)

jjain1259:
## Summary
Backport of #926 (originally targeted at 15.1.x, closed in favor of this line of work) to 14.1.x. Supersedes #927, which was opened from a fork — this is the same branch pushed directly to this repo instead, per team convention of raising PRs directly rather than cross-fork.

A log-leak audit of this repo found several places where sensitive data (Kafka record keys, header values, and a connection URL) could reach connector logs:

- `ElasticsearchClient`: the version-conflict handler in `handleResponse` logged `response.getId()` at WARN and DEBUG, which for `key.ignore=false` is the raw Kafka record key (customer-controlled data such as user ids, emails, order ids). Removed the record key from the WARN and DEBUG log lines, keeping only safe fields (`opType`, `version`, `index`). The existing TRACE log line is left unchanged, since it retains the document-id detail for troubleshooting when explicitly enabled.
- `DataConverter` (follow-up): the above fix still chained the original `NumberFormatException` as the cause, and that exception's own message embeds the invalid header value. Since the outer exception propagates uncaught and Connect logs the full cause chain on task failure, the value was still leaking one level down. The cause is now dropped entirely.

Also fixed two pre-existing local test-infrastructure gaps found while verifying this branch, both invisible before because surefire was silently discovering 0 tests:
- Added `junit-vintage-engine` (test scope) — `junit-jupiter-api` being present made surefire select the JUnit Platform provider, but with no vintage engine to bridge it, every JUnit4-style test in the module silently ran 0 assertions.
- Bumped `mockito.version` from `2.28.2` to `4.6.1` (matching 15.1.x/master) — `2.28.2`'s bundled Byte Buddy can't instrument classes under JDK 17, which broke every test mocking a final/interface type once tests actually started running.

## Test plan
- [x] `mvn test` (excluding the Docker-dependent `ElasticsearchClientTest`) — 187/187 pass
- [x] `mvn package` succeeds
- [x] All 4 leak paths re-verified end-to-end against a real Elasticsearch Cloud cluster via kafka-docker-playground, on the final commit (`9f40d52a`), using canary values to confirm they no longer appear in connector logs — see Verification below

## Verification (re-run against final commit 9f40d52a)

Built `confluentinc-kafka-connect-elasticsearch-14.1.8-SNAPSHOT.zip` from this branch and ran it in `kafka-docker-playground` against a live Elastic Cloud cluster, exercising each fix with a distinct canary value that would show up in logs if a fix regressed. Full container logs were grepped for each canary after the run.

**1. `ElasticsearchClient` — raw record key in version-conflict log**
Wrote 3 docs keyed `CUSTOMER-KEY-CANARY-*`, then replayed the same Kafka offsets from a fresh consumer group to force a real ES version conflict:
```
DEBUG Ignoring EXTERNAL version conflict for operation INDEX version 0 in index 'cc42849-key-canary'. (io.confluent.connect.elasticsearch.ElasticsearchClient:614)
```
`grep -c "CUSTOMER-KEY-CANARY"` across the full container log → **0**.

**2. `ConfigCallbackHandler` — credential in `connection.url`**
Configured `connection.url` with an embedded `demo-user:URL-PASSWORD-CANARY-9945@...` credential pointing at an unreachable host:
```
INFO Using unsecured connection to [https://cc42849-unreachable-host.invalid:9243]. (io.confluent.connect.elasticsearch.ConfigCallbackHandler:118)
```
Credential fully stripped from the connector's own log line. (As noted separately: Kafka Connect's own generic config dump and the raw `UnknownHostException` message still echo the credential — pre-existing framework/JDK-level paths, out of scope for this connector's code.)

**3. `ElasticsearchSinkTask` — full exception logged on conversion failure**
Produced a record with no key while `key.ignore=false` (null document id):
```
ERROR [cc42849-null-key|task-0] Can't convert record from topic=cc42849-null-key partition=0 offset=0. Failure type: org.apache.kafka.connect.errors.DataException (io.confluent.connect.elasticsearch.ElasticsearchSinkTask:296)
```
Single line, no stack trace, no leaked exception message.

**4. `DataConverter` — raw header value in version-parse exception, including its cause**
Produced a record with header `my-version:customer-email-canary+42849@example.com` (non-numeric):
```
Caused by: org.apache.kafka.connect.errors.ConnectException: Error converting external version header 'my-version' to long for record from topic=cc42849-bad-version-header partition=0 offset=0
	at io.confluent.connect.elasticsearch.DataConverter.maybeAddExternalVersioning(DataConverter.java:276)
	...
```
Exception chain ends there — no nested `Caused by: NumberFormatException`, confirming the follow-up fix. `grep -c "customer-email-canary"` across the full container log → **0**.

Test connectors and environment torn down after verification; no state left running against the test cluster.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,7 @@
         <es.version>7.17.24</es.version>
         <lz4.version>1.10.1</lz4.version>
         <hamcrest.version>1.3</hamcrest.version>
-        <mockito.version>2.28.2</mockito.version>
+        <mockito.version>4.6.1</mockito.version>
         <gson.version>2.9.0</gson.version>
         <test.containers.version>1.16.3</test.containers.version>
         <kafka.connect.maven.plugin.version>0.11.1</kafka.connect.maven.plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -175,6 +175,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>connect-runtime</artifactId>
             <version>${kafka.version}</version>
--- a/src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
@@ -22,7 +22,10 @@ import java.security.PrivilegedExceptionAction;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
 import javax.security.auth.Subject;
@@ -106,16 +109,34 @@ public class ConfigCallbackHandler implements HttpClientConfigCallback {
     }
 
     if (config.isKerberosEnabled() && config.isSslEnabled()) {
-      log.info("Using Kerberos and SSL connection to {}.", config.connectionUrls());
+      log.info("Using Kerberos and SSL connection to {}.", redactedConnectionUrls());
     } else if (config.isKerberosEnabled()) {
-      log.info("Using Kerberos connection to {}.", config.connectionUrls());
+      log.info("Using Kerberos connection to {}.", redactedConnectionUrls());
     } else if (config.isSslEnabled()) {
-      log.info("Using SSL connection to {}.", config.connectionUrls());
+      log.info("Using SSL connection to {}.", redactedConnectionUrls());
     } else {
-      log.info("Using unsecured connection to {}.", config.connectionUrls());
+      log.info("Using unsecured connection to {}.", redactedConnectionUrls());
     }
 
     return builder;
+  }
+
+  /**
+   * Returns the configured connection URLs with any embedded {@code user:password@} credential
+   * stripped out, so they are safe to log. Credentials should normally be supplied via the
+   * dedicated {@code connection.username}/{@code connection.password} configs, but nothing
+   * prevents a URL from carrying them directly.
+   */
+  private Set<String> redactedConnectionUrls() {
+    return config.connectionUrls().stream()
+        .map(ConfigCallbackHandler::redactUserInfo)
+        .collect(Collectors.toCollection(LinkedHashSet::new));
+  }
+
+  private static String redactUserInfo(String url) {
+    int authorityStart = url.indexOf("://") >= 0 ? url.indexOf("://") + 3 : 0;
+    int atIndex = url.indexOf('@', authorityStart);
+    return atIndex < 0 ? url : url.substring(0, authorityStart) + url.substring(atIndex + 1);
   }
 
   /**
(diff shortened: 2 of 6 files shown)


src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
```diff
@@ -106,18 +109,36 @@ public HttpAsyncClientBuilder customizeHttpClient(HttpAsyncClientBuilder builder
...
+   * dedicated {@code connection.username}/{@code connection.password} configs, but nothing
+   * prevents a URL from carrying them directly.
+   */
+  private Set<String> redactedConnectionUrls() {
+    return config.connectionUrls().stream()
+        .map(ConfigCallbackHandler::redactUserInfo)
+        .collect(Collectors.toCollection(LinkedHashSet::new));
+  }
+
+  private static String redactUserInfo(String url) {
```
ubishnoi, line 136: `redactUserInfo()` strips the credential up to the first @ instead of the last one before the host, so a password containing @ is only partially redacted 
We can use something like

```
private static String redactUserInfo(String url) {
  URI uri = new URI(url);
  return new URI(uri.getScheme(), null, uri.getHost(), uri.getPort(),
                  uri.getPath(), uri.getQuery(), uri.getFragment()).toString();
}
```

pom.xml
```diff
@@ -174,6 +174,11 @@
             <artifactId>junit-jupiter-api</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
```
ubishnoi, line 181: Why do we need this. Can we not write new test case without it?

ubishnoi (commented):
Hey @jjain1259 
Thanks for the PR.
Added some comments.PTAL

pom.xml
jjain1259: It's a pre-existing gap in this branch's pom.xml. The whole test suite (all ~186 existing JUnit4-style tests) has junit-jupiter-api as a dependency with no matching engine, which makes surefire silently discover zero tests and report a false "0 tests / BUILD SUCCESS." Writing my new test differently (e.g. JUnit5-style) wouldn't fix that — the other 186 tests are JUnit4 and still wouldn't run. This dependency is the minimal fix to make the entire suite actually execute again, not scope creep tied to one test case.

PFA:
`14.1.x` runs only 41 tests - all are ITs, no UTs run on this branch:
image

`15.0.x` branch runs 246 tests - combining both UTs and ITs:
image

src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
jjain1259: Confirmed, this is a real bug — good catch. Reproduced it: for `https://user:p@ssw0rd@host:9243`, the old code found the *first* `@` and left `ssw0rd@host:9243` in the log, i.e. the tail of the password plus the host.

Your suggested `URI`-based fix has an issue of its own though: `java.net.URI`'s parser can't disambiguate multiple unencoded `@`s either — for that same input it returns `getHost() == null`/`getUserInfo() == null` (no exception at parse time), and reconstructing via the 7-arg `URI(...)` constructor with a `null` host then throws `URISyntaxException`. So it would turn a quiet leak into a thrown exception from a `log.info` call in `configureHttpClient`, right on the code path that matters most.

Went with bounding the search to the authority component (before any `/`, `?`, `#`) and taking the *last* `@` within it, since only the password — never the host — can legitimately contain `@`. No URI parsing involved, so no new exception path.

Fixed in 96ca64da..4fdf18cf on `jjain/log_sev1_ficx`, which is #930 (this branch supersedes #928) — see confluentinc/kafka-connect-elasticsearch@4fdf18cf, plus regression tests in `ConfigCallbackHandlerTest` covering `@` in the password, multiple `@`s, and an `@` in the path (which should NOT be stripped).

PR #939, opened 2026-08-10
Guard stop() against a task whose start() never completed

asehgal-confluent:
## Problem

Kafka Connect's framework can call `stop()` on a task whose `start()` never ran or never completed (e.g. a sibling task's startup failure aborts the whole connector before this task finishes starting). `ElasticsearchSinkTask.stop()` unconditionally dereferenced `client`, which is only assigned near the end of `start()`. Earlier statements in `start()` can throw first (`new ElasticsearchSinkTaskConfig(props)` can throw `ConfigException`; the topic-to-resource-map parsing block explicitly rethrows as `ConnectException`). If either throws, `client` stays null and a framework-invoked `stop()` throws an uncaught `NullPointerException` instead of shutting down cleanly.

## Fix

Added a null check for `client` before `client.close()`.

## Test

Added `testStopBeforeStartDoesNotThrow`, which constructs an `ElasticsearchSinkTask` directly (never started) and calls `.stop()`, asserting it doesn't throw.

## Verification note

`mvn compile test-compile` succeeds cleanly. Running the full test suite (`mvn test -Dtest=ElasticsearchSinkTaskTest`) currently fails in this environment for all 28 pre-existing tests (confirmed unrelated to this change by reverting and re-running), due to Mockito's inline mock agent being unable to instrument classes on this environment's JDK 25 with the pinned Mockito version. This is a pre-existing, unrelated environment limitation, not something introduced by this change. Compilation was verified; full test execution was not.

## Precedent

Same bug pattern and fix shape as confluentinc/kafka-connect-azure-blob-storage#531 and confluentinc/kafka-connect-aws-dynamodb#157.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -144,7 +144,11 @@ public class ElasticsearchSinkTask extends SinkTask {
   @Override
   public void stop() {
     log.debug("Stopping Elasticsearch client.");
-    client.close();
+    // The framework can call stop() on a task whose start() never ran or never completed
+    // (e.g. a sibling task's startup failure aborts this one first), leaving client unset.
+    if (client != null) {
+      client.close();
+    }
   }
 
   @Override
--- a/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
@@ -144,8 +144,6 @@ public class ElasticsearchSinkTask extends SinkTask {
   @Override
   public void stop() {
     log.debug("Stopping Elasticsearch client.");
-    // The framework can call stop() on a task whose start() never ran or never completed
-    // (e.g. a sibling task's startup failure aborts this one first), leaving client unset.
     if (client != null) {
       client.close();
     }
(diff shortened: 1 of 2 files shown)


asehgal-confluent:
recheck

src/main/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTask.java
```diff
@@ -144,7 +144,11 @@ public Map<TopicPartition, OffsetAndMetadata> preCommit(
   @Override
   public void stop() {
     log.debug("Stopping Elasticsearch client.");
-    client.close();
+    // The framework can call stop() on a task whose start() never ran or never completed
+    // (e.g. a sibling task's startup failure aborts this one first), leaving client unset.
```
jjain1259: Lets remove the comment. Not needed since its a fairly obvious change

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchSinkTaskTest.java
```diff
@@ -340,6 +340,17 @@ public void testStartAndStop() {
     task.stop();
   }
 
+  @Test
+  public void testStopBeforeStartDoesNotThrow() {
+    // The framework can call stop() on a task whose start() never ran or never completed
+    // (e.g. a sibling task's startup failure aborts this one first), leaving client unset.
+    // Uses a local task so the shared fixture is left untouched, and initialize() to match how
+    // Connect builds a task before start().
```
jjain1259: Lets remove comment from here as well. Test name is explicit on what the test does

jjain1259 (commented):
Thanks for PR. Added minor nits

jjain1259 (approved):

PR #940, opened 2026-08-24
CC-35256: Migrate from HLRC to co.elastic.clients:elasticsearch-java 8.19.19

### LINKED ISSUE #847: Add support for Elasticsearch 9.0.0

jjain1259:
## Problem

The connector runs on the deprecated `RestHighLevelClient` 7.17.24 plus the full `org.elasticsearch:elasticsearch` server artifact. HLRC is a dead artifact (never published past 7.17.29) and was removed in Elasticsearch 9, blocking ES 9.x support, and the HLRC/server jars carry the connector's outstanding client-side CVE backlog (CC-35256).

Fixes #847.

**Breaking change (16.0.x major):** minimum supported Elasticsearch server is now **8.0**. The Validator rejects older servers and points 7.x users to the 15.x connector line. ES 9.x is supported through the client's automatic REST API compatibility (`compatible-with=8`).

## Solution

Replace HLRC with `co.elastic.clients:elasticsearch-java:8.19.19` on top of the retained low-level `elasticsearch-rest-client` — so `ConfigCallbackHandler` (SSL, Kerberos SPNEGO, proxy, auth, timeouts, compression) is unchanged.

Every config key keeps its name, default, and meaning.

### Core client
- `BulkProcessor` → `BulkIngester<BulkOpContext>`: per-operation contexts aligned 1:1 with bulk response items replace the executionId/request-map bookkeeping.
- **Deadlock fix**: BulkIngester completes bulk futures on the rest client's I/O reactor threads, which can hold the HTTP pool lock while the task thread holds the ingester lock (JVM-detected ABBA deadlock during ITs). The subclassed async client (`RetryingElasticsearchAsyncClient`) hops every bulk response onto a connector-owned dispatcher pool.
- Retry layers are disjoint: **whole-request failures** (transport errors, whole-response 429s) are retried **in-slot** by `RetryingElasticsearchAsyncClient` with jittered exponential backoff from `retry.backoff.ms` up to `max.retries` — the bulk future stays incomplete across backoffs, so BulkIngester's in-flight slot is held and record order is preserved at `max.in.flight.requests=1`; on exhaustion the task fails with "Bulk request failed". **Item-level failures**, per-item 429s included, are terminal and flow to the DLQ/error latch exactly as on master (BulkProcessor's `noBackoff` never retried them; the ingester's `backoffPolicy` is deliberately omitted so its internal retry pool — and its tail-requeue reordering — never exists). **Synchronous admin calls** → existing `RetryUtil` (unchanged).
- **`max.in.flight.requests` now allows N concurrent requests, not N-1** — the old code passed `maxInFlightRequests - 1` to BulkProcessor (whose parameter counted requests beyond the one being built), so the config always allowed one fewer request than its name and docs promise (the NetworkIT carried a TODO calling this misleading and the fix breaking). 16.0 passes the configured value directly; upgraders gain one concurrent request at the same setting. `testConcurrentRequests` now pins 4-for-4.
- Documents pass through as `BinaryData` verbatim (no per-record re-parse); upserts emit `{"doc":X,"upsert":X}` textually.
- Error classification walks the structured `ErrorCause.causedBy()` chain; `resource_already_exists` detection additionally falls back to a reason-substring check. Same malformed-doc set, version-conflict handling, and RCCA-7507 data-stream DLQ guard.
- `close()` drains buffered records within `flush.timeout.ms` like master's `awaitClose` (sibling batches carrying good records still get their offsets marked), and now also surfaces a failure latched during that final drain instead of losing it.

### Shutdown & failure-path hardening
- Deterministic teardown order in `closeResources()`: the ingester's scheduler outlives both `bulkIngester.close()` and `transport.close()` (the ingester's close parks forever if its scheduler is killed first), and the transport closes before the connector-owned executors so no bulk completion can land on a dead dispatcher.
- `failAllPending()` completes every still-pending bulk future exceptionally at close, so a retry discarded mid-backoff can never leave the ingester's in-flight accounting stuck.
- A synchronous `Error` escaping the transport (e.g. OOM serializing a bulk body — the client converts only `Exception`s to failed futures) is contained at the retry seam and completes the bulk future, failing the task with the real cause instead of silently wedging the pipeline.
- The client constructor builds the transport before its executors and cleans up on failure, so a mid-constructor throw can't orphan I/O-reactor threads across task restarts.

### Edge-config hardening (previously-working configs that the new client would break)
- `linger.ms=0` and `bulk.size.bytes=0` are clamped with a one-time WARN (BulkIngester rejects/mishandles zero; both used to mean "flush immediately/per record").
- Ping 403s surface as `TransportException` in the new client; validation still passes for least-privilege sink principals.
- `hasMapping()` counts dynamic-only mappings (`dynamic`, `dynamic_templates`, `_meta`) so pre-created mappings aren't overwritten.
- `retry.backoff.ms=0` no longer breaks the retry path; a missing configured `external.version.header` throws a clear `ConnectException` instead of an NPE.

### Dependencies / build
- Removed: `elasticsearch-rest-high-level-client`, `org.elasticsearch:elasticsearch` (+ transitive Lucene/snakeyaml tail) — the CVE payload. Packaged zip verified free of them.
- Added: `elasticsearch-java` 8.19.19 (Apache-2.0), explicit `elasticsearch-rest-client` 8.19.19 pin. Jackson stays at the CVE-remediated 2.21.5.
- Testcontainers 1.16.3 → 1.21.4 (modern Docker compatibility); `-Werror` enabled (per the pom's own migration note); default test image `elasticsearch:8.19.19`.

## Test Strategy

- [x] Unit tests — 256/256 on JDK 17 at the current head (257 total, 1 pre-existing skip; includes container-backed client tests against real ES 8.19.19)
- [x] Integration tests — 43/43 (`mvn clean verify`; 3 pre-existing skips), incl. SSL, Kerberos, data streams, WireMock network/retry suites and the deadlock regression scenarios; Semaphore runs the full suite per push
- [x] ES 9.5.2 spot-check — full client suite and `ElasticsearchConnectorIT` (17/17) against a real 9.x server
- [x] Manual — kafka-docker-playground end-to-end runs during development, including outage/recovery and shutdown-ordering scenarios against a live ES

## Release Plan

Ships as the 16.0.x major on master; no backport (not compatible with ES 7.x servers). Version bump to 16.0.0-SNAPSHOT per the release process.

## Notes

- `licenses/` and `notices/` regeneration (`mvn package -P licenses-source`) is deliberately split into a follow-up PR.
- CVE-2024-52980 is server-side and is not remediated by a client change — customers must patch their ES server.
- Supersedes #938 and #920. Credit to @yeikel's #920 for independently identifying the BulkIngester deadlock class and the compound-retry hazard.


yeikel:
Thanks for the effort!

How does this compare to my PR and why couldn't we review and merge mine instead? Seems like double effort unnecessarily 

I am from mobile but I'll check later

src/test/java/io/confluent/connect/elasticsearch/helper/ElasticsearchContainer.java
```diff
@@ -443,7 +441,7 @@ public String getConnectionUrl(boolean useContainerIpAddress) {
     return String.format(
         "%s://%s:%d",
         protocol,
-        useContainerIpAddress ? getContainerIpAddress() : hostMachineIpAddress(),
+        useContainerIpAddress ? getHost() : hostMachineIpAddress(),
         getMappedPort(ELASTICSEARCH_DEFAULT_PORT)
```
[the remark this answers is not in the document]
jjain1259:
```
Checked against the Testcontainers 1.21.4 source (already had it extracted). Copilot's finding doesn't hold up — the two methods are not just similar, they're identical:

@Deprecated
default String getContainerIpAddress() {
    return getHost();
}

default String getHost() {
    return DockerClientFactory.instance().dockerHostIpAddress();
}

getContainerIpAddress()'s entire body is return getHost(); — it was never anything but a deprecated alias. Both javadocs say the same thing verbatim: "Get the [IP address/host] that this container may be reached on (may not be the local machine)." There is no version where getContainerIpAddress() returned a container-internal address and getHost() returns something different — they return the exact same value from the exact same underlying call (DockerClientFactory.instance().dockerHostIpAddress()), always have.
```

pom.xml
ubishnoi, line 1: Can we drop the parsson pin?

It's pinned to `1.0.5` for `CVE-2023-4043`, from when elasticsearch-java was `8.2.2` (that declared the vulnerable 1.0.0).

`8.19.19` now declares `parsson:1.1.9`, which already has the fix. But our pin is a direct dependency, so nearest-wins picks it and we ship the older one:

```
+- co.elastic.clients:elasticsearch-java:jar:8.19.19:compile
|  +- (org.eclipse.parsson:parsson:jar:1.1.9:compile - omitted for conflict with 1.0.5)
\- org.eclipse.parsson:parsson:jar:1.0.5:compile
```

So the new client runs on a JSON-P provider four minor versions older than the one it was built against — and parsson handles every request/response body via JacksonJsonpMapper.

Can we remove the pin and take the transitive `1.1.9`? 

Same question for the `httpclient` pin — `4.5.13` while `elasticsearch-rest-client` declares `4.5.14`. Not from this PR (`7.17.24` declared `4.5.14` too), and no CVE in the gap, but might be worth clearing while the dependency block is open.

src/main/java/io/confluent/connect/elasticsearch/Validator.java
```diff
@@ -458,36 +468,24 @@ private void validateSsl() {
     }
   }
 
-  private void validateVersion(RestHighLevelClient client) {
-    MainResponse response;
+  private void validateVersion(ElasticsearchClient client) {
```
ubishnoi, line 471: **No hard check on the server version at runtime**

The < 8.0.0 gate only exists in Validator.validateVersion(), and it silently returns when client.info() throws — a least-privilege sink principal getting a 403 there passes validation. At task startup, ElasticsearchClient.getServerVersion() just logs a warning and returns "Unknown"; nothing compares it against the minimum.

Since 8.0 is now a hard requirement, should we fail the task at startup when the version is readable and below it? The clear message already exists in the Validator — it just never gets reached on that path.

If not, could you confirm what a user actually sees when they point this at a 7.x cluster? Does the connector fail with something that points at the version, or does it surface as an opaque write error?

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
```diff
@@ -765,27 +803,274 @@ public void testThreadNamingWithConnectorNameAndTaskId() throws Exception {
...
+      props.put(BATCH_SIZE_CONFIG, "1");
+      props.put(LINGER_MS_CONFIG, "10");
+      props.put(FLUSH_TIMEOUT_MS_CONFIG, "1000");
+      config = new ElasticsearchSinkConnectorConfig(props);
+      converter = new DataConverter(config);
+      ElasticsearchClient client = new ElasticsearchClient(config, null, () -> offsetTracker.updateOffsets(), 1, "elasticsearch-sink");
+
+      writeRecord(sinkRecord(0), client);
+
+      ConnectException e = assertThrows(ConnectException.class, client::close);
```
ubishnoi, line 896: Could we also assert the helper threads did?

On this path `closeResources()` skips `bulkIngester.close()`, so a `bulk-ingester-*` thread can be parked inside `flush()` waiting for a free in-flight slot. `failAllPending()` looks like it should wake it (it decrements requestsInFlightCount and signals that condition), but nothing here verifies it — and a leaked thread wouldn't fail this test, because the threads are daemons and awaitTerminationWithin calls shutdownNow() and moves on without reporting the outcome. Since every BulkIngester wait is awaitUninterruptibly(), shutdownNow() can't reclaim a thread parked in one.

All the pools are named with the `<connector>-<task>-elasticsearch-` prefix, so this could be checked without touching field visibility:

```
Thread.sleep(6_000);   // past awaitTerminationWithin's 5s budget
List<String> leaked = Thread.getAllStackTraces().keySet().stream()
    .filter(Thread::isAlive)
    .map(Thread::getName)
    .filter(n -> n.startsWith("elasticsearch-sink-1-elasticsearch-"))
    .collect(Collectors.toList());
```
assertTrue("threads still alive after close: " + leaked, leaked.isEmpty());
Mainly I'd like it as a regression guard on the teardown ordering — if someone later moves `failAllPending()` after `bulkIngesterScheduler.shutdown()`, this is the test that should catch it, and today it wouldn't.

ubishnoi (commented):
@jjain1259 Thanks for the PR. Added some comments

ubishnoi (commented):
@jjain1259 Found a case where the task thread hangs forever after a timed-out close().PTAL

Setup: flush.synchronously=true, max.in.flight.requests=1.

1. Bulk X completes and frees the slot. Its afterBulk is still running on the scheduler thread and will latch error.
2. Before that latch lands, [A,B] is sent into the free slot and gets stuck — ES has stopped responding. Record C arrives, doesn't fill a batch, sits in the buffer behind it.
3. X's afterBulk finishes → error latched.
4. Record D arrives → index(D) → throwIfFailed() → close().
5. close() waits flush.timeout.ms, buffer can't drain (slot still held by [A,B]) → closeResources() skips bulkIngester.close(), closes the transport, shuts down bulkIngesterScheduler.
6. Transport close fails [A,B] → their afterBulk runs → numBufferedRecords 3 → 1.
7. C was never sent. Its afterBulk needs bulkIngesterScheduler.submit() — already shut down → rejected → numBufferedRecords stays 1 forever.
8. put() throws, but the framework still calls preCommit() (closeAllPartitions() is in a finally).
9. preCommit() → SyncOffsetTracker.offsets() → waitForInFlightRequests() → while (numBufferedRecords > 0) await() — nothing left to signal it.
10. Task thread parked at ElasticsearchClient.java:370. Never reaches FAILED, shows RUNNING, does nothing. One bulk-ingester-* thread also left alive.


Suggestion: after close(), waitForInFlightRequests() shouldn't wait for numBufferedRecords to reach zero — the client is already torn down, nothing will drain the count, and since error is latched those records won't be committed and will simply be redelivered on restart.

pom.xml
jjain1259: Thanks for pointing this out. Made the changes

src/main/java/io/confluent/connect/elasticsearch/Validator.java
jjain1259: Confirmed on both points, and combined they change things here:

- **7.x is fully EOL already.** Per Elastic's own EOL page (https://www.elastic.co/support/eol), 7.17.x hit End of Maintenance 15-Apr-2025 and End of Support 15-Jan-2026 — so today, a 7.x server is an unsupported release on Elastic's side regardless of anything this connector does.

- **This isn't new behavior — master has the identical bypass.** `Validator.validateVersion()` on master does the same silent `catch (IOException | ElasticsearchStatusException e) { return; }` around `RestHighLevelClient.info()`. A 403 there has always skipped the version check, pre-migration too; this PR ported that logic unchanged (down to the same code comment). 

- The runtime side is the same story: master's `getServerVersion()` also never enforced a minimum — its only consumer was `shouldSetCompatibilityToES8()`, an HLRC-7.17-specific compatibility-mode toggle that doesn't exist anymore since the new client speaks the 8.x wire protocol natively. So the version has never been a runtime invariant, on either client.

Given that, the design intent (unchanged by this PR) is: the validate-time check is the intended gate, it gives a clear and actionable message whenever `client.info()` succeeds, and that's considered sufficient — the connector doesn't otherwise need to know or care what ES version it's talking to once it's running, and we'd rather not fail an otherwise-healthy task over a diagnostic call that a least-privilege principal is expected to 403 on. Not planning to add a runtime hard-check as part of this migration; open to revisiting that policy separately if we think it's worth it, but it'd be an intentional new invariant, not a regression fix.

src/test/java/io/confluent/connect/elasticsearch/ElasticsearchClientTest.java
jjain1259: Added — the assertion passes as-is, confirming all bulk-executor threads terminate within the close-wait budget on this path (`1aca3351`).

jjain1259:
> Jainam Jain (@jjain1259) Found a case where the task thread hangs forever after a timed-out close().PTAL
> 
> Setup: flush.synchronously=true, max.in.flight.requests=1.
> 
> 1. Bulk X completes and frees the slot. Its afterBulk is still running on the scheduler thread and will latch error.
> 2. Before that latch lands, [A,B] is sent into the free slot and gets stuck — ES has stopped responding. Record C arrives, doesn't fill a batch, sits in the buffer behind it.
> 3. X's afterBulk finishes → error latched.
> 4. Record D arrives → index(D) → throwIfFailed() → close().
> 5. close() waits flush.timeout.ms, buffer can't drain (slot still held by [A,B]) → closeResources() skips bulkIngester.close(), closes the transport, shuts down bulkIngesterScheduler.
> 6. Transport close fails [A,B] → their afterBulk runs → numBufferedRecords 3 → 1.
> 7. C was never sent. Its afterBulk needs bulkIngesterScheduler.submit() — already shut down → rejected → numBufferedRecords stays 1 forever.
> 8. put() throws, but the framework still calls preCommit() (closeAllPartitions() is in a finally).
> 9. preCommit() → SyncOffsetTracker.offsets() → waitForInFlightRequests() → while (numBufferedRecords > 0) await() — nothing left to signal it.
> 10. Task thread parked at ElasticsearchClient.java:370. Never reaches FAILED, shows RUNNING, does nothing. One bulk-ingester-* thread also left alive.
> 
> Suggestion: after close(), waitForInFlightRequests() shouldn't wait for numBufferedRecords to reach zero — the client is already torn down, nothing will drain the count, and since error is latched those records won't be committed and will simply be redelivered on restart.

https://github.com/confluentinc/kafka-connect-elasticsearch/pull/940#pullrequestreview-5236901626 (task thread hangs forever after a timed-out `close()`):

Reproduced deterministically (`testPreCommitReturnsAfterCloseOrphansUnsentRecord`: A,B stuck against a blackhole holding the mif=1 slot, C buffered and never dispatched, latched error, `index(D)` → `close()`; the `flush()` + `waitForInFlightRequests()` sequence parked at `ElasticsearchClient.java:370` exactly as described). Also confirmed against `WorkerSinkTask` that `closeAllPartitions()` is a try-with-resources close action, so `preCommit()` really does run after `put()` throws.

Fixed as suggested in 3d967889: `waitForInFlightRequests()` now returns once the client is closed; the error is latched on that path so `SyncOffsetTracker.offsets()` commits nothing and the records are redelivered on restart. With the fix the repro also shows no surviving pool threads.

For the record, master wedges in the same scenario one call earlier — `BulkProcessor.awaitClose()` flushes pending ops through an untimed `semaphore.acquire()` before its timed wait — so this is a pre-existing class of failure, not a migration regression.

jjain1259:
> Yeikel Santana (@yeikel) Thanks for the effort!
>
> How does this compare to my PR and why couldn't we review and merge mine instead? Seems like double effort unnecessarily
>
> I am from mobile but I'll check later

Thanks @yeikel for the head start on #920, and sorry for the slow reply here.

The short answer on why this landed as a separate PR: the migration turned out to be considerably more involved than a client swap, and the work needed a tight loop between design review, code changes, and testing that was easier to run on a branch we controlled end to end.

src/main/java/io/confluent/connect/elasticsearch/Validator.java
ubishnoi: Fine with no hard check — but a customer on ES 7.x whose user can't call info() will never learn why it broke: validation passes, then every bulk fails with an unrelated 400 after all the retries. Can we at least log the "ES 7.x is not supported, use 15.x" message as a WARN when the version is readable?

ubishnoi (approved):
Thanks for addressing the comments. Overall LGTM.
PTAL for this [comment](https://github.com/confluentinc/kafka-connect-elasticsearch/pull/940#discussion_r4069473498) once.

ubishnoi (approved):

PR #941, opened 2026-08-25
fix: redact connection.url credentials and record content from logs and exceptions

sumit-confluent:
## What & why

Remediation of the sensitive-log audit findings for **v15.0.5** (`346dd7c`), the shipped Elasticsearch sink (private wrapper `kafka-connect-elasticsearch-private` v0.104.0). All four findings are **internal** severity — three `connection.url` credential-echo paths and one record-conversion exception path. Only the report's flagged findings are changed; Cleared lines are untouched.

Report: https://drive.google.com/file/d/1Ab4J8wiBpCzy6lR2lOdFxhFcsfZAhQGr/view?usp=sharing

## Findings fixed

| Finding | file:line | Fix | Test |
|---|---|---|---|
| **F1** — raw `connection.url` echoed to forwarded ERROR log + config-validate response | `Validator.java:177` (source at `:642`) | Route the `HttpHost.create` call sites through a new `ConfigCallbackHandler.createRedactedHttpHost()` that strips `user:password@`; the `:177` log/`addErrorMessage` are left untouched and simply receive the redacted message | `ConfigCallbackHandlerTest.createRedactedHttpHost*` |
| **F2** — `connection.url` echoed into `ConfigException` message + `value` (also via config-validate REST) | `ElasticsearchSinkConnectorConfig.java:1377` | Redact each URL via `redactUserInfo` before the exception message and pass the redacted list as the `value` argument | `ElasticsearchSinkConnectorConfigTest.shouldRedactCredentialFromInvalidUrlErrorMessage` |
| **F3** — raw record-conversion `DataException` rethrown to the framework | `ElasticsearchSinkTask.java:300` | Rethrow a sanitized `DataException` (Kafka coordinates + failure type, **no cause**); the full exception still reaches the DLQ via `reportBadRecord` | `ElasticsearchSinkTaskTest.putRethrowsSanitizedConversionExceptionWithoutRecordContent` |
| **F4** — `connection.url` printed verbatim by inherited `AbstractConfig.logAll()` | `ElasticsearchSinkConnectorConfig.java:965`/`970` (helper at `:987`) | Strip any embedded `user:password@` from `connection.url` at the source, **before `super()`**, via `sanitizeConnectionUrl()`. The framework's own `logAll()` is left **enabled** — with credentials gone from the parsed config, the dump (and every other `connection.url` consumer) is safe without suppressing it. Auth is unaffected: credentials come from `connection.username`/`connection.password`, never URL user-info (an `HttpHost` carries none). `ElasticsearchSinkTaskConfig` inherits the protected constructor, so it is covered by the same change | `ElasticsearchSinkConnectorConfigTest.shouldRedactConnectionUrlCredentialsFromConfigLog` |

Each test was proven to **fail on the pre-fix code and pass after the fix**. (F1's helper test is green-only: it shares F2's root cause and the helper does not exist pre-fix.)

## Behavior changes (deliberate)
- Malformed-URL messages change prefix (`Invalid HTTP host:` → `Invalid Elasticsearch connection URL:`) and redact embedded credentials — this also changes the Connect config-validate REST response (F1, F2).
- The task-status trace for a conversion failure becomes a fixed `Can't convert <topic-partition-offset>. Failure type: <class>` with no cause (F3). The full exception is still delivered to the DLQ.
- Any embedded `user:password@` in `connection.url` is stripped from the parsed config at the source (F4). It was never used for authentication (the client authenticates via `connection.username`/`connection.password`), so this is a no-op for a correctly configured connector. The framework's `AbstractConfig.logAll()` still runs unchanged — same logger, format and `[hidden]` password masking — and simply no longer contains URL credentials.

## Out of scope
- **DLQ failure-detail stripping** (`ElasticsearchClient` `ReportingException`): the audit classifies the DLQ path as out of scope (trust boundary), so it is not changed here.
- **Log-redaction rules** (handled by the separate add-log-redaction workflow).

## ⚠️ Coordinate with PR #933
The maintainer's open PR #933 (`jjain/log-leak-audit-followup`, base `14.1.x`) already implements the F1/F2 connection-URL redaction — this PR mirrors that approach on `15.0.x`. When #933 forward-merges `14.1.x → 15.0.x` it may conflict with this PR; please coordinate merge order.

## Test
`ConfigCallbackHandlerTest`, `ElasticsearchSinkConnectorConfigTest`, `ElasticsearchSinkTaskTest`, `ValidatorTest` all green on JDK 8; checkstyle clean.

Audited commit: `346dd7ccc43b0706bbaeca1d455d7fdd9f6fbabf`.

## Follow-up: `redactUserInfo` still leaked for non-standard authority markers

Independent review + live repro against a running Connect worker found that `redactUserInfo` (the shared primitive behind F1, F2, and F4 above) only located the authority component via `indexOf("://")`, falling back to offset 0 for anything else — but still trusted a bare `/` as a path boundary even in that fallback case. That let the credential slip through unredacted for:

| `connection.url` shape | Example | Why it leaked |
|---|---|---|
| Scheme-relative (bare `//`) | `//user:pass@badhost:9243` | The URL's own leading `/` was mistaken for a path separator, collapsing the search window before it reached the `@` |
| Malformed single-slash scheme | `https:/user:pass@badhost:9243` | Same root cause — the single `/` after the scheme colon isn't a valid `://` marker, but was still treated as a path boundary |

Both were confirmed live: `HttpHost.create()` accepts both forms without throwing (treating the whole string as the hostname), so the subsequent `UnknownHostException` — surfaced via the same config-validate path as F1/F2 — echoed the raw password.

**Fix:** `redactUserInfo` now only trusts `/` as an authority-boundary when a real RFC 3986 authority marker (`scheme://` or bare `//`) is actually recognized via a new `AUTHORITY_PREFIX` pattern; otherwise it searches the rest of the string for the credential rather than risk a false negative.

**Test:** `redactUserInfoStripsCredentialsFromSchemeRelativeUrl`, `redactUserInfoStripsCredentialsFromMalformedSingleSlashScheme`, and `createRedactedHttpHostRedactsCredentialWhenUnderlyingExceptionEchoesIt` (the existing `createRedactedHttpHostRedactsCredentialOnParseFailure` used a space-in-host input whose failure message never echoed the credential in the first place, so it didn't actually exercise this class of bug).

Verified against a real build of this branch (`15.0.7-SNAPSHOT`) deployed into a Connect worker in kafka-docker-playground: both inputs above leaked the password in the `/config/validate` response before the fix and no longer do after it. Full existing suite still green aside from two pre-existing Testcontainers tests that can't pull `elasticsearch:8.15.2` in this environment (unrelated).

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
@@ -151,6 +151,20 @@ public class ConfigCallbackHandler implements HttpClientConfigCallback {
   }
 
   /**
+   * Parses {@code url} into an {@link HttpHost}, redacting any embedded {@code user:password@}
+   * credential from the thrown exception's message if parsing fails. {@link HttpHost#create}
+   * otherwise echoes the raw url verbatim in a malformed-url {@link IllegalArgumentException}.
+   */
+  static HttpHost createRedactedHttpHost(String url) {
+    try {
+      return HttpHost.create(url);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException(
+          "Invalid Elasticsearch connection URL: " + redactUserInfo(url));
+    }
+  }
+
+  /**
    * Configures HTTP authentication and proxy authentication according to the client configuration.
    *
    * @param builder the HttpAsyncClientBuilder
@@ -159,7 +173,7 @@ public class ConfigCallbackHandler implements HttpClientConfigCallback {
     CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
     if (config.isAuthenticatedConnection()) {
       config.connectionUrls().forEach(url -> credentialsProvider.setCredentials(
-              new AuthScope(HttpHost.create(url)),
+              new AuthScope(createRedactedHttpHost(url)),
               new UsernamePasswordCredentials(config.username(), config.password().value())
           )
       );
--- a/src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
@@ -135,10 +135,6 @@ public class ConfigCallbackHandler implements HttpClientConfigCallback {
         .collect(Collectors.toCollection(LinkedHashSet::new));
   }
 
-  // Matches the only two ways RFC 3986 introduces an authority component: a real
-  // "scheme://" prefix, or a bare "//" network-path reference. Anything else - including a
-  // malformed single-slash scheme like "https:/host" - does NOT reliably mark where a path
-  // begins, so it must not be trusted as a delimiter (see below).
   private static final Pattern AUTHORITY_PREFIX =
       Pattern.compile("^[A-Za-z][A-Za-z0-9+.-]*://|^//");
 
@@ -147,13 +143,6 @@ public class ConfigCallbackHandler implements HttpClientConfigCallback {
     boolean recognizedPrefix = schemeMatcher.find();
     int authorityStart = recognizedPrefix ? schemeMatcher.end() : 0;
 
-    // '?' and '#' always mark the end of the authority, recognized prefix or not. '/' only
-    // does when we've confirmed a real authority marker above - otherwise a malformed scheme
-    // (or no scheme at all) can put a '/' before the credential that isn't a path separator,
-    // which previously let the credential slip through unredacted (e.g. "//user:pass@host",
-    // "https:/user:pass@host", "user:pass@host" with no scheme at all). When we can't trust
-    // '/' as a boundary, it's safer to search the rest of the string than to risk a false
-    // negative.
     char[] delimiters = recognizedPrefix ? new char[] {'/', '?', '#'} : new char[] {'?', '#'};
     int authorityEnd = url.length();
     for (char delimiter : delimiters) {
--- a/src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
@@ -25,6 +25,8 @@ import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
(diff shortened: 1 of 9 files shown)


src/main/java/io/confluent/connect/elasticsearch/ConfigCallbackHandler.java
```diff
@@ -150,6 +150,20 @@ static String redactUserInfo(String url) {
         : url.substring(0, authorityStart) + url.substring(atIndex + 1);
   }
 
+  /**
+   * Parses {@code url} into an {@link HttpHost}, redacting any embedded {@code user:password@}
+   * credential from the thrown exception's message if parsing fails. {@link HttpHost#create}
+   * otherwise echoes the raw url verbatim in a malformed-url {@link IllegalArgumentException}.
+   */
+  static HttpHost createRedactedHttpHost(String url) {
```
sumit-confluent, line 158: @jjain1259 please check if we are good to redact url using this approach

jjain1259:
## Changes on top of the original fix (review follow-up)

Independent review + live repro found one real gap and a couple of test/code-quality items. Summary of the 6 commits added on top of `64c55260`:

**Functional fixes:**
- **`c9d1e4b7`** — `redactUserInfo` only located the authority component via `indexOf("://")`, falling back to offset 0 otherwise, but still trusted a bare `/` as a path boundary even in that fallback. That let the credential through unredacted for a scheme-relative URL (`//user:pass@host`) and a malformed single-slash scheme (`https:/user:pass@host`) — both reproduced live against a running Connect worker before the fix, both closed after it. New tests: `redactUserInfoStripsCredentialsFromSchemeRelativeUrl`, `redactUserInfoStripsCredentialsFromMalformedSingleSlashScheme`, `createRedactedHttpHostRedactsCredentialWhenUnderlyingExceptionEchoesIt` (the existing `createRedactedHttpHostRedactsCredentialOnParseFailure` used a space-in-host input whose failure message never echoed the credential in the first place, so it didn't actually exercise this class of bug).
- **`43c12928`** — added a `log.trace(..., convertException)` right before the sanitized `DataException` rethrow in `ElasticsearchSinkTask`, so the real converter failure is still recoverable for debugging (opt-in, off by default) even though the framework only ever sees the sanitized message.

**Cleanup (`9b393eaa`, `5d6686f9`, `a878f3eb`, `6b1a5c8a`):** trimmed several inline comments down or removed them where the code + this PR description already carry the explanation, and shortened `sanitizeConnectionUrl`'s javadoc.

All existing tests plus the new ones green; verified against a real build of this branch deployed into a Connect worker in kafka-docker-playground (both the original leak and the two new gaps confirmed closed).

jjain1259 (approved):

lakshayjoshi28 (approved):

PR #942, opened 2026-09-08
CC-35256: Coalesce bulk request bodies to fix ES connector throughput regression

jjain1259:
Stacked on #940 (base: `jjain/es-client-8.19-migration`). Tracks the performance work for the Java API client migration separately from the migration itself.

## Problem

A/B on a 3-core devel worker against the same Elastic Cloud deployment, same topic, same starting lag:

| | HLRC (master) | Java API client (#940) |
|---|---|---|
| Record rate | ~39-44K rec/s | ~20-22K rec/s, CPU-pinned |
| Container CPU | ~0.8-1.2 cores | 2.6-2.9 cores (87-98% of request) |
| I/O dispatcher threads | ~2.2 µs/record | ~115 µs/record (2.4 live cores) |
| Task thread | 15.4 µs/record | 16.6 µs/record (parity) |

async-profiler put 85% of samples on the four Apache HttpAsyncClient `I/O dispatcher` threads, 87% of those under the write path (`SSLIOSession.outboundTransport`, `SSLEngine`, `epoll`, syscall self time) — 75% of all samples process-wide.

**Root cause** (elasticsearch-java 8.19.19, unchanged on upstream `main` and in Rest5Client): `ElasticsearchTransportBase.collectNdJsonLines` emits one `ByteBuffer` per NDJSON line plus a shared 1-byte separator (four buffers per index op). `RestClientHttpClient` wraps that in a chunked `MultiBufferEntity` that writes exactly one buffer per `produceContent` call, and httpcore-nio makes one call per writable event. Every buffer therefore becomes its own HTTP chunk, TLS record and `write()` syscall: ~7,500 per 166 KB bulk versus ~20 for the HLRC's single `NByteArrayEntity`.

**Upstream:** reported as https://github.com/elastic/elasticsearch-java/issues/1339 (2026-09-09).

## Change

`CoalescingHttpClient` decorates the `TransportHttpClient` seam between the typed transport and `RestClientHttpClient`, merging a multi-buffer bulk body into a single buffer before it reaches the low-level client (bodies of 0 or 1 buffer pass through untouched, so only bulk pays the one extra copy). The chunk encoder then fills its session buffer per event and the reactor cost drops sharply — from ~115 µs/record to ~4 µs/record — though it remains somewhat higher than the HLRC's own ~2.2 µs/record; the fix targets the outbound write-storm the HLRC never had, not an exact reproduction of its cost profile. Total connector CPU per record still comes out at or below the HLRC's, since the HLRC pays its own separate cost decoding bulk responses on a dedicated thread pool this client doesn't need.

`ElasticsearchClient` builds its transport through `CoalescingHttpClient.transport(...)` instead of `RestClientTransport`. Bytes on the wire are identical, only framed into fewer chunks. No config changes.

An earlier version of this decorator also re-completed the response future on the connector's dispatcher pool, to keep response decoding off the I/O reactor threads. That was dropped: the regression was entirely on the outbound write path (flamegraph put response decoding at 1.5% of samples), `RetryingElasticsearchAsyncClient` already hops bulk completions onto the dispatcher pool one layer up, and the extra hop only added complexity (a wrapped future for cancellation, a rejected-hop path, `closeQuietly`) without addressing anything measured here.

## Verification

- `mvn validate` (checkstyle) clean.
- JDK 17: `RetryingElasticsearchAsyncClientTest`, `ValidatorTest`, `DataConverterTest`, `ConfigCallbackHandlerTest` (101 tests) and the container-backed `ElasticsearchClientTest` (38 tests) all pass on the coalescing-only decorator; after dropping the inbound hop, `mvn validate` and `RetryingElasticsearchAsyncClientTest` (10/10) were rerun clean, and the container suite is unaffected since it never exercised the removed path.
- Full-backlog-drain rerun (devel, 2026-09-13/14, `lcc-devcq2z12v7`, coalescing-only decorator — verified via bytecode to have no inbound hop): **all acceptance targets met** over a complete drain of the same ~124.5M-record backlog (not a short window). I/O dispatcher: 0.14–0.20 live cores mid-run (down from ~2.5 pre-fix), whole-run bookend 4.24 µs/record; container CPU 0.79–0.96 cores at 39.0–42.4K rec/s (up from 20.0–21.5K pre-fix, matching the HLRC's own 39.3–43.5K); task-thread µs/record unchanged (15.17 vs 16.6 pre-fix vs 15.43 HLRC); ~19.41 µs/record total vs ~132.1 before and ~22.90 for the HLRC. Drain time closed from ~101–102 min (pre-fix) to ~51–52 min, matching the HLRC's own ~50 min. Full tables, per-pool split and flamegraph summary in the verification comment below.

No unit tests added for the decorator by request; the perf rerun above and the existing suites are the verification.

### DIFF
--- a/src/main/java/io/confluent/connect/elasticsearch/CoalescingHttpClient.java
+++ b/src/main/java/io/confluent/connect/elasticsearch/CoalescingHttpClient.java
@@ -1,0 +1,198 @@
+/*
+ * Copyright 2020 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.connect.elasticsearch;
+
+import co.elastic.clients.json.JsonpMapper;
+import co.elastic.clients.transport.ElasticsearchTransport;
+import co.elastic.clients.transport.ElasticsearchTransportBase;
+import co.elastic.clients.transport.TransportOptions;
+import co.elastic.clients.transport.http.TransportHttpClient;
+import co.elastic.clients.transport.rest_client.RestClientHttpClient;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import org.elasticsearch.client.RestClient;
+
+/**
+ * Decorates the Java API client's HTTP layer at the {@link TransportHttpClient} seam, the
+ * boundary between the typed transport ({@link ElasticsearchTransportBase}: JSON encode and
+ * decode) and the byte-level {@link RestClientHttpClient}. It does two things there.
+ *
+ * <p><b>Outbound: coalesce the request body into one buffer.</b> The transport hands a
+ * bulk request down as one {@code ByteBuffer} per NDJSON line plus a shared one-byte
+ * separator, four buffers per index operation. {@code RestClientHttpClient} wraps that
+ * iterable in a chunked {@code MultiBufferEntity} that writes exactly one buffer per
+ * {@code produceContent} call, and httpcore-nio makes one such call per writable event.
+ * Each buffer therefore becomes its own HTTP chunk, TLS record and {@code write()} syscall:
+ * thousands per bulk instead of the ~20 the High Level REST Client produced from a single
+ * byte-array entity. Measured on a 3-core worker this pinned the four I/O reactor threads
+ * at ~2.5 cores and capped throughput at half of the old client's. With one merged buffer the
+ * chunk encoder fills its session buffer per event and the reactor cost returns to parity.
+ * The copy costs one extra pass over the body on the calling thread. Bodies of zero or one
+ * buffer (every non-bulk request) pass through untouched.
+ *
+ * <p><b>Inbound: complete the response future on the connector's dispatcher pool.</b> The
+ * delegate completes its future on an I/O reactor thread, and the transport's own
+ * continuation decodes the JSON response right there, before any connector code runs.
+ * Re-completing on {@code dispatcher} moves that decode, and everything downstream of it
+ * (listener callbacks, retry scheduling, offset bookkeeping), off the reactor for every
+ * async endpoint. This is the deadlock guard described at the connector's dispatcher pool:
+ * connector callbacks must never run on a thread that the HTTP client needs to make
+ * progress, and must not share a pool with the ingester's flush scheduler either
+ * ({@link RetryingElasticsearchAsyncClient} keeps its own hop as a second line of defence).
+ * Success and failure both hop; a dispatcher that rejects the hop (shutdown race) fails the
+ * future rather than leaving it incomplete, and cancelling the returned future cancels the
+ * in-flight HTTP request as the {@link TransportHttpClient} contract requires.
+ */
+final class CoalescingHttpClient implements TransportHttpClient {
+
+  private final TransportHttpClient delegate;
+  private final Executor dispatcher;
+
+  CoalescingHttpClient(TransportHttpClient delegate, Executor dispatcher) {
+    this.delegate = delegate;
+    this.dispatcher = dispatcher;
+  }
+
+  /**
+   * Builds the connector's transport: the stock REST client HTTP layer wrapped by this
+   * decorator, beneath the stock typed transport. Closing the transport closes the
+   * {@code RestClient} beneath it, as with {@code RestClientTransport}.
+   */
+  static ElasticsearchTransport transport(
+      RestClient restClient,
+      Executor dispatcher,
+      JsonpMapper mapper
+  ) {
+    return new Transport(
+        new CoalescingHttpClient(new RestClientHttpClient(restClient), dispatcher), mapper);
+  }
+
+  /** Named rather than anonymous so stack traces and thread dumps identify it. */
+  static final class Transport extends ElasticsearchTransportBase {
+    Transport(TransportHttpClient httpClient, JsonpMapper mapper) {
+      super(httpClient, null, mapper);
+    }
+  }
+
+  // Must forward: RestClientHttpClient's options carry the client's default headers.
+  @Override
+  public TransportOptions createOptions(TransportOptions options) {
+    return delegate.createOptions(options);
+  }
+
+  @Override
+  public Response performRequest(
+      String endpointId,
+      Node node,
+      Request request,
+      TransportOptions options
+  ) throws IOException {
+    return delegate.performRequest(endpointId, node, coalesce(request), options);
+  }
+
+  @Override
+  public CompletableFuture<Response> performRequestAsync(
+      String endpointId,
+      Node node,
+      Request request,
+      TransportOptions options
+  ) {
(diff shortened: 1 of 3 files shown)


jjain1259:
## Performance verification (updated: full-backlog-drain rerun, 2026-09-13/14)

**Superseded verification note:** the run originally documented here (2026-09-08, `lcc-devcq2pg3mm`) was against an earlier build of this decorator that still included the inbound response-completion hop later dropped from this PR — that's why the original numbers below showed the `bulk-dispatcher` pool carrying JSON response decode ("what the inbound hop is for"). That behavior does not exist in what's actually merged. This comment has been rewritten against the actual coalescing-only build, verified over a complete drain of the same fixed backlog rather than a short sampled window.

Build from this branch (`fb967265`, coalescing-only — verified via bytecode: 2-param `CoalescingHttpClient.transport(RestClient, JsonpMapper)`, no `Executor` param, no `closeQuietly`/`whenCompleteAsync` present, confirming the inbound hop is absent) deployed as `lcc-devcq2z12v7` on the same PCC, same Elastic Cloud 8.19.9 deployment, same topic and starting backlog (~124.5M records across 6 partitions) as the HLRC and pre-fix runs — each drained to completion rather than sampled over a short window.

### Acceptance table (full-drain, whole-run bookend figures)

| Metric | Pre-fix (#940, before this PR) | Target | This PR | HLRC (master) |
|---|---|---|---|---|
| Total drain time (full ~124.5M-record backlog) | ~101–102 min | ≈ HLRC | **~51–52 min** | ~50 min |
| I/O dispatcher live cores (4 reactor threads), mid-run | 2.38–2.44 | < 0.25 | **0.14–0.20** | ~0.09–0.10 |
| I/O dispatcher whole-run µs/record | 115.5 | — | **4.24** | 2.18 |
| Connect container CPU | 2.60–2.94, pinned (87–98%) | ≈ old client | **0.79–0.96** (26–32%) | 0.82–1.17 (27–39%) |
| Record rate | 20.0–21.5K rec/s, CPU-bound | rises to match HLRC | **39.0–42.4K rec/s** | 39.3–43.5K rec/s |
| Task-thread µs/record | 16.6 | unchanged | **15.17** | 15.43 |
| Total connector CPU µs/record (task + reactor) | 132.1 | | **19.41** | 22.90 (+5.29 on a bulk-response-decode pool the new client doesn't need) |

### Per-pool CPU split (whole-run bookend: final jstack cumulative CPU ÷ 124,502,000 records)

| Pool | Threads | This PR | Pre-fix (before) | HLRC |
|---|---|---|---|---|
| task-thread | 1 | 15.17 | 16.6 | 15.43 |
| `I/O dispatcher` | 4 | 4.24 | 115.5 | 2.18 |
| `bulk-ingester` scheduler | 5 | 0.27 | 0.32 | – |
| `bulk-dispatcher` | 4 (mif=4) | 0.05 | 0.065 | (bulk-executor 5.29, HLRC-only pool) |
| `bulk-retry` | – | never started | never started | – |
| GC | 8 | 0.83 | 1.40 | 0.89 |

Note: unlike the earlier inbound-hop build, `bulk-dispatcher` here is just `BulkIngester`'s own idle scheduling pool — it does not carry response decoding. That hop was dropped from this PR; response decoding happens inline in the async client's own callback, not on a connector-owned pool.

### Flamegraph (async-profiler itimer, 60s, programmatically parsed against the raw HTML output — not eyeballed)

- task-thread: 74.70% of samples — generic Kafka Connect/Avro/Jackson record-processing machinery, the same composition as the HLRC and pre-fix runs, confirming this cost is throughput-driven, not client-specific.
- 4 reactor threads combined: 18.86–19.05% of samples (down from 85.65% pre-fix), split between `SSLIOSession.readPlain` (17–18%) and residual `__syscall_cancel_arch_end` write cost (23–25% of the reactor's own slice) — a mix of read and residual write cost, not dominated by either.
- The write-syscall path that dominated every pre-fix profile is down to ~4.5–5% of all samples process-wide.

### Root cause recap (unchanged)

`ElasticsearchTransportBase.collectNdJsonLines` emits one `ByteBuffer` per NDJSON line plus a shared 1-byte separator (four per index op); `RestClientHttpClient` wraps them in a chunked `MultiBufferEntity` that writes one buffer per `produceContent` call, and httpcore-nio makes one call per writable event. Coalescing the body at the `TransportHttpClient` seam restores single-entity framing on the outbound path. Filed upstream as elastic/elasticsearch-java#1339.

Raw artifacts (30 jstacks, 10 flamegraphs, Prometheus cross-checks, full analysis writeup) available on request.

src/main/java/io/confluent/connect/elasticsearch/CoalescingHttpClient.java
```diff
@@ -0,0 +1,122 @@
...
+    for (ByteBuffer buffer : body) {
+      buffers.add(buffer);
+      size += buffer.remaining();
+    }
+    if (buffers.size() <= 1) {
+      return request;
+    }
+    ByteBuffer merged = ByteBuffer.allocate(size);
+    for (ByteBuffer buffer : buffers) {
+      merged.put(buffer.duplicate());
```
ubishnoi, line 112: Nit: For the duplicate() here — could we add a UT so a future maintainer doesn't drop it? NdJsonSeparator is a single shared buffer, so without it the second bulk through the transport would lose all its newlines.

ubishnoi (commented):
@jjain1259 Thanks for the PR — I added one minor comment.
One general question: this introduces an additional transient copy of each bulk body via ByteBuffer.allocate(size). At worst, that could add heap usage of roughly bulk.size.bytes × max.in.flight.requests, with a new allocation for each attempt. Was this tested with larger bulk.size.bytes and max.in.flight.requests values?

jjain1259:
> Jainam Jain (@jjain1259) Thanks for the PR — I added one minor comment. One general question: this introduces an additional transient copy of each bulk body via ByteBuffer.allocate(size). At worst, that could add heap usage of roughly bulk.size.bytes × max.in.flight.requests, with a new allocation for each attempt. Was this tested with larger bulk.size.bytes and max.in.flight.requests values?

@ubishnoi — tested this directly on a live devel cluster at `batch.size=4000`/`max.in.flight.requests=5` (~4MB bulks, ~4x the original footprint), on connector `lcc-devc1259wv3`. Two things worth sharing:

- **Code-level:** `coalesce()` allocates the merged buffer once per bulk and hands that exact object to the I/O reactor thread (`MultiBufferEntity.produceContent()`), which drains it directly — no further copies, no pooling/caching. It's GC-eligible the moment the request completes. So worst-case transient footprint at `mif=5` is ~5 × bulk size (e.g. ~20MB at 4MB bulks), not something that accumulates.

- **Empirically:** ran this live for several hours under a mix of conditions — idle, sustained full-backlog drain (confirmed via heap histogram that bulks were reaching 100% of `batch.size`, i.e. genuinely ~4MB each), a throughput increase, and a couple of task restarts. G1 Old Gen stayed in a bounded ~565-850MB band throughout; no growth trend that didn't resolve to something else. Reactor threads stayed at ~1.6% CPU busy even while draining full-size bulks, consistent with this being one big write rather than many small ones (matches the coalescing fix's measured reactor-CPU drop from ~2.5 cores to ~0.11 cores).

So: the heap-cost scaling you flagged is real and worth keeping in mind at very large `bulk.size.bytes`/`mif` combinations, but at the sizes we tested it's a small, transient, per-request allocation with no retention — not a regression against the old client.

src/main/java/io/confluent/connect/elasticsearch/CoalescingHttpClient.java
jjain1259: Good catch — added `CoalescingHttpClientTest` covering exactly this (9526591c, 0ea621a8):

- The shared `SEPARATOR` buffer isn't consumed across two sequential requests (mutation-verified: removing `.duplicate()` makes it fail exactly as you describe — the second bulk's newlines disappear).
- Plus two more cases for the method's full contract: the empty-body pass-through branch, and that merging preserves the original request's method/path/queryParams/headers, not just the body.

ubishnoi (approved):
LGTM

PR #943, opened 2026-09-19
CC-35256: Document the design intent behind the Java API Client migration

jjain1259:
## What

Comment-only change stacked on #942 (which stacks on #940). Adds short javadoc and inline comments at the points where the migration's design decisions are load-bearing but were not visible in the code, so the next maintainer does not "simplify" them back into a deadlock, a record reorder, or a thread leak. No code or test-logic changes.

## Where and why

| File | Comment added |
|---|---|
| `ElasticsearchClient` | Thread model of the three connector-owned pools: why the dispatcher exists (ingester-lock vs. HTTP-pool-lock deadlock on reactor threads) and why it must stay a separate pool from the ingester scheduler (slot-freeing work must not queue behind parked flushes / DLQ-blocked `afterBulk`). |
| `ElasticsearchClient` | Why the retry executor discards delayed tasks at shutdown; why `BulkIngester` gets no `backoffPolicy` (omitted, not `noBackoff()`: any policy creates an internal retry pool that reorders and leaks); `maxConcurrentRequests` passes N as-is. |
| `ElasticsearchClient` | Why the close-path flush is submitted async; why `bulkIngester.close()` is skipped with stuck records; the transport-before-executors shutdown ordering and the "ingester scheduler must outlive ingester.close()" rule. |
| `ElasticsearchClient` | `afterBulk(Throwable)` is terminal by design; `hasMapping()` compares the serialized mapping so dynamic-only mappings are not overwritten. |
| `RetryingElasticsearchAsyncClient` | The deadlock the dispatcher hop prevents; why `handleAsync` (not `whenCompleteAsync`/`thenApplyAsync`) is load-bearing. |
| `CoalescingHttpClient` | Why this is the only usable seam, effect and cost, upstream issue elastic/elasticsearch-java#1339, why the transport is built on `ElasticsearchTransportBase`. |
| `DataConverter` | Pre-serialized documents (fidelity, no parse, repeatable re-send); the literal 15.x upsert request; the explicit id on data-stream creates. |
| `Mapping` | Why the mapping is a Jackson tree rather than the typed builders. |
| `Validator` | Why it uses the plain transport and client. |
| `RetryUtil` | The zero-backoff guard. |
| `pom.xml` | Warning not to exclude `opentelemetry-api` (the transport loads it unconditionally at construction). |

## Stale comments retired (audit of every changed/new file)

- `AsyncOffsetTracker` javadoc and `ElasticsearchClientTest` still said "BulkProcessor"; `DataConverter` said "DeleteRequest".
- The disabled 7.x backwards-compatibility ITs in `ElasticsearchConnectorIT` claimed a cgroupv2 reason; they now say the real one (the 8.x client cannot talk to 7.x).
- The dangling "see the deadlock note at this client's construction site" pointer in `RetryingElasticsearchAsyncClient` on #940's head was already dropped by #942; the note it pointed at now exists in `ElasticsearchClient`'s class javadoc.

Deliberately not commented: anything the code already states, mechanical API ports, and pre-existing parity code. No review-ticket ids in the comments; the design ledger stays in the wiki.

## Verification

- `mvn validate test-compile` on JDK 17: `BUILD SUCCESS`, 0 Checkstyle violations.
- `RetryingElasticsearchAsyncClientTest`, `DataConverterTest`, `MappingTest`: 46/46 passed.
- Diff is comment-only: every changed Java line begins with `//`, `*`, or `/**`.


ubishnoi (approved):
LGTM

